PR #75, opened 2016-09-21
loosen version requirements to permit install on magento 2.0.x

annybs:
We have Adyen_Payment 1.4.3 installed, but need the latest release to fix the "blank page on 3DS failure" issue. We aren't on Magento 2.1 yet so composer install fails.

This commit loosens the version requirements so that Adyen_Payment can be installed on Magento 2.0.x. As far as I can tell, there are no significant changes in [1.4.3 -> 1.4.5](https://github.com/Adyen/adyen-magento2/compare/1.4.3...1.4.5) that actually necessitate 2.1.

I have installed this successfully from my fork but would prefer to install it from an official release presuming this PR is accepted.

Thanks!


rikterbeek:
Hi @annybs,

Thanks for the PR! We can do that I merged your PR.

Regards,
Rik
Adyen

PR #100, opened 2017-03-28
Fix for how auth is rewritten in Apache + PHP FPM

adamlevenson:
When using PHP-FPM with Apache 2.4 the HTTP Auth headers are being re-written to $_SERVER['REDIRECT_HTTP_AUTHORIZATION']. Since this variable is not being checked the authentication fails and throws an error for an undefined index.


rikterbeek:
Hi @adamlevenson,

Thanks for the PR! I have merged it into the code.

Regards,
Rik
Adyen

PR #138, opened 2017-07-17
Added new column processing to adyen notification table to know if no…

shruti-ranium:
…tification record is already processing under other cron-job

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -224,7 +224,16 @@ class Cron
         // create collection
         $notifications = $this->_notificationFactory->create();
         $notifications->addFieldToFilter('done', 0);
+        $notifications->addFieldToFilter('processing', 0);
         $notifications->addFieldToFilter('created_at', $dateRange);
+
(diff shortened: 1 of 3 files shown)


rikterbeek:
Hi @shruti-ranium,

Thanks for the PR!  I was looking if Magento2 has an option for mass-update but it looks like that is not possible. The two foreach loop would something that would be nice to resolve.

Could you make the version to 2.0.7 and the updates to this version so on the upgrade from 2.0.6 all users this execute will take place ?

Regards,
Rik
Adyen

shruti-ranium:
Hi Rik,

I have made the version changes .


Thanks,
Shruti

rikterbeek:
Hi @shruti-ranium,

Thanks for the change I merged it into the code.

Regards,
Rik
Adyen

rikterbeek:
Hi @shruti-ranium,

Saw this was causing issues with recurring_contract notifications so I made the following fix see below.

Regards,
Rik
Adyen

PR #152, opened 2017-07-26
Enable translations for Adyen cronjob, useful for order confirmation …

lancergr:
…email sending for Magento 2.1

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -26,7 +26,9 @@ namespace Adyen\Payment\Model;
 use Magento\Framework\Webapi\Exception;
 use Magento\Sales\Model\Order\Email\Sender\OrderSender;
 use Magento\Framework\App\Area;
-use Magento\Framework\ObjectManagerInterface;
+use Magento\Framework\App\AreaList;
+use Magento\Framework\Phrase\Renderer\Placeholder;
+use Magento\Framework\Phrase;
 
 class Cron
 {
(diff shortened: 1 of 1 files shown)


Model/Cron.php
```diff
@@ -190,7 +198,8 @@ public function __construct(
         \Adyen\Payment\Model\Resource\Billing\Agreement\CollectionFactory $billingAgreementCollectionFactory,
         \Adyen\Payment\Model\Api\PaymentRequest $paymentRequest,
         \Adyen\Payment\Model\Order\PaymentFactory $adyenOrderPaymentFactory,
-        \Adyen\Payment\Model\Resource\Order\Payment\CollectionFactory $adyenOrderPaymentCollectionFactory
+        \Adyen\Payment\Model\Resource\Order\Payment\CollectionFactory $adyenOrderPaymentCollectionFactory,
+        ObjectManagerInterface $objectManager
```
rikterbeek: better to use:
\Magento\Framework\App\AreaList $areaList

Model/Cron.php
```diff
@@ -212,6 +222,12 @@ public function __construct(
      */
     public function processNotification()
     {
+        // needed for Magento < 2.2.0 https://github.com/magento/magento2/pull/8413
+        $areaList = $this->_objectManager->get(\Magento\Framework\App\AreaList::class);
```
rikterbeek: then you can use:
$this->areaList->getArea(....)

rikterbeek: we could add version check as well

boldbart:
When will this be released? We're having issues with English order confirmations on Dutch store views.

Issue #154, opened 2017-07-27
Icon not showing on HPP integrations


kuwiano:
Hi everyone, we are using Magento version 2.1.7 and version 2.0.6 of your plugin and what is descibed in  #66 still happens

The generated link:
```/static/version1500970041/frontend/_view/fr_FR/Adyen_Payment/images/logos/cartebancaire.png```
The correct one:
```/static/version1500970041/frontend/Codazon/fastest_company/fr_FR/Adyen_Payment/images/logos/cartebancaire.png```

rikterbeek:
Hi @kuwiano,

Are you able to upgrade to  2.1.7 ? The reason why I do not like to hardcode the location is that you can't update your own icons in your own theme.

Regards,
Rik
Adyen

rikterbeek:
Hi @kuwiano,

The fix above should solve your issues. It is a core issue so hopefully they will fix it in the 2.2 release of Magento. When we merge the PR we can do the release next week for the time being just apply the changes manually and validate if this solves your issues.

Regards,
Rik
Adyen

PR #156, opened 2017-07-28
Fix: #154 HPP Icons not showing on checkout

kuwiano:
Add additional parameter for theme to generate correct url. 

Causes asset repo to use '<Vendor>/<theme>' instead of '_view' in the url.

See #154 for reference.

### DIFF
--- a/Helper/PaymentMethods.php
+++ b/Helper/PaymentMethods.php
@@ -212,7 +212,7 @@ class PaymentMethods extends AbstractHelper
 
                     $params = [];
                     // use frontend area
-                    $params = array_merge(['area' => 'frontend', '_secure' => $this->_request->isSecure()], $params);
+                    $params = array_merge(['area' => 'frontend', 'theme' => 'Magento/blank', '_secure' => $this->_request->isSecure()], $params);
 
                     $asset = $this->_assetRepo->createAsset('Adyen_Payment::images/logos/' .
                         $paymentMethodCode . '.png', $params);


rikterbeek:
Hi @kuwiano,

Are you able to upgrade to  2.1.7 ? The reason why I do not like to hardcode the location is that you can't update your own icons in your own theme.

Regards,
Rik
Adyen

kuwiano:
Hi @rikterbeek,

I get that hardcoding it isn't the perfect solution but I am already on the latest version (verified via backend and cli), that is version: 2.1.7. Do you know a different solution?

Regards,
Dominik

lancergr:
Hi @kuwiano 

What "themeModel" do you get back when calling the $this->_assetRepo->updateDesignParams without providing the "theme" parameter? Do you get a valid Magento Theme instance returned back?

Best Regards,
Giorgos
Adyen

kuwiano:
Hi @lancergr 

It seems I do. But look for yourself in the attachement.
pasted image at 2017_08_01 09_54 am 1
pasted image at 2017_08_01 09_54 am

lancergr:
Hi @kuwiano 

Finally we were able to reproduce it on Magento 2.1.7

Unfortunately is not yet resolved in Magento core (see https://github.com/magento/magento2/pull/7686 and https://github.com/magento/magento2/issues/10402)

In order to not overwrite already configured custom themes, we need to detect whether a valid themeModel has been set. I have pushed a PR for that (#158).

Best Regards,
Giorgos
Adyen

kuwiano:
Hi @lancergr,

Thank you this should just fine.

When will this be released?

Best Regards,

Dominik

rikterbeek:
Hi @kuwiano,

When we merge this into the develop branch we can do a new release. This is something we can do next week. I would advise you to adopt the changes manually so you have this fixed the issue directly.

Regards,
Rik
Adyen

PR #159, opened 2017-08-02
Display a nicer label for for adyen_hpp payment info

alessandroniciforo:
At this moment brand_code (e.g.: "paypal") is used, while method_title (e.g.: "PayPal") is nicer to show as payment method info


rikterbeek:
Hi @alessandroniciforo,

Thanks for the PR! 
This template is only used for the mailing to the shoppers so indeed title makes more sense.
Will merge the PR.

Regards,
Rik
Adyen

PR #165, opened 2017-08-10
Changed version to 2.1.0

lancergr:


Setup/UpgradeSchema.php
```diff
@@ -59,8 +59,8 @@ public function upgrade(SchemaSetupInterface $setup, ModuleContextInterface $con
             $this->updateSchemaVersion204($setup);
         }
 
-        if (version_compare($context->getVersion(), '2.0.7', '<')) {
-            $this->updateSchemaVersion207($setup);
+        if (version_compare($context->getVersion(), '2.1.0', '<')) {
+            $this->updateSchemaVersion210($setup);
```
rikterbeek: Why do you change the version you can keep this on 2.0.7. Changing it could lead to problems for developers using the develop branch.

lancergr: In case version 2.0.7 is released, this will cause problems to the ones upgrading from 2.0.7 to 2.1.0. Maybe is better idea to release 2.0.7 instead of 2.1.0 ?

rikterbeek: No let's keep 2.1.0 but 2.0.7 will be never released the next release will be 2.2.0 or 2.1.1.

PR #172, opened 2017-08-21
PW-243: Do not send CSE token for recurring cards if the updated expiry date is not changed,

Aleffio:
- Do not send CSE token for recurring cards if the updated expiry date is not changed
- isDuplicate now checks for the 'original_reference' field, this allows now to process two notifications with the same billing agreement but different payment.
- Reworked Cron.php to update ExpiryDate on existing billing agreements.


view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
```diff
@@ -84,16 +85,14 @@ define(
                     "method": self.method,
                     "additional_data": {
                         variant: self.agreement_data.variant,
+                        // recurring_detail_reference: "8315028947219783"
```
lancergr: remove

PR #183, opened 2017-08-30
adyen payment notification stalled by indexers

uabassguy:
We are having an issue where adyen payment is being held up by the rest of the indexers. Can you implement this patch to move it to its own group?


rikterbeek:
Hi @uabassguy,

Makes sense that we use our own cron group so you can configure it differently then the rest.
We will review and merge it into the develop branch.

Regards,
Rik
Adyen

PR #187, opened 2017-09-01
PW-251: Update device fingerprint for ratepay

Aleffio:


Helper/Data.php
```diff
@@ -770,6 +770,16 @@ public function isPaymentMethodOpenInvoiceMethod($paymentMethod)
         }
     }
 
+    public function getRatePayId()
+    {
+        return $this->getAdyenHppConfigData("ratepay_id");
+    }
+
+    public function calculateDeviceIdentToken($quote)
```
rikterbeek: This method is absolute ?

PR #188, opened 2017-09-05
Set correct param type

FreekVandeursen:
code compilation fails in Magento 2.2 because of mismatching type
with the parent class


rikterbeek:
Hi @FreekVandeursen,

Thanks for the PR! indeed wrong defined.

Regards,
Rik
Adyen

PR #196, opened 2017-10-04
PW-291: on OFFER_CLOSED, cancel the order

Aleffio:


ebaschiera:
I'm looking forward to a new release containig this PR :grin:

Issue #200, opened 2017-10-17
'Scan Credit Card' option should be available in all fields.


justinpappalil:
Use standard credit card id values so Safari can scan cards on mobile phones- We are using this option in our website but only see this function when you click into the card owner field.
Can this be visible when clicking in any card field,not just in card owner?

rikterbeek:
Hi @justinyatnam,

Thanks for letting us know. I think we need to add some classes and ids but it should be possible.
If you already have a solution please let us know and do a PR so we can add this.

Regards,
Rik
Adyen

justinyatnam:
Hello @rikterbeek 

We have added a solution https://github.com/Adyen/adyen-magento2/pull/205 here. Could you please review the same. When we trying to change the input type from number to text, the scan credit card option is showing for us.

Regards,
Justin

msilvagarcia:
I'm closing this issue due to inactivity.

Feel free to create a new one if there are any questions left.

Cheers!
Marcos
Adyen

PR #205, opened 2017-10-26
scan credit card in all available fields #200

Vinu-yatnam:


rikterbeek:
Hi @Vinu-yatnam,

Changing this field is not ideal for mobile devices because this field makes it possible to automatically show the numbers rather then the full keyboard layout. I think there are other solution to do this. I think by providing specific IDs to the fields makes this possible and as well allow to scan the card from the mobile browser. If you can modify your PR to support this that would be great. 

Regards,
Rik
Adyen

Issue #204, opened 2017-10-25
Integrity constraint violation


ajpevers:
Same problem as in #107

```
SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '8815088272901681'
for key 'ADYEN_ORDER_PAYMENT_PSPREFERENCE', query was: INSERT INTO
`adyen_order_payment` (`pspreference`, `merchant_reference`, `payment_id`,
`payment_method`, `amount`, `total_refunded`, `created_at`, `updated_at`) VALUES
(?, ?, ?, ?, ?, ?, '2017-10-24 12:20:16', '2017-10-24 12:20:16')
```

@rikterbeek I will let you know as soon as I find out if errors are occurring before this message and what the end result of the payment is.

This is with Magento 2.1.9 and adyen/module-payment 2.1.2

ajpevers:
These are test payments:

There are logs containing this PSPReference in 3 files:
var/log/adyen/result.log (1x)
var/log/adyen/cronjob.log (2x)
var/log/adyen/notification.log (2x)

@rikterbeek does this mean that 2 notifications are received by the webshop concerning the same payment?

var/log/adyen/result.log:
```
[2017-10-24 06:41:34] AdyenLoggerTest.ADYEN_RESULT: Array
(
    [authResult] => AUTHORISED
    [merchantReference] => 13110152
    [merchantSig] => EwcQmu6R3nu5E4qkphXwEIlpWp1QTrKFWBSi4d5xv84=
    [paymentMethod] => ideal
    [pspReference] => 8815088272901681
    [shopperLocale] => nl_NL
    [skinCode] => EIyhP2o0
)
 {"is_exception":false} []
[2017-10-24 06:41:34] AdyenLoggerTest.ADYEN_RESULT: Processing ResultUrl {"is_exception":false} []
[2017-10-24 06:41:34] AdyenLoggerTest.ADYEN_RESULT: Updating the order {"is_exception":false} []
[2017-10-24 06:41:34] AdyenLoggerTest.ADYEN_RESULT: Do nothing wait for the notification {"is_exception":false} []
```

2 consecutive entries in var/log/adyen/cronjob.log:
```
[2017-10-24 12:20:14] AdyenLoggerTest.ADYEN_NOTIFICATION_CRONJOB: Processing notification 443 {"is_exception":false} []
[2017-10-24 12:20:14] AdyenLoggerTest.ADYEN_NOTIFICATION_CRONJOB: Array
(
[...]

PR #208, opened 2017-11-03
#204 Add duplicate check into notificaiton processing and fix formatting

rikterbeek:
**Description**

If a duplicate notification is received do not process it.

**Tested scenarios**

Send duplicate notifications into the system at the same timeperiod. Or go into the database duplicate a non processed notification.

**Fixed issue**:  
This is a fix for issue #204

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -226,7 +226,7 @@ class Cron
     {
         // needed for Magento < 2.2.0 https://github.com/magento/magento2/pull/8413
         $renderer = Phrase::getRenderer();
-        if($renderer instanceof Placeholder) {
+        if ($renderer instanceof Placeholder) {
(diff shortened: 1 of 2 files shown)


Model/Cron.php
```diff
@@ -247,34 +247,36 @@ public function processNotification()
...
-            $notification->save();
+            $this->_updateNotification($notification, true, false);
         }
 
         // loop over the notifications
         $count = 0;
         foreach ($notifications as $notification) {
 
-            $this->_adyenLogger->addAdyenNotificationCronjob(
-                sprintf("Processing notification %s", $notification->getEntityId())
```
Aleffio: Don't we still need to log the entityID?
Or are we doing it somewhere else?

rikterbeek: we log the entity id of the notification so you can trace what specific notification is being processed.

PR #213, opened 2017-11-14
Fixes #186 Remove minify_exclude and load df.js on runtime

lancergr:
**Description**
Remove minify_exclude and load df.js on runtime

**Tested scenarios**
checkout with Ratepay and Klarna in order to send the correct dfValue
tested minified and non-minified in wysiwyg

**Fixed issue**:  #186 #207

PR #212, opened 2017-11-14
Added sending of invoice emails if enabled in backend

robinwestra:
**Description**
Replaced old Magento 1 code for invoice mails with Magento 2 methods. 
Also checks the default Magento 2 setting for invoice mails instead of a module specific setting.

**Tested scenarios**
Tested sending of invoice mails with the setting enabled and disabled, behaves as expected.

**Fixed issue**:  #210

PR #197, opened 2017-10-09
always populate cc_type column in sales_order_payment to feed order g…

ebaschiera:
…rid later

PR #195, opened 2017-10-03
Fix duplicate notifications

florisgoumans:
If $originalReference is NULL it will try to filter and eventually create duplicate notifications in the table.

PR #214, opened 2017-11-16
Hotfix release 2.1.3

rikterbeek:
**Description**

Merge from develop to master for new release

**Tested scenarios**

Each issue has been tested

**Fixed issue**:  
#196, #195, #197, #198,#202,#204,#208,#186,#212,#213


Model/Resource/Notification.php
```diff
@@ -36,21 +36,26 @@ public function _construct()
...
+    public function getNotification($pspReference, $eventCode, $success, $originalReference, $done = null)
     {
         $select = $this->getConnection()->select()
             ->from(['notification' => $this->getTable('adyen_notification')])
             ->where('notification.pspreference=?', $pspReference)
             ->where('notification.event_code=?', $eventCode)
             ->where('notification.success=?', $success);
 
+        if ($done) {
+            $select->where('notification.done=?', $done);
```
lancergr: this will not be executed if done=false

shall we change it to if($done===null) instead?

rikterbeek: False is not used in the code at the moment but agree better to check if it is null so we can support false in the future.

PR #218, opened 2017-11-28
Config import with serialized value

mwienk:
**Description**
I have added a check if the configuration value is already serialized. When using the app:config:import command of M2.2 (http://devdocs.magento.com/guides/v2.2/config-guide/cli/config-cli-subcommands-config-mgmt-import.html), the configuration import for Adyen fails because it is already serialized. I think it isn't necessary to parse the config again so I just return $this, but I could be mistaken.

**Tested scenarios**
Tested import with config and configuration saving from backend

**Fixed issue**: -

### DIFF
--- a/Model/Config/Backend/Installments.php
+++ b/Model/Config/Backend/Installments.php
@@ -62,6 +62,10 @@ class Installments extends \Magento\Framework\App\Config\Value
     public function beforeSave()
     {
         $value = $this->getValue();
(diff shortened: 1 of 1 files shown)


rikterbeek:
Hi @mwienk,

Thanks for your Pull Request. I understand your problem we will review the code and merge it.

Regards,
Rik
Adyen

PR #229, opened 2017-12-20
Fix installments

Aleffio:
**Description**
Fix installments to show only configured values

**Tested scenarios**
Credit Card and OneClick

**Fixed issue**:  Installments dropdown menu was showing all the values ranging from 1 to the max configured number of installments. Now it is only showing the values set in the backend.


rikterbeek:
@Aleffio can you fix the PR, it is now conflicting

PR #235, opened 2018-01-09
fix-hpp-gender-layout: gender field in hpp did not layout correctly. …

victorhuisman:
…Fixed by removing type from class.
**Description**
In the hpp page, the layout of the gender field is not ok. See screenshot. Souce of this issue is a 'type' class for the gener field. This results in a margin setting for the control field. Removing the type class (so it is the same as the DoB field) results in not applying the margin-left, which results in a correct layout.
image

**Tested scenarios**
Removed class "type" which resulted in the correct rendering of the gender field.


rikterbeek:
Hi @victorhuisman,

Many thanks for the PR! we will merge it.

Regards,
Rik
Adyen

PR #238, opened 2018-01-12
PW-349: Show message if not on latest version of the extension

Aleffio:
**Description**
After login create the adminnotification item in the inbox with the info about new version releases.
Show a message after logging in if the version installed is not the latest one available.

**Tested scenarios**
Tested the message and the adminnotification item in the inbox

**Fixed issue**:  N/A

### DIFF
--- a/AdminMessage/VersionMessage.php
+++ b/AdminMessage/VersionMessage.php
@@ -29,7 +29,6 @@ class VersionMessage implements \Magento\Framework\Notification\MessageInterface
     protected $_authSession;
     protected $_adyenHelper;
     protected $_inboxFactory;
-    protected $_githubVersion;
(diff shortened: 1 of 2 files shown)


AdminMessage/VersionMessage.php
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Retrieve system message text
+     *
+     * @return \Magento\Framework\Phrase
+     */
+    public function getText()
+    {
+        try {
+            $githubContent = $this->getDecodedContentFromGithub();
```
lancergr: is this doing a second request? if so, we can either prepare the text or store the GH content from "isDisplayed" method

PR #239, opened 2018-01-12
Fix PayPal capture mode

maassenbas:
**Description**
PayPal capture mode set to immediate/manual based on configuration "Use manual capture for PayPal".

**Tested scenarios**
Capture delay on Manual and PayPal capture on Immediate, results in Immediate capture. 
Capture delay on Manual and PayPal capture on Manual, results in Manual capture

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -210,7 +210,8 @@ class Cron
         \Adyen\Payment\Model\Order\PaymentFactory $adyenOrderPaymentFactory,
         \Adyen\Payment\Model\Resource\Order\Payment\CollectionFactory $adyenOrderPaymentCollectionFactory,
         AreaList $areaList
-    ) {
+    )
+    {
         $this->_scopeConfig = $scopeConfig;
         $this->_adyenLogger = $adyenLogger;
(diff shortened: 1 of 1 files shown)


Model/Cron.php
```diff
@@ -1248,18 +1249,19 @@ protected function _isAutoCapture()
...
             // if PayPal capture modues is different from the default use this one
-            if (strcmp($this->_paymentMethod, 'paypal') === 0 && $captureModePayPal != "") {
-                if (strcmp($captureModePayPal, 'auto') === 0) {
-                    $this->_adyenLogger->addAdyenNotificationCronjob(
-                        'This payment method is paypal and configured to work as auto capture'
-                    );
-                    return true;
-                } elseif (strcmp($captureModePayPal, 'manual') === 0) {
+            if (strcmp($this->_paymentMethod, 'paypal') === 0) {
+                if ($captureModePayPal) {
```
lancergr: good one, I would also suggest to change the variable name to a less misleading one (e.g. $manualCapturePayPal) 
..and maybe return !$manualCapturePayPal in both cases

PR #240, opened 2018-01-16
fix for klarna_account

rikterbeek:
**Description**

klarna_account always goes to auto-capture where it is a manual capture payment method

**Tested scenarios**

run all test cases successfully

**Fixed issue**:  
Klarna_account not going into manual capture

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,6 +5,6 @@
 
 **Tested scenarios**
 <!-- Description of tested scenarios -->
-<!-- Please verify that the unit tests are passing by running "vendor/bin/phpunit -c ." -->
+<!-- Please verify that the unit tests are passing by running "vendor/bin/phpunit -c dev/tests/unit/phpunit.xml.dist vendor/adyen/module-payment/Test/" -->
 
 **Fixed issue**:  <!-- #-prefixed issue number -->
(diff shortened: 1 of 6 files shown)


Helper/Data.php
```diff
@@ -849,13 +849,13 @@ public function isPaymentMethodOpenInvoiceMethod($paymentMethod)
     {
         if (strlen($paymentMethod) >= 9 && substr($paymentMethod, 0, 9) == 'afterpay_') {
             return true;
-        } else {
-            if ($paymentMethod == 'klarna' || $paymentMethod == 'ratepay') {
-                return true;
-            } else {
-                return false;
-            }
+        } elseif (strlen($paymentMethod) >= 6 && substr($paymentMethod, 0, 6) == 'klarna') {
```
lancergr: strlen can be omitted here since substr will return a string containing at most 6 characters
we could even switch to strpos($paymentMethod, 'klarna') === 0 to avoid the letter counting

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,6 +5,6 @@
 
 **Tested scenarios**
 <!-- Description of tested scenarios -->
-<!-- Please verify that the unit tests are passing by running "vendor/bin/phpunit -c ." -->
+<!-- Please verify that the unit tests are passing by running "vendor/bin/phpunit -c dev/tests/unit/phpunit.xml.dist vendor/adyen/module-payment/Test/" -->
```
lancergr: then let's remove this as well https://github.com/Adyen/adyen-magento2/blob/10f2a982c9e6a33e840e4434188928856a27fb3a/phpunit.xml

rikterbeek: is removed

PR #246, opened 2018-01-28
Cron notification processing: Save payment after making changes to it.

evktalo:
Sending capture for Klarna invoices when making an invoice in Magneto (2.1.10) does not seem to work without this.

**Description**

Save payment in two places after making changes to it. Otherwise the data in sales_order_payment will not be updated.

**Tested scenarios**

* Make a test order with Klarna open invoice as the payment method (within HPP)
* From Magento's order view, create invoice, choose "Capture Online", we get "We can't save the invoice right now." error in admin and in adyen/error.log we can see the originalReference field sent in the capture request is empty

After changes, the invoice can be created, the capture request is sent successfully and in ca I can see status of payment advance to "SentForSettle".

Module version tested is 2.1.3.

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -670,6 +670,8 @@ class Cron
         if (isset($authCode) && $authCode != "") {
             $this->_order->getPayment()->setAdditionalInformation('adyen_auth_code', $authCode);
         }
+
+        $this->_order->getPayment()->save();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


rikterbeek:
Hi @evktalo,

Thanks for your comment and PR. What I found strange is that I don't have this problem on my end. What version of magento2 are you running on ? i would assume the $order->save() would save all this information. see https://github.com/Adyen/adyen-magento2/blob/develop/Model/Cron.php#L379 could you verify if this is being executed ?

Regards,
Rik
Adyen

evktalo:
We are running Magento 2.10.0 where this problem is occurring. I'll get to checking that the $order->save is executed in a couple of days. Thank you for the reply!

rikterbeek:
Hi @evktalo,

Were you able to check it ?

Regards,
Rik
Adyen

Issue #248, opened 2018-01-31
Migrate Magento 2 place order function race condition bug


NickdeK:
**Magento version**: 2.2.1
**Plugin version**: 2.2.0
**Description**
I noticed you still use the old way of disabling the order button when placing a order. Magento changed  this a few months ago because this could cause a race condition. Could you do the same in your code to prevent this from happening?

Fix from magento:
https://github.com/magento/magento2/pull/11901

Affected files:
https://github.com/Adyen/adyen-magento2/blob/develop/view/frontend/web/template/payment/cc-form.html#L286-L287
https://github.com/Adyen/adyen-magento2/blob/develop/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js#L152-L159
(and this of course for all payment types)

NickdeK:
Hi @rikterbeek,

I just create a pull request but I only applied it to the creditcard payment method because I'm not sure if the other payment types could have the same issue. Maybe someone could check into this.

With kind regards,
Nick de Kleijn

msilvagarcia:
I'm closing this issue due to inactivity.

Feel free to create a new one if there are any questions left.

Cheers!
Marcos
Adyen

PR #249, opened 2018-01-31
Place Order function race condition fix

NickdeK:
**Description**
Migrate changes from Magento 2 race condition fix to Adyen Payment Module

**Tested scenarios**
1. Add items to cart
2. Enter order details
3. Select Adyen (CreditCard) payment type
4. Enter credentials (using adyen test mode and test credentials)
5. Continually click the "Place Order" button
6. Only one request should be sent to Adyen

**Todo**

- [ ] Apply fix on other payment methods where this could also be an issue

**Fixed issue**:  https://github.com/Adyen/adyen-magento2/issues/248

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
@@ -152,6 +152,15 @@ define(
             isActive: function () {
                 return true;
             },
+
+            /**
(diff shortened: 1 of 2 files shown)


rikterbeek:
Hi NickdeK,

Many thanks for the PR looks good, we will review it and check how much work it is to fix this for the other payment methods.

Regards,
Rik
Adyen

PR #257, opened 2018-02-22
Made payment URL generation method public

Maikel-Koek:
**Description**
We want to use the payment URL generation method outside of the PayByMail gateway command.


rikterbeek:
Hi @Maikel-Koek,

Thanks for the PR. Yes I will merge this.

Regards,
Rik
Adyen

Issue #254, opened 2018-02-15
Possibility to create an Adyen payment link for any order, and correctly process the payment on the order


Maikel-Koek:
Plus, the possibility to do this with a new 'Payment Required' status.

Maikel-Koek:
https://github.com/Adyen/adyen-magento2/pull/257

msilvagarcia:
I'm closing this issue due to inactivity.

Feel free to create a new one if there are any questions left.

Cheers!
Marcos
Adyen

PR #258, opened 2018-02-22
Show payment link in admin for remaining order balance

Maikel-Koek:
This link only shows when the original order payment method is an
Adyen method, the total due is greater than 0 and the invoiced total
is greater than 0.

Solves #254.

Payment info at order in backend:

schermafbeelding 2018-02-22 om 15 12 57

### DIFF
--- a/Block/Info/PaymentLink.php
+++ b/Block/Info/PaymentLink.php
@@ -86,6 +86,8 @@ class PaymentLink extends AbstractInfo
     public function _toHtml()
     {
         return strpos($this->getPayment()->getMethod(), 'adyen_') === 0
+            && $this->_scopeConfig->getValue('payment/adyen_hpp/active')
+            && $this->_scopeConfig->getValue('payment/adyen_pay_by_mail/active')
             && $this->getOrder()->getTotalInvoiced() > 0
             && $this->getOrder()->getTotalDue() > 0
                 ? parent::_toHtml()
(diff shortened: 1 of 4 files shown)


rikterbeek:
Hi @Maikel-Koek,

Looking good. I would say only to do an extra check to only show it when the payment method was adyen_hpp could you add that ?


Regards,
Rik
Adyen

Maikel-Koek:
Hi @rikterbeek ,

We actually discussed that statement here (@paales), and we thought the opposite, to even remove it completely, so show the remainder payment link for every payment method.

Why would you only want to show it when an HPP method was used?

rikterbeek:
Hi @Maikel-Koek,

Problem that can occur is that adyen_HPP needs to be configured otherwise it will fail the payment link. Maybe we can say if adyen_paybymail and adyen_hpp are enabled to show it? as well you do not want to show the button if a card payment is reserved but the invoice is not created (in a auth/cap flow the invoice will be generated on capture payment).

Regards,
Rik
Adyen

Maikel-Koek:
Hi @rikterbeek,

I see, I can add a check if those methods are enabled, but not necessarily used on the order.

To prevent it from showing when a card payment is pending, I check the `total invoiced > 0 AND total due > 0`, which I think will suffice for most cases? Because most order payments are made at once? Which will have a total invoiced of zero.

PR #260, opened 2018-03-21
Update module-vault dependency to version 100.2

jzahedieh:
**Description**
In Magento 2.1.8 they added a major version upgrade to the vault: 100.2.* in order to keep compatibility with this new feature with the 2.1.* version of Magento the composer file needs to be updated.

Currently it fails with the following error on `composer update`:

	  Problem 1
	    - Installation request for adyen/module-payment dev-feature/vault -> satisfiable by adyen/module-payment[dev-feature/vault].
	    - Conclusion: remove magento/product-community-edition 2.1.12
	    - magento/product-enterprise-edition 2.1.12 requires magento/product-community-edition 2.1.12 -> satisfiable by magento/product-community-edition[2.1.12].
	    - magento/product-enterprise-edition 2.1.12 requires magento/product-community-edition 2.1.12 -> satisfiable by magento/product-community-edition[2.1.12].
	    - Conclusion: don't install magento/product-community-edition 2.1.12
	    - Installation request for magento/product-enterprise-edition 2.1.12 -> satisfiable by magento/product-enterprise-edition[2.1.12].

**Tested scenarios**

Composer installed successfully from `dev-vault-version-fix` fork branch:

	  - Updating adyen/module-payment (dev-master 787f9f5 => dev-vault-version-fix 73cab08)
	    Checking out 73cab0812df6e4eaba4c8f0d5bf0b07af1047851

**Fixed issue**:  
Relates to #252 Register a new Vault entry when processing a RECURRING_CONTRACT notification (for CC)

Issue #252, opened 2018-02-15
Register a new Vault entry when processing a RECURRING_CONTRACT notification (for CC)


Maikel-Koek:

rikterbeek:
check https://github.com/Adyen/adyen-magento2/tree/feature/vault

msilvagarcia:
I'm closing this issue due to inactivity.

Feel free to create a new one if there are any questions left.

Cheers!
Marcos
Adyen

PR #261, opened 2018-03-21
Vault detail handler fix

jzahedieh:
Relates to pull request https://github.com/Adyen/adyen-magento2/pull/260

**Description**
Magento have changed how `PaymentTokenInterfaceFactory` can be used, as it wasn't part of the official API it has broken in the version I am integrating (2.1.12) there is a core issue and they suggested using the core concrete `CreditCardTokenFactory` class instead 
https://github.com/magento/magento2/issues/8083

Additionally I've removed some dead code and tidied up the code to be PSR compliant 

**Tested scenarios**

I can now successfully tokenize credit cards on 2.1.12 whilst placing admin orders, unfortunately I do not have a lower version of 2.1.* to test this on, it should be backwards compatible though.

**Fixed issue**:  
Relates to #252 Register a new Vault entry when processing a RECURRING_CONTRACT notification (for CC)

### DIFF
--- a/Gateway/Response/VaultDetailsHandler.php
+++ b/Gateway/Response/VaultDetailsHandler.php
@@ -23,15 +23,18 @@
 
 namespace Adyen\Payment\Gateway\Response;
 
-use Magento\Vault\Api\Data\PaymentTokenFactoryInterface;
+use Magento\Vault\Model\CreditCardTokenFactory;
+use Magento\Vault\Api\Data\PaymentTokenInterface;
 use Magento\Payment\Gateway\Response\HandlerInterface;
 use Magento\Payment\Model\InfoInterface;
+use Magento\Sales\Api\Data\OrderPaymentExtensionInterface;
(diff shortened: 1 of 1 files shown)


elioermini:
Magento will keep supporting 2.0 2.1 and 2.2. along with 2.3 when it will be released as these are feature releases

jzahedieh:
Ah that is frustrating, we're stuck on 2.1.12 for the time being, whilst the 2.2.* upgrade is on the horizon  we're having some issues with database schema changes so it's taking longer.

From the main README.md

> This plugin supports Magento2 version 2.1 and higher

Is using the Magento Vault only going to be supported with 2.2 and higher, since the factory will be depreciated in 2.3 which isn't released yet.

rikterbeek:
@jzahedieh,

Is it still an issue to update to later version ? We really want to focus on the release that supports all later version of Magento.

Regards,
Rik
Adyen

PR #268, opened 2018-04-19
Replace "Place order" by "Place Order" like Magento

VincentMarmiesse:
Hello,

**Description**
This PR replaces `Place order` by `Place Order` because that's what Magento is using.
Currently we have to translate `Place Order` and `Place order` which is not very convenient and can cause some misunderstandings.

What do you think?


rikterbeek:
HI @VincentMarmiesse,

Thanks for your PR. This makes sense will merge it.

Regards,
Rik
Adyen

PR #271, opened 2018-05-07
[VM-336] +: Made return path (checkout/cart) configurable

Maikel-Koek:
**Description**

**Tested scenarios**

**Fixed issue**:

PR #263, opened 2018-03-26
Compatibility with PHPUnit 6

JeroenVanLeusden:
**Description**
Add compatibility for PHPUnit 6 + some code formatting.

PR #262, opened 2018-03-23
Create nl_NL.csv

peterjaap:
**Description**
Dutch translation

**Tested scenarios**

**Fixed issue**:

PR #281, opened 2018-06-22
Merge for new release 2.3.0

rikterbeek:
**Description**

#238 PW-349: Show message if not on latest version of the extension
#239 Fix PayPal capture mode 
#240 fix for Klarna account integration
#242 Remove Cron controller should not be accessible from browser
#247 Added boleto logo
#249 Place Order function race condition fix 
#250 Fix for showing only selected boleto types in checkout 
#257 Made payment URL generation method public 
#258 Show payment link in admin for remaining order balance
#262 Added NL translation file
#263 Compatibility with PHPUnit 6
#267 showInStore missing on some groups
#268 Replace "Place order" by "Place Order" like Magento
#271 Made return path (checkout/cart) configurable
#273 Use store scope when populating HPP methods 
#274 Removed BCMC from manual capture payment methods
#276 PW-256 Support multiple partial Captures for open invoice payment methods
#277 Added WeChatPay logos
#278 PW-455 Support partial refunds for open invoice payment methods
#280 Fixed mismatching method signatures (API) 
#281 Use CC backend template for oneclick payments
Use core methods to place order instead of overwritten all core code

**Tested scenarios**

**Fixed issue**:


Setup/UpgradeSchema.php
```diff
@@ -63,6 +64,10 @@ public function upgrade(SchemaSetupInterface $setup, ModuleContextInterface $con
             $this->updateSchemaVersion207($setup);
         }
 
+        if (version_compare($context->getVersion(), '2.2.1', '<')) {
```
Aleffio: Should we change this to 2.3.0?

rikterbeek: no not needed this will be executed as release before was 2.2.0.

PR #287, opened 2018-07-05
Set state 

maassenbas:
Set State based on Order Status

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -188,6 +188,11 @@ class Cron
     protected $_areaList;
 
     /**
+     * @var \Magento\Sales\Model\ResourceModel\Order\Status\CollectionFactory
+     */
+    protected $_orderStatusCollection;
+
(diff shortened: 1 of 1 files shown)


Model/Cron.php
```diff
@@ -220,8 +225,10 @@ public function __construct(
         \Adyen\Payment\Model\Order\PaymentFactory $adyenOrderPaymentFactory,
         \Adyen\Payment\Model\ResourceModel\Order\Payment\CollectionFactory $adyenOrderPaymentCollectionFactory,
         \Adyen\Payment\Model\InvoiceFactory $adyenInvoiceFactory,
-        AreaList $areaList
-    ) {
+        AreaList $areaList,
+        \Magento\Sales\Model\ResourceModel\Order\Status\CollectionFactory $orderStatusCollection
```
rikterbeek: Add this into the comments as well.

PR #290, opened 2018-07-17
PW-514 Invoice on shipment option for openinvoice payment methods

rikterbeek:
**Description**

If setting enabled it is possible to auto invoice payments on shipment for openinvoice payment methods. (Klarna,AfterPay and RatePay)
**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Observer/Adminhtml/BeforeShipmentObserver.php
+++ b/Observer/Adminhtml/BeforeShipmentObserver.php
@@ -1,0 +1,93 @@
+<?php
+/**
+ *                       ######
+ *                       ######
+ * ############    ####( ######  #####. ######  ############   ############
+ * #############  #####( ######  #####. ######  #############  #############
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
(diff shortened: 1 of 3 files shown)


anup4ranium:
@rikterbeek Does it cater for partial shipments and creates partial invoice ?
If not , then ideally it should create invoice only when all items of order are shipped .

rikterbeek:
Hi @anup4ranium,

No works only for full invoices. Do you know easily how to support partial captures ?

Regards,
Rik
Adyen

anup4ranium:
@rikterbeek  We have implemented it as below : 

1. Added hook on **sales_order_shipment_save_after** event
2. Creating Invoice when `shipped-items == ordered-items`

rikterbeek:
Hi @anup4ranium,

I want to do it before shipment as if invoice failed there will be no shipment. For now this cover a use-case of a customer ideally this is beging handled by the warehouse solution.

Regards,
Rik
Adyen

anup4ranium:
Hi @rikterbeek  

Makes sense to halt shipment if Invoice fails 👍 
But its true that here capture can happen before ALL items are shipped in case of partial shipments. 
Not sure if it is acceptable legally with different open-invoice payment providers.

Thanks

PR #294, opened 2018-08-02
Vault support for 3D secure payments

Maikel-Koek:
**Description**
Vault support for 3D secure payments

@rikterbeek Is the feature/vault branch merged to develop somewhere in the near future?
We're running the Vault integration (including the changes in the PR) for some time now, live at one of our customers, and are not running into any problems.


rikterbeek:
Hi @Maikel-Koek,

Many thanks for the PR. Yes we have a ticket to merge it unfortunately it is not that easy.
We still need to support oneclick as well and this does not work with token so we need an options to select what one you want. At the moment we have some other prioritisation. For now just leave it here I will make sure it will be in sync with the develop branch.

Regards,
Rik
Adyen

PR #295, opened 2018-08-07
B: Use order repository to load order instead of orderFactory

Maikel-Koek:
Else the afterGet plugins won't be loaded

**Description**
Use the order repository, using the order factory is legacy.
Plugins on the order repository are not triggered when the factory is used.


rikterbeek:
Hi @Maikel-Koek,

Yes makes total sense. I will merge it.

Regards,
Rik
Adyen

PR #296, opened 2018-08-07
[VM-297] +: Also show frontend payment link when nothing is invoiced yet

Maikel-Koek:
**Description**
You also want to be able to send a payment link of an order to the customer when there's nothing paid yet.


rikterbeek:
Hi @Maikel-Koek,

Thanks for the PR. Makes sense.

Regards,
Rik
Adyen

PR #298, opened 2018-08-08
B: Only delete Vault token remotely when payment method is Adyen

### LINKED ISSUE #297: Deleting Vault token plugin throws exception
**Magento version**: 2.2.5
**Plugin version**: feature/vault
**Description**

Can be reproduced by running `\Magento\Braintree\Controller\Cards\DeleteActionTest::testExecute`.

The `assertSessionMessages` assertion will fail, because the exception "Failed to disable this contract" will be triggered, which is added here:
https://github.com/Adyen/adyen-magento2/commit/5f1ff32cf1ab5e8d4b229515aa3e0a86afd2794a

The plugin probably needs to check if the payment token contains an Adyen payment method code before proceeding, so the errors isn't triggered when another payment provider is trying to delete a token.

Maikel-Koek:
Fixes #297

**Description**
Only try to delete a Vault token remotely at Adyen, when payment method of token is Adyen.

**Tested scenarios**
Running the integration test `\Magento\Braintree\Controller\Cards\DeleteActionTest::testExecute`.

**Fixed issue**:  #297

### DIFF
--- a/Plugin/PaymentVaultDeleteToken.php
+++ b/Plugin/PaymentVaultDeleteToken.php
@@ -27,7 +27,6 @@ use Magento\Vault\Api\Data\PaymentTokenInterface;
 
 class PaymentVaultDeleteToken
 {
-
     /**
      * @var \Adyen\Payment\Model\Api\PaymentRequest
      */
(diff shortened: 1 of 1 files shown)


rikterbeek:
Hi @Maikel-Koek,

Thanks for the PR makes sense indeed.

Regards,
Rik
Adyen

PR #299, opened 2018-08-10
Adding cardBin and expiryDate additionalData

lancergr:


rikterbeek:
do we not have to add it to the admin templates then ?

PR #304, opened 2018-08-20
PW-503: Parse Payment receipt from TerminalAPI PaymentResponse

Aleffio:
**Description**
Adding customer receipt for terminal api on the database

**Tested scenarios**
My Orders menu(frontend)
Orders menu(backend)

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -96,8 +96,7 @@ class Data extends AbstractHelper
         \Magento\Framework\View\Asset\Repository $assetRepo,
         \Magento\Framework\View\Asset\Source $assetSource,
         \Adyen\Payment\Model\ResourceModel\Notification\CollectionFactory $notificationFactory
-    )
-    {
+    ) {
         parent::__construct($context);
         $this->_encryptor = $encryptor;
         $this->_dataStorage = $dataStorage;
@@ -813,9 +812,7 @@ class Data extends AbstractHelper
                     if ($this->showLogos()) {
                         $logoName = $agreementData['variant'];
 
-                        $asset = $this->createAsset(
-                            'Adyen_Payment::images/logos/' . $logoName . '.png'
-                        );
+                        $asset = $this->createAsset('Adyen_Payment::images/logos/' . $logoName . '.png');
 
                         $icon = null;
                         $placeholder = $this->_assetSource->findSource($asset);
@@ -896,9 +893,7 @@ class Data extends AbstractHelper
      */
     public function isVatCategoryHigh($paymentMethod)
     {
-        if ($paymentMethod == "klarna" ||
-            strlen($paymentMethod) >= 9 && substr($paymentMethod, 0, 9) == 'afterpay_'
-        ) {
+        if ($paymentMethod == "klarna" || strlen($paymentMethod) >= 9 && substr($paymentMethod, 0, 9) == 'afterpay_') {
             return true;
         }
         return false;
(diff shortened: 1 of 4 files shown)


view/adminhtml/templates/info/adyen_pos_cloud.phtml
```diff
@@ -61,3 +61,10 @@ $_isDemoMode = $block->isDemoMode();
 <?php endif;?>
 
 <?php echo $block->getChildHtml()?>
+    <br>
```
rikterbeek: can you remove this BR and handle the margin on the table ?

Helper/Data.php
```diff
@@ -1024,4 +1019,27 @@ public function getAdyenMerchantAccount($method, $storeId)
         return $merchantAccount;
     }
 
+    public function formatTerminalAPIReceipt($paymentReceipt)
+    {
+        $paymentReceipt = json_decode($paymentReceipt);
+        $formatted = "<!DOCTYPE html><html><body>";
```
rikterbeek: it is inside an already html block right so we don't need this I assume ?

Helper/Data.php
```diff
@@ -1024,4 +1019,27 @@ public function getAdyenMerchantAccount($method, $storeId)
...
+                            $formatted .= "<td class='terminal-api-receipt-name'>" . $get_array['name'] . "</td>";
+                        }
+                        if (!empty($get_array['value'])) {
+                            $formatted .= "<td class='terminal-api-receipt-value' align='right'>" . $get_array['value'] . "</td>";
+                        }
+                        $formatted .= "</tr>";
+                }
+            }
+        }
+        $formatted .= "</body></html>";
```
rikterbeek: same with this.

Helper/Data.php
```diff
@@ -1024,4 +1019,27 @@ public function getAdyenMerchantAccount($method, $storeId)
...
     }
 
+    public function formatTerminalAPIReceipt($paymentReceipt)
+    {
+        $paymentReceipt = json_decode($paymentReceipt);
+        $formatted = "<table class='terminal-api-receipt'>";
+        foreach ($paymentReceipt as $receipt) {
+            if ($receipt->DocumentQualifier == "CustomerReceipt") {
+                foreach ($receipt->OutputContent->OutputText as $item) {
+                    parse_str($item->Text, $get_array);
```
lancergr: let's use camelCase eg. textParts instead of get_array

Helper/Data.php
```diff
@@ -1024,4 +1019,27 @@ public function getAdyenMerchantAccount($method, $storeId)
...
+                    $formatted .= "<tr class='terminal-api-receipt'>";
+                    if (!empty($get_array['name'])) {
+                        $formatted .= "<td class='terminal-api-receipt-name'>" . $get_array['name'] . "</td>";
```
lancergr: let's append the TDs even if they have empty content and ideally put &nbsp; http://jkorpela.fi/HTML/emptycells.html

view/adminhtml/templates/info/adyen_pos_cloud.phtml
```diff
@@ -61,3 +61,10 @@ $_isDemoMode = $block->isDemoMode();
 <?php endif;?>
 
 <?php echo $block->getChildHtml()?>
+    <style>
+        table.terminal-api-receipt{margin-top: 10px}
+        td.terminal-api-receipt-name{padding:inherit; font-weight: 700}
+        td.terminal-api-receipt-value{padding:inherit;}
+        tr.terminal-api-receipt{padding-bottom: 5px;}
+    </style>
+<table><?php $receipt = $this->getMethod()->getInfoInstance()->getAdditionalInformation('receipt'); echo $receipt;?></table>
```
lancergr: do we need `<table>` since there is already one in receipt variable?
Let's also use multiple lines

Helper/Data.php
```diff
@@ -1024,4 +1019,27 @@ public function getAdyenMerchantAccount($method, $storeId)
         return $merchantAccount;
     }
 
+    public function formatTerminalAPIReceipt($paymentReceipt)
+    {
+        $paymentReceipt = json_decode($paymentReceipt);
```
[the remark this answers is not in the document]
Aleffio: $paymentReceipt is initially an array, applying json_encode and then json_decode will change it to an Object: https://stackoverflow.com/questions/4790453/php-recursive-array-to-object/6909087

The alternative would be just to use the array in the function for example: 
on L1027:
$receipt['DocumentQualifier']
instead of
$receipt->DocumentQualifier

I prefer to keep the receipt as object at the moment, but we can change it in another PR.

PR #308, opened 2018-08-28
Feature/pw 582 make is possible to store card

rikterbeek:
**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Model/AdyenInitiateTerminalApi.php
+++ b/Model/AdyenInitiateTerminalApi.php
@@ -59,7 +59,8 @@ class AdyenInitiateTerminalApi implements AdyenInitiateTerminalApiInterface
         \Adyen\Payment\Model\RecurringType $recurringType,
         \Magento\Checkout\Model\Session $_checkoutSession,
         array $data = []
-    ) {
+    )
+    {
         $this->_encryptor = $encryptor;
         $this->_adyenHelper = $adyenHelper;
         $this->_adyenLogger = $adyenLogger;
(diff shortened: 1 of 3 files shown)


Model/AdyenInitiateTerminalApi.php
```diff
@@ -107,6 +109,23 @@ public function initiate()
...
+        // if custom is logged in send data accross
+        $customerId = $quote->getCustomerId();
+
+        $storeCustomer = "";
+        if (!empty($customerId)) {
+            $shopperEmail = $quote->getCustomerEmail();
+            $recurringContract = $this->_adyenHelper->getAdyenPosCloudConfigData('recurring_type');
+
+            $jsonValue = '{
+                "shopperEmail": "' . $shopperEmail . '",
```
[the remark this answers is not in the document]
rikterbeek: Agree then we should change the full json request to an array.

Model/AdyenInitiateTerminalApi.php
```diff
@@ -72,6 +73,7 @@ public function __construct(
         $client = new \Adyen\Client();
         $client->setApplicationName("Magento 2 plugin");
         $client->setXApiKey($apiKey);
+        $client->setInputType('json');
```
lancergr: shouldn't we better implement http://php.net/manual/en/function.is-string.php or http://php.net/manual/en/function.is-array.php on lib instead of explicitly asking the inputType?

rikterbeek: @lancergr agree if you can do a quick PR on the PHP library I would say just do it if you think it cost more time create an issue on youtrack to refine.

Model/AdyenInitiateTerminalApi.php
```diff
@@ -106,47 +106,69 @@ public function initiate()
...
+                                            'TransactionID' => $reference,
+                                            'TimeStamp' => $timeStamper,
+                                        ],
+                                ],
+                            'PaymentTransaction' =>
+                                [
+                                    'AmountsReq' =>
+                                        [
+                                            'Currency' => $quote->getCurrency()->getQuoteCurrencyCode(),
+                                            'RequestedAmount' => $quote->getGrandTotal(),
```
Aleffio: (double)$quote->getGrandTotal() makes sure it is not converted to string by the json_encode

rikterbeek: double does not seems to exists can you check if doubleval() works as well ?
http://php.net/manual/en/function.doubleval.php

Model/AdyenInitiateTerminalApi.php
```diff
@@ -106,47 +106,69 @@ public function initiate()
...
+                ],
+        ];
+
+        // If customer exists add it into the request to store request
+        if (!empty($customerId)) {
+            $shopperEmail = $quote->getCustomerEmail();
+            $recurringContract = $this->_adyenHelper->getAdyenPosCloudConfigData('recurring_type');
+
+            if (!empty($recurringContract)) {
+                $jsonValue = '{
```
lancergr: here we better do array and base64_encode(json_encode($saleToAcquirerData))

PR #316, opened 2018-09-21
remove storeId check so we can show all billing agreements

rikterbeek:
We do not need to filter on StoreID, magento already decide if the customer is allowed to login. If he is allowed we can show them all of the billing agreements this customer has.

**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -773,7 +773,6 @@ class Data extends AbstractHelper
 
         $baCollection = $this->_billingAgreementCollectionFactory->create();
         $baCollection->addFieldToFilter('customer_id', $customerId);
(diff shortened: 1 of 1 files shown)


Helper/Data.php
```diff
@@ -773,7 +773,6 @@ public function getOneClickPaymentMethods($customerId, $storeId, $grandTotal, $r
 
         $baCollection = $this->_billingAgreementCollectionFactory->create();
         $baCollection->addFieldToFilter('customer_id', $customerId);
-        $baCollection->addFieldToFilter('store_id', $storeId);
```
lancergr: can we omit storing store_id at first place?

rikterbeek: Yes we can but why?

PR #318, opened 2018-09-25
Revert "remove storeId check so we can show all billing agreements"

rikterbeek:
Reverts Adyen/adyen-magento2#316


rikterbeek:
Revert back the PR. High risk for current merchants as if they have multiple stores with different merchantAccounts and they do not share contracts on company level ( by default this is the case) we can show contracts of a other merchantAccount causing shoppers to see there card but getting a error using it as it is not connected to that store. What we can do is introduce a setting where you can say if you share contracts on company account (and make clear this needs to be set by a support engineer on the account). For now as there is no merchant request we can leave it for a later moment.

PR #319, opened 2018-09-28
PW-642: Creating credit memos after notification from BO

Aleffio:
**Description**
Uncommenting code after Magento2 core issue #2656 has been resolved.
Now creating Credit Memos after triggering a refund from Adyen Customer Area

**Tested scenarios**
Full amount credit memo successful.
Partial amount credit memo unsuccessful with comment on order:
`IPN "Refunded". Refund issued by merchant. Registered notification about refunded amount of €***. Transaction ID: "". Credit Memo has not been created. Please create offline Credit Memo.`

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -1076,19 +1076,15 @@ class Cron
         $lastTransactionId = $this->_order->getPayment()->getLastTransId();
         if ($lastTransactionId != $this->_pspReference) {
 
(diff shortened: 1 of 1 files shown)


lancergr:
From which version has this been resolved? can we add dependency on composer?

Aleffio:
> From which version has this been resolved? can we add dependency on composer?

It was solved in Magento 2.1.0, we already have a dependency for that

PR #321, opened 2018-10-10
PW-382 Set IssuerId in checkout

cyattilakiss:
**Description**
Remove IDEAL check when filtering the available payment methods
Let other payment methods also use the issuer drop down list

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
@@ -152,7 +152,7 @@ define(
                         fullScreenLoader.stopLoader();
                     }
                 ).fail(function (error) {
-                    console.log(JSON.stringify(error));
+                    //console.log(JSON.stringify(error));
                     fullScreenLoader.stopLoader();
                 });
             },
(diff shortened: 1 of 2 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -152,7 +152,7 @@ define(
                         fullScreenLoader.stopLoader();
                     }
                 ).fail(function (error) {
-                    console.log(JSON.stringify(error));
+                    //console.log(JSON.stringify(error));
```
rikterbeek: let's remove this line at all.

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -193,10 +193,22 @@ define(
...
+
+                    if (value.hasOwnProperty("issuers")) {
+                        if (value.issuers.length == 0) {
+                            return false;
+                        }
+
                         result.issuerIds = value.issuers;
                         result.issuerId = ko.observable(null);
-                    } else if (value.isPaymentMethodOpenInvoiceMethod) {
+                    }else if (value.isPaymentMethodOpenInvoiceMethod) {
```
rikterbeek: what formatter do you use? Think we can keep this space.

cyattilakiss: Nothing extraordinary I usually follow how they style the js code:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/if...else
If we have a common setting I can set it up 👍

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
cyattilakiss: agree

PR #325, opened 2018-10-23
customer reference prefix

stkams:
**Description**

Optional prefix for the customer reference
Removed 5 minutes delay
Get billing agreements across stores

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1151,15 +1151,16 @@ class Data extends AbstractHelper
         return $formFields;
     }
 
-    public function getApiKey()
+    /**
+     * @param integer|null $storeId
+     * @return string the X API Key for the specified or current store
+     */
+    public function getApiKey($storeId = null)
     {
-        switch ($this->isDemoMode()) {
-            case true:
-                $apiKey = $this->_encryptor->decrypt(trim($this->getAdyenPosCloudConfigData('api_key_test')));
-                break;
-            default:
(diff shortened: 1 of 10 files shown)


Helper/Data.php
```diff
@@ -793,7 +793,7 @@ public function getOneClickPaymentMethods($customerId, $storeId, $grandTotal, $r
 
         $baCollection = $this->_billingAgreementCollectionFactory->create();
         $baCollection->addFieldToFilter('customer_id', $customerId);
-        $baCollection->addFieldToFilter('store_id', $storeId);
+        // $baCollection->addFieldToFilter('store_id', $storeId);
```
cyattilakiss: instead of commenting out if won't be used in the future delete it I think

Model/Cron.php
```diff
@@ -930,7 +914,8 @@ protected function _processNotification()
                 $recurringDetailReference = $this->_pspReference;
 
                 $storeId = $this->_order->getStoreId();
-                $customerReference = $this->_order->getCustomerId();
+                $customerId = $this->_order->getCustomerId();
+                $customerReference = $this->_adyenHelper->getCustomerReference($customerId);
```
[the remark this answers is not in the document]
stkams: done

Helper/Data.php
stkams: Done

stkams:
Now also updated the module version to 2.3.3

PR #326, opened 2018-10-25
PW-306 Secure Fields integration for card payments

cyattilakiss:
**Description**
- Add admin message to add the API key instead of the username and password
- Remove username, password (test, live) from admin/store configuration/
required settings. Add API Key instead
- Get secure fields SDK based on isDemo
- Include checkout js components
- Remove CSE library and tokens


AdminMessage/APIKeyMessage.php
```diff
@@ -0,0 +1,149 @@
...
+
+	/**
+	 * Check whether the system message should be shown
+	 *
+	 * @return bool
+	 */
+	public function isDisplayed()
+	{
+		// Only execute the query the first time you access the Admin page
+		if ($this->authSession->isFirstPageAfterLogin() && $this->_adyenHelper->getWsUsername()) {
```
lancergr:
```suggestion
		if ($this->authSession->isFirstPageAfterLogin() && empty($this->_adyenHelper->getAPIKey() ) {
```

lancergr: and we can always return true on L119

AdminMessage/APIKeyMessage.php
```diff
@@ -0,0 +1,149 @@
...
+			try {
+				$title = "Adyen extension requires the API KEY!";
+				$messageData[] = array(
+					'severity' => $this->getSeverity(),
+					'date_added' => date("Y-m-d"),
+					'title' => $title,
+					'description' => "Please provide API-KEY for the webserver user " . $this->_adyenHelper->getWsUsername() . "  for default/store " . $this->storeManagerInterface->getStore()->getName(),
+					'url' => "https://docs.adyen.com/developers/plug-ins-and-partners/magento-2/set-up-the-plugin-in-magento#step3configuretheplugininmagento",
```
rkewlani: minor: webservice user instead of webserver user.

Gateway/Http/Client/TransactionPayment.php
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Class TransactionSale
+ */
+class TransactionPayment implements ClientInterface
+{
+
+	/**
+	 * @var \Adyen\Client
+	 */
+	protected $client;
```
lancergr: performance-wise I believe it would be better to call initializeAdyenClient within placeRequest and remove this property

Helper/Data.php
```diff
@@ -1196,4 +1193,62 @@ public function initializeAdyenClient($storeId = null)
...
+			"originDomains" => array(
+				$url
+			)
+		);
+
+		$client = $this->initializeAdyenClient();
+
+		$service = $this->adyenServiceFactory->createCheckoutUtility($client);
+		$respone = $service->originKeys($params);
+		return $respone['originKeys'];
```
lancergr: For a given url there should be one originKey, so we better return the orginKey for this domain instead of an array.
This will simplify the logic in L1212
```suggestion
		return current($respone['originKeys']);
```

view/frontend/web/js/adyen.checkout.js
```diff
@@ -0,0 +1,287 @@
+var AdyenCheckout = (function () {
```
lancergr: unused file?

cyattilakiss: yes - gonna remove it

Issue #327, opened 2018-11-01
adyen/process/json incompatible with Magento 2.3


AntonEvers:
**Magento version**: 2.3-develop
**Plugin version**: 2.3.3
**Description**

POST calls to adyen/process/json will return a 302 redirect because the `CsrfAwareActionInterface` is not implemented and CSRF validation fails. This is new functionality introduced in the 2.3-develop version of Magento.

A request is marked as invalid if the request:

1. Comes from an external domain
2. Is a POST request
3. Doesn't contain the parameter "isAjax"

PR is on the way.

Probably this will require a version bump that signifies backward incompatible changes.

rikterbeek:
Hi @AntonEvers,

Thanks for letting us know. This would have a big impact indeed to all our customers upgrading. Is there not a other way to solve this issue. If you can do a PR that would be great.

Regards,
Rik
Adyen

kzobov:
Hi @AntonEvers,
Can you please share the planned release ETA?
This has blocked our project. 

Thank you.

NikZar:
Is this issue still open? 
I have to choose between a Magento 2.2 or the 2.3 version and I need the Adyen plugin to work well in my installation...

AntonEvers:
Fixed with https://github.com/Adyen/adyen-magento2/commit/9589d4a595110253e6d7c499a1a74bece3de56de

wintermute-84:
Getting 302 redirects with adyen module-payment 4.2.1 and magento version 2.3.2
The constructor fix is in place.

For now fixed this by making Json controller implement CsrfAwareActionInterface


here is a full patched file:
```
<?php
/**
 *                       ######
 *                       ######
 * ############    ####( ######  #####. ######  ############   ############
 * #############  #####( ######  #####. ######  #############  #############
 *        ######  #####( ######  #####. ######  #####  ######  #####  ######
 * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
 * ###### ######  #####( ######  #####. ######  #####          #####  ######
 * #############  #############  #############  #############  #####  ######
 *  ############   ############  #############   ############  #####  ######
 *                                      ######
 *                               #############
 *                               ############
 *
 * Adyen Payment module (https://www.adyen.com/)
 *
 * Copyright (c) 2015 Adyen BV (https://www.adyen.com/)
 * See LICENSE.txt for license details.
 *
 * Author: Adyen <magento@adyen.com>
 */

[...]

PR #328, opened 2018-11-01
Make adyen/process/json compatible with Magento 2.3-develop

AntonEvers:
**Description**

CSRF checks have been introduced on all requests in Magento 2.3. URLs that function as webhooks do not require these checks however. I have applied the same approach here as is used in other built in payment methods in M2.3

**Tested scenarios**

Click "test configuration" when testing the server communication settings (standard notification) in the Adyen portal.

Before PR: 302
After PR: 200 [accepted]

**Fixed issue**:  #327

### DIFF
--- a/Controller/Process/Json.php
+++ b/Controller/Process/Json.php
@@ -20,6 +20,7 @@
  *
  * Author: Adyen <magento@adyen.com>
  */
+declare(strict_types=1);
 
 namespace Adyen\Payment\Controller\Process;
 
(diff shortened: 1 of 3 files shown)


cyattilakiss:
Hi @AntonEvers ,

Thanks for the PR we will look into this.

Regards,
Attila
Adyen

rikterbeek:
Hi @AntonEvers,

Thanks for the PR. We found a better solution so we can be for now backwards compatible see:
#357

Regards,
Rik
Adyen

AntonEvers:
@rikterbeek excellent!

PR #329, opened 2018-11-06
PW-708 Add live_endpoint_url_prefix to support checkout live endpoints

cyattilakiss:
**Description**
Add live endpoint prefix to admin settings
Add live endpoint prefix to initialize client

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -646,11 +646,7 @@ class Data extends AbstractHelper
 	 */
 	public function getLiveEndpointPrefix($storeId = null)
 	{
-		if ($this->isDemoMode($storeId)) {
-			$prefix = "";
-		} else {
-			$prefix = trim($this->getAdyenAbstractConfigData('live_endpoint_url_prefix', $storeId));
-		}
+		$prefix = trim($this->getAdyenAbstractConfigData('live_endpoint_url_prefix', $storeId));
 		return $prefix;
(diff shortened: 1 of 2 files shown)


Helper/Data.php
```diff
@@ -635,6 +638,22 @@ public function getWsUsername($storeId = null)
 		return $wsUsername;
 	}
 
+	/**
+	 * @desc Retrieve the Live endpoint prefix key
+	 *
+	 * @param int|null $storeId
+	 * @return string
+	 */
+	public function getLiveEndpointPrefix($storeId = null)
```
lancergr: minor: shouldn't getLiveEndpointPrefix return always live_endpoint_url_prefix even if store is in demo mode?
This logic exists in L1187 anyway

etc/adminhtml/system/adyen_required_settings.xml
```diff
@@ -61,6 +61,13 @@
             <backend_model>Magento\Config\Model\Config\Backend\Encrypted</backend_model>
             <config_path>payment/adyen_abstract/api_key_live</config_path>
         </field>
+        <field id="live_endpoint_url_prefix" translate="label" type="obscure" sortOrder="70" showInDefault="1" showInWebsite="1" showInStore="0">
```
Aleffio: type="text"

PR #333, opened 2018-11-07
PW-720: add Descriptor for Ratepay

Aleffio:
**Description**
Added Descriptor for Ratepay, it is now set after an AUTHORISATION notification inside additional_information as adyen_ratepay_descriptor

**Tested scenarios**
Ratepay transaction

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -122,6 +122,11 @@ class Cron
     /**
      * @var
      */
+    protected $ratepayDescriptor;
+
+    /**
+     * @var
+     */
     protected $_eventCode;
 
     /**
@@ -531,6 +536,10 @@ class Cron
             if ($acquirerReference != "") {
                 $this->_acquirerReference = $acquirerReference;
             }
+            $ratepayDescriptor = isset($additionalData['openinvoicedata.descriptor']) ? $additionalData['openinvoicedata.descriptor'] : null;
+            if ($ratepayDescriptor != "") {
+                $this->ratepayDescriptor = $ratepayDescriptor;
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


Model/Cron.php
```diff
@@ -531,6 +536,10 @@ protected function _declareVariables($notification)
             if ($acquirerReference != "") {
                 $this->_acquirerReference = $acquirerReference;
             }
+            $ratepayDescriptor = isset($additionalData['openinvoicedata.descriptor']) ? $additionalData['openinvoicedata.descriptor'] : null;
```
cyattilakiss:
```suggestion
            $ratepayDescriptor = isset($additionalData['openinvoicedata.descriptor']) ? $additionalData['openinvoicedata.descriptor'] : "";
```

if you do this you can perform a strict check in the next line for the empty string - maybe it's a better approach

Model/Cron.php
```diff
@@ -531,6 +536,10 @@ protected function _declareVariables($notification)
...
+            if ($ratepayDescriptor != "") {
```
cyattilakiss:
```suggestion
            if ($ratepayDescriptor !== "") {
```

like this

Model/Cron.php
```diff
@@ -730,6 +739,12 @@ protected function _updateOrderPaymentWithAdyenAttributes($additionalData)
         if (!empty($expiryDate)) {
             $this->_order->getPayment()->setAdditionalInformation('adyen_expiry_date', $expiryDate);
         }
+        if ($this->ratepayDescriptor != "") {
```
cyattilakiss: here also

```suggestion
        if ($this->ratepayDescriptor !== "") {
```

Model/Cron.php
```diff
@@ -119,6 +119,11 @@ class Cron
      */
     protected $_acquirerReference;
 
+    /**
+     * @var
+     */
+    protected $ratepayDescriptor;
```
lancergr: minor: shall we keep the convention of underscore ($_ratepayDescriptor) ?

rikterbeek: Let's use the PSR-2: Coding Style Guide and there it is specific mentioned to not using the underscore anymore see: https://www.php-fig.org/psr/psr-2/


quote:
```
Property names SHOULD NOT be prefixed with a single underscore to indicate protected or private visibility.
```

rikterbeek:
@Aleffio please add it as well into the payment details view.

Aleffio:
> @Aleffio please add it as well into the payment details view.

Added on Payment Information under adminhtml page

PR #332, opened 2018-11-07
PW-615 Separate Initiate/Status flow for TerminalAPI

Aleffio:
**Description**
Gateway Client now returns a **PaymentResponse** after **Initiate** call or **Status** call.
Check for timeout through the proper error code: _CURLE_OPERATION_TIMEOUTED_

PR #337, opened 2018-11-12
Re-merging "PW-615" and "customer reference prefix"

Aleffio:
4 commits were lost during the rebase on CloudAPI, add them again to the CloudAPI branch:
https://github.com/Adyen/adyen-magento2/pull/332
https://github.com/Adyen/adyen-magento2/pull/325

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -825,7 +825,6 @@ class Data extends AbstractHelper
 
         $baCollection = $this->_billingAgreementCollectionFactory->create();
         $baCollection->addFieldToFilter('customer_id', $customerId);
-        $baCollection->addFieldToFilter('store_id', $storeId);
         $baCollection->addFieldToFilter('method_code', 'adyen_oneclick');
         $baCollection->addActiveFilter();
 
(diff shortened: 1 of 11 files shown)


Model/Cron.php
```diff
@@ -300,7 +300,7 @@ public function execute()
         $dateStart = new \DateTime();
         $dateStart->modify('-5 day');
         $dateEnd = new \DateTime();
-        $dateEnd->modify('-1 minute');
```
lancergr: why is this different from https://github.com/Adyen/adyen-magento2/blob/develop/Model/Cron.php#L303 ?

Helper/Data.php
```diff
@@ -825,7 +825,6 @@ public function getOneClickPaymentMethods($customerId, $storeId, $grandTotal, $r
 
         $baCollection = $this->_billingAgreementCollectionFactory->create();
         $baCollection->addFieldToFilter('customer_id', $customerId);
-        $baCollection->addFieldToFilter('store_id', $storeId);
```
lancergr: why is this removed ( https://github.com/Adyen/adyen-magento2/blob/develop/Helper/Data.php#L808 )?

PR #338, opened 2018-11-13
Cloud API optimizations

Aleffio:
**Description**
- Added phpdoc where missing
- Added Storeid
- Use array instead of jsondecode(jsonencode()) for the receipt

### DIFF
--- a/Gateway/Http/Client/TransactionPosCloudSync.php
+++ b/Gateway/Http/Client/TransactionPosCloudSync.php
@@ -30,6 +30,10 @@ use Magento\Payment\Gateway\Http\ClientInterface;
 
 class TransactionPosCloudSync implements ClientInterface
 {
+    /**
+     * @var int
+     */
+    protected $storeId;
 
     public function __construct(
         \Magento\Framework\Model\Context $context,
@@ -37,6 +41,7 @@ class TransactionPosCloudSync implements ClientInterface
         \Adyen\Payment\Helper\Data $adyenHelper,
         \Adyen\Payment\Logger\AdyenLogger $adyenLogger,
         \Adyen\Payment\Model\RecurringType $recurringType,
+        \Magento\Store\Model\StoreManagerInterface $storeManager,
         array $data = []
     ) {
         $this->_encryptor = $encryptor;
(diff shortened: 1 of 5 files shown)


Gateway/Http/Client/TransactionPosCloudSync.php
```diff
@@ -30,36 +30,41 @@
...
         \Adyen\Payment\Model\RecurringType $recurringType,
+        \Magento\Store\Model\StoreManagerInterface $storeManager,
         array $data = []
     ) {
         $this->_encryptor = $encryptor;
         $this->_adyenHelper = $adyenHelper;
         $this->_adyenLogger = $adyenLogger;
         $this->_recurringType = $recurringType;
         $this->_appState = $context->getAppState();
+        $this->storeManager = $storeManager;
```
cyattilakiss: please create a property for the storemanager also

rikterbeek: do we need to expose the storeManager. As it is only used for storeId we can just do: 
```
this->storeId = $storeManager->getStore()->getId();
```

Helper/Data.php
```diff
@@ -1188,58 +1188,79 @@ public function getOpenInvoiceLineData(
...
+     * If it is not configured for the specific payment method,
+     * return the merchant account name defined in required settings.
+     *
+     * @param $paymentMethod
+     * @param int|null $storeId
+     * @return string
+     */
+    public function getAdyenMerchantAccount($paymentMethod, $storeId = null)
     {
         $merchantAccount = $this->getAdyenAbstractConfigData("merchant_account", $storeId);
```
rikterbeek: as storeId can be null should we not retrieve it here or not make sure storeId needs to have an value ?

Model/AdyenInitiateTerminalApi.php
```diff
@@ -37,43 +37,52 @@ class AdyenInitiateTerminalApi implements AdyenInitiateTerminalApiInterface
...
         \Magento\Checkout\Model\Session $_checkoutSession,
+        \Magento\Store\Model\StoreManagerInterface $storeManager,
         array $data = []
     )
     {
         $this->_adyenHelper = $adyenHelper;
         $this->_adyenLogger = $adyenLogger;
         $this->_checkoutSession = $_checkoutSession;
+        $this->storeManager = $storeManager;
+        $this->storeId = $this->storeManager->getStore()->getId();
```
rikterbeek: We can simplify this with
```
this->storeId = $storeManager->getStore()->getId();
```
so we do not have to expose storeManager variable.

Helper/Data.php
Aleffio: I will revert it to mandatory, as in all the places where we call it, it is specified

PR #345, opened 2018-11-21
PW-704 Display list of APMs in Magento frontend using /paymentMethods

cyattilakiss:
**Description**
Using paymentMethods instead of directory lookup and some optimizations

### DIFF
--- a/Helper/PaymentMethods.php
+++ b/Helper/PaymentMethods.php
@@ -33,72 +33,60 @@ class PaymentMethods extends AbstractHelper
     /**
      * @var \Magento\Quote\Api\CartRepositoryInterface
      */
-    protected $_quoteRepository;
-
-    /**
-     * @var \Magento\Quote\Model\QuoteIdMaskFactory
-     */
-    protected $_quoteIdMaskFactory;
+    protected $quoteRepository;
 
     /**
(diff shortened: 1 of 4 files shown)


Helper/PaymentMethods.php
```diff
@@ -326,46 +310,20 @@ protected function _getCurrentCountryCode($store, $country)
         return "";
     }
 
-    /**
-     * @param $store
-     * @return mixed|string
-     */
-    protected function _getCurrentLocaleCode($store)
```
maassenbas: Why exclude the shopperLocale from the request to /paymentMethods?

cyattilakiss: We can pass it as an optional parameter but in the API explorer the example excluded it so I assumed it's not necessary anymore

Helper/PaymentMethods.php
```diff
@@ -188,54 +173,53 @@ protected function _addHppMethodsToConfig($store, $country)
      * @param $country
      * @return array
      */
-    protected function _fetchHppMethods($store, $country)
+    protected function fetchHppMethods($store, $country)
```
lancergr: a better name would be "fetchAlternativeMethods" since is not using hpp anymore

PR #347, opened 2018-11-23
PW-685 Route simple APMs via /payments

cyattilakiss:
**Description**
Route HPP / APMs through checkout /payments


Block/Redirect/Redirect.php
```diff
@@ -102,6 +102,50 @@ public function __construct(
...
+				if ($redirectUrl = $paymentObject->getAdditionalInformation('redirectUrl')) {
+					return $redirectUrl;
+				}
+			}
+		} catch (Exception $e) {
+			// do nothing for now
+			throw($e);
+		}
+
+		return "";
```
rikterbeek: This should throw an exceptions I believe as empty value will break.

Gateway/Validator/CheckoutResponseValidator.php
```diff
@@ -95,23 +95,59 @@ public function validate(array $validationSubject)
...
+					if (isset($response['redirect']['url']) && !empty($response['redirect']['url'])) {
+						$redirectUrl = $response['redirect']['url'];
+					}
+
+					if (!empty($response['paymentData'])) {
+						$paymentData = $response['paymentData'];
+					}
+
+					// If the redirect data is there then the payment is a card payment with 3d secure
+					if (isset($response['redirect']['data'])) {
```
rikterbeek: Shall we be more specific here. Check if $response['redirect']['data']['PaReq'] && $response['redirect']['data']['MD']) are set move to this logic otherwise go into alternative route.

Gateway/Validator/CheckoutResponseValidator.php
```diff
@@ -95,23 +95,59 @@ public function validate(array $validationSubject)
...
 					} else {
-						$isValid = false;
-						$errorMsg = __('3D secure is not valid.');
-						$this->adyenLogger->error($errorMsg);;
-						$errorMessages[] = $errorMsg;
+						$payment->setAdditionalInformation('CheckoutAPM', true);
+
+						if ($redirectUrl && $paymentData) {
+							$payment->setAdditionalInformation('redirectUrl', $redirectUrl);
+							$payment->setAdditionalInformation('paymentData', $paymentData);
```
rikterbeek: do we clear paymentData after usage ?

Block/Redirect/Redirect.php
```diff
@@ -102,6 +102,50 @@ public function __construct(
...
 
+	/**
+	 * Returns if the payment should follow the old HPP or the new Checkout flow
+	 *  - hpp will submit a form with all the additional information that the API requests\
+	 *  - checkout will redirect to a url without form submission
+	 *
+	 * @return bool
+	 * @throws \Exception
+	 */
+    public function isCheckoutAPM() {
```
rikterbeek: let's make sure that all alternative payment methods are routed throught the HPP. Separate task to route this to checkout.

Gateway/Validator/CheckoutResponseValidator.php
cyattilakiss: Should I do it with the paReq and MD as well?

Controller/Process/Result.php
```diff
@@ -279,31 +279,37 @@ protected function _validateUpdateOrder($order, $response)
...
-        $result = [];
-        $pairs = explode("&", $queryString);
-
-        foreach ($pairs as $pair) {
-            $nv = explode("=", $pair);
-            $name = urldecode($nv[0]);
-            $value = urldecode($nv[1]);
-            $result[$name] = $value;
-        }
+    	if (!empty($response['merchantSig'])) {
```
lancergr: Why not using different urls when in checkout and when in hpp?
This way we can have different functions and clearer method signatures instead of if-else statements.

cyattilakiss: Hopefully we are going to depricate HPP in the near future so I tried to solve it with the less effort it can be. :)

PR #349, opened 2018-11-28
PW-705 Handle return from Checkout API redirect methods including payload

cyattilakiss:
**Description**
Handle response from Checkout API redirect

### DIFF
--- a/Controller/Process/Result.php
+++ b/Controller/Process/Result.php
@@ -57,6 +57,11 @@ class Result extends \Magento\Framework\App\Action\Action
      */
     protected $_adyenLogger;
 
+	/**
+	 * @var \Magento\Store\Model\StoreManagerInterface
+	 */
+	protected $storeManager;
+
     /**
      * Result constructor.
      *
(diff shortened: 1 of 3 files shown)


Controller/Process/Result.php
```diff
@@ -217,7 +237,7 @@ protected function _validateUpdateOrder($order, $response)
         // needed because then we need to save $order objects
         $order->setAdyenResulturlEventCode($authResult);
 
-        switch ($authResult) {
+        switch (strtoupper($authResult)) {
```
lancergr: "Received" is a code that can be expected from checkout api but not from hpp so that's why is not already here. Maybe is good idea to add it on this switch condition.

Controller/Process/Result.php
```diff
@@ -204,7 +219,12 @@ protected function _validateUpdateOrder($order, $response)
 
         $this->_adyenLogger->addAdyenResult('Updating the order');
 
-        $authResult = $response['authResult'];
+        if (!empty($response['authResult'])) {
+			$authResult = $response['authResult'];
+		} elseif (!empty($response['authResponse'])) {
+			$authResult = $response['authResponse'];
```
lancergr: authResponse should have been deprecated in v40, so better use "resultCode"

Controller/Process/Result.php
cyattilakiss: Add the case

PR #365, opened 2018-12-21
The UCD needs to get the customer ID via a different way

stkams:
**Description**

The UCD needs to retrieve the customer ID in a different way

**Tested scenarios**

**Fixed issue**:


Model/AdyenInitiateTerminalApi.php
```diff
@@ -188,4 +190,13 @@ public function initiate()
         $quote->save();
         return $response;
     }
+
+    /**
```
cyattilakiss:
```suggestion
     /**
       * Used for a plugin, please use this function and don't remove it
```

rikterbeek:
@cyattilakiss let's document it a bit better so we know where it is being used for.

PR #368, opened 2018-12-28
Store billing agreement after Terminal API transaction

rikterbeek:
**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Gateway/Response/PaymentPosCloudHandler.php
+++ b/Gateway/Response/PaymentPosCloudHandler.php
@@ -40,7 +40,7 @@ class PaymentPosCloudHandler implements HandlerInterface
     private $adyenLogger;
 
     /**
-     * PaymentDataBuilder constructor.
+     * PaymentPosCloudHandler constructor.
      *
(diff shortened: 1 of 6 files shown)


Gateway/Response/PaymentPosCloudHandler.php
```diff
@@ -74,6 +74,42 @@ public function handle(array $handlingSubject, array $paymentResponse)
...
+                $nv = explode('=', $pair);
+
+                if ($nv[0] == 'recurring.recurringDetailReference') {
+                    $recurringDetailReference = $nv[1];
+                    break;
+                }
+            }
+
+            if (!empty($recurringDetailReference) &&
+                !empty($paymentResponse['PaymentResult']['PaymentInstrumentData']['CardData'])
```
cyattilakiss: be careful because if any of the higher level keys in ```$paymentResponse['PaymentResult']['PaymentInstrumentData']['CardData']``` doesn't exist (PaymentResult or PaymentInstrumentData) then it will generate a Notice or a Warning. In this case use it with isset.

PR #373, opened 2019-01-08
PW-706 Handle iDeal payments using Checkout component

cyattilakiss:
**Description**
Add iDeal component

### DIFF
--- a/Model/Ui/AdyenHppConfigProvider.php
+++ b/Model/Ui/AdyenHppConfigProvider.php
@@ -134,8 +134,6 @@ class AdyenHppConfigProvider implements ConfigProviderInterface
             }
         }
 
-		$config['payment']['adyenHpp']['originKey'] = $this->adyenHelper->getOriginKeyForBaseUrl();
-		$config['payment']['adyenHpp']['checkoutUrl'] = $this->adyenHelper->getCheckoutContextUrl($this->storeManager->getStore()->getId());
 		$config['payment']['adyenHpp']['locale'] = $this->adyenHelper->getStoreLocale($this->storeManager->getStore()->getId());
 
         // add to config
--- a/Model/Ui/AdyenHppConfigProvider.php
+++ b/Model/Ui/AdyenHppConfigProvider.php
@@ -35,34 +35,39 @@ class AdyenHppConfigProvider implements ConfigProviderInterface
     /**
      * @var PaymentHelper
      */
-    protected $_paymentHelper;
+    protected $paymentHelper;
 
     /**
      * @var \Adyen\Payment\Helper\Data
      */
-    protected $_adyenHelper;
+    protected $adyenHelper;
 
     /**
      * Request object
      *
      * @var \Magento\Framework\App\RequestInterface
      */
-    protected $_request;
+    protected $request;
 
     /**
      * @var \Magento\Framework\UrlInterface
      */
-    protected $_urlBuilder;
+    protected $urlBuilder;
 
     /**
      * @var \Magento\Customer\Model\Session
      */
-    protected $_customerSession;
+    protected $customerSession;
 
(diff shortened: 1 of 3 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -199,6 +199,64 @@ define(
...
+                     * Renders the secure fields,
+                     * creates the ideal component,
+                     * sets up the callbacks for ideal components and
+                     */
+                    result.renderIdealComponent = function () {
+
+                        self.isPlaceOrderActionAllowed(false);
+
+                        var secureFieldsNode = document.getElementById('iDealContainer');
+
```
rkewlani: Minor: seeing empty lines at some places.

Model/Ui/AdyenHppConfigProvider.php
```diff
@@ -114,19 +121,23 @@ public function getConfig()
...
             // format to calendar date
-            $dob = $this->_customerSession->getCustomerData()->getDob();
+            $dob = $this->customerSession->getCustomerData()->getDob();
             if ($dob) {
                 $dob = strtotime($dob);
                 $dob = date('m/d/Y', $dob);
             }
         }
 
+		$config['payment']['adyenHpp']['originKey'] = $this->adyenHelper->getOriginKeyForBaseUrl();
```
rkewlani: I don't think origin key is needed here. Can you please double check?

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -199,6 +199,64 @@ define(
...
+                            originKey: self.getOriginKey(),
+                            loadingContext: self.getLoadingContext(),
+                            items: result.getIssuerListForComponent(),
+                            onChange: function (state) {
+                                // isValid is not present on start
+                                if (typeof state.isValid !== 'undefined' && state.isValid === false) {
+                                    self.isPlaceOrderActionAllowed(false);
+                                }
+                            },
+                            onValid: function (state) {
```
rkewlani: We can remove onValid method. onChange will take care of state change.

Model/Ui/AdyenHppConfigProvider.php
cyattilakiss: Checked it, indeed is working without them. Going to remove them.

rikterbeek: originKey is only for secured fields/card component indeed

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -199,6 +199,64 @@ define(
 
                         return false;
                     };
+                    // Can be removed after checkout api feature branch goes live since the issuerId key is changed to
+                    // id there and just use the value.issuers in the component
```
rikterbeek: what do you mean by: checkout api feature branch goes live ?

cyattilakiss: I need this "converter" because this branch was based on the secure-fields-integration feature branch (now rebased to develop) where I still used the directory lookup which returned issuerIds instead of just ids. In the checkout-api feature branch since I will rebase that after we merged this to develop I won't need it anymore since there we are going to work with the checkout api that returns the correct keys.

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
cyattilakiss: right now I'll keep as it is since at the cc and oneclick the implementation is the same

PR #377, opened 2019-01-15
[PW-932] Fix maestro oneclick to allow order button and fix to keep the cc and…

rikterbeek:
… oneclick authorisation seperate to provide better structure what is used for what purpose

**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Gateway/Request/OneclickAuthorizationDataBuilder.php
+++ b/Gateway/Request/OneclickAuthorizationDataBuilder.php
@@ -55,6 +55,18 @@ class OneclickAuthorizationDataBuilder implements BuilderInterface
         $paymentDataObject = \Magento\Payment\Gateway\Helper\SubjectReader::readPayment($buildSubject);
         $payment = $paymentDataObject->getPayment();
 
+        // If ccType is set use this. For bcmc you need bcmc otherwise it will fail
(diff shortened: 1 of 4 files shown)


Gateway/Request/OneclickAuthorizationDataBuilder.php
```diff
@@ -79,6 +91,13 @@ public function build(array $buildSubject)
             }
         }
 
+        // if installments is set add it into the request
+        if ($payment->getAdditionalInformation(AdyenOneclickDataAssignObserver::NUMBER_OF_INSTALLMENTS) &&
+            $payment->getAdditionalInformation(AdyenOneclickDataAssignObserver::NUMBER_OF_INSTALLMENTS) > 0
```
cyattilakiss: duplicated lines?

rikterbeek: copied it over from ccAuthorisationBuilder. We could probably simplify it although it could be that > 0 could lead to an notice not sure.

cyattilakiss: okay keep it as it is for now

PR #380, opened 2019-01-17
PW-752 Move apple pay to checkout

cyattilakiss:
**Description**
Move apple pay to checkout


Setup/UpgradeData.php
```diff
@@ -23,6 +23,7 @@
 
 namespace Adyen\Payment\Setup;
 
+use Magento\Framework\App\Config\ReinitableConfigInterface;
```
Aleffio: duplicate

PR #382, opened 2019-01-21
Extend customer data with telephone number

cyattilakiss:
**Description**
Add customer telephone number as a customer data


Gateway/Request/CustomerDataBuilder.php
```diff
@@ -52,7 +52,8 @@ public function build(array $buildSubject)
             $result['shopperReference'] = $customerId;
         }
 
-        $result ['shopperEmail'] = $customerEmail;
+		$result['telephoneNumber'] = trim($billingAddress->getTelephone());
+        $result['shopperEmail'] = $customerEmail;
```
rikterbeek: let's add some defending programming so check if billingAddress is present en if telephone number is not null as this could be stripped out of the checkout. Is the trim needed ?

cyattilakiss: Okay, will do that.
The trim is necessary, I tested it.

PR #385, opened 2019-01-25
Update RefundDataBuilder.php

remco-schouten:
Should refunditem->getQty be casted to int like in capturedatabuilder? this to prevent decimal value in request and server producing an error?

**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Gateway/Request/RefundDataBuilder.php
+++ b/Gateway/Request/RefundDataBuilder.php
@@ -191,7 +191,7 @@ class RefundDataBuilder implements BuilderInterface
 
         foreach ($creditMemo->getAllItems() as $refundItem) {
             ++$count;
-            $numberOfItems = $refundItem->getQty();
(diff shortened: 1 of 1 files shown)


cyattilakiss:
Hi @remco-schouten ,

Thank you for the pull request. We will look into it, but after a quick check I think we are going to merge it. As you pointed out, in the captureDataBuilder we already cast it to integer and also in the Adyen documentation ( https://docs.adyen.com/developers/api-reference/payments-api/paymentrequest/paymentrequest-additionaldata/open-invoice-fields#openinvoicedata ) it is required to be integer.
Best,
Attila
Adyen

cyattilakiss:
Hi @remco-schouten ,

Thank you for the contribution. In our next release your change is also going to be included.

Best,
Attila
Adyen

PR #387, opened 2019-01-28
PW-930 finalise checkout component

cyattilakiss:
**Description**
Style holder name input field
Shareable checkout component for oneclick
Move onValid and onError to OnChange

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
@@ -89,7 +89,6 @@ define(
                     return;
                 }
 
-
                 installments.setInstallments(0);
 
                 // installments enabled ??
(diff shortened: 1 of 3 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -109,8 +108,16 @@ define(
                     groupTypes: self.getAvailableCardTypeAltCodes(),
 
                     onChange: function (state) {
-                        // isValid is not present on start
-                        if (typeof state.isValid !== 'undefined' && state.isValid === false) {
+                        if (!!state.isValid) {
```
rikterbeek: why !! ?

view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
```diff
@@ -100,6 +100,20 @@ define(
             getAdyenBillingAgreements: function () {
                 var self = this;
 
+                // shareable adyen checkout component
+                var checkout = new AdyenCheckout({
```
rikterbeek: how is this shared ?

rikterbeek: never mind is clear. Agree better like this.

cyattilakiss: All the hpp payment methods that we render in the `_.map(window.checkoutConfig.payment.adyenOneclick.billingAgreements, function (value) {` van use this component

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
cyattilakiss: With this check we can do the same as we did previously with the `typeof state.isValid !== 'undefined' && state.isValid === false`

cyattilakiss: TBH I learned it from the checkout guys, tested it and really does the same

PR #391, opened 2019-01-31
#PW-756

AlexandrosMor:
Move boleto to checkout api


ScottPowell14:
HibW Ian

Gateway/Request/CheckoutDataBuilder.php
```diff
@@ -25,188 +25,236 @@
...
+    {
+        /** @var \Magento\Payment\Gateway\Data\PaymentDataObject $paymentDataObject */
+        $paymentDataObject = \Magento\Payment\Gateway\Helper\SubjectReader::readPayment($buildSubject);
+        $payment = $paymentDataObject->getPayment();
+        $order = $payment->getOrder();
+        $storeId = $order->getStoreId();
+        $request = [];
+
+        // do not send email
+        $order->setCanSendNewEmailFlag(false);
```
rikterbeek: for Boleto we want to send out email after order creation. THis will provide them the PDF so they can go to there bank.

etc/di.xml
```diff
@@ -633,16 +634,16 @@
     <virtualType name="AdyenPaymentCcResponseHandlerComposite" type="Magento\Payment\Gateway\Response\HandlerChain">
         <arguments>
             <argument name="handlers" xsi:type="array">
-                <item name="payment_details" xsi:type="string">Adyen\Payment\Gateway\Response\CheckoutPaymentsDetailsHandler</item>
+                <item name="payment_details" xsi:type="string">Adyen\Payment\Gateway\Response\CheckoutPaymentsDetailsHandler</item>*24
```
rikterbeek: where is the *24 for ?

Gateway/Request/CheckoutDataBuilder.php
```diff
@@ -25,188 +25,236 @@
...
+                    date("s"),
+                    date("m"),
+                    date("j") + $deliveryDays,
+                    date("Y")
+                )
+            );
+
+            $request['deliveryDate'] = $deliveryDate;
+        }
+
```
rikterbeek: better to move to the boletoDataBuilder as this is still a seperate payment method or not ?

rikterbeek: I think we can put it in AdyenPaymentBoletoAuthorizeRequest as we load this before for boleto specifics.

Gateway/Request/CheckoutDataBuilder.php
```diff
@@ -25,188 +25,236 @@
...
+     * @param \Magento\Store\Model\StoreManagerInterface $storeManager
+     * @param \Magento\Checkout\Model\Session $checkoutSession
+     * @param \Magento\Tax\Model\Config $taxConfig
+     */
+    public function __construct(
+        \Adyen\Payment\Helper\Data $adyenHelper,
+        \Magento\Store\Model\StoreManagerInterface $storeManager,
+        \Magento\Checkout\Model\Session $checkoutSession,
+        \Magento\Tax\Model\Config $taxConfig,
+        \Adyen\Payment\Logger\AdyenLogger $adyenLogger
```
cyattilakiss: can be removed here as well

etc/di.xml
AlexandrosMor: Looks like a typo

Gateway/Request/CheckoutDataBuilder.php
Aleffio: we removed the specific boletobuilder, do we really need it?

Gateway/Request/CheckoutDataBuilder.php
AlexandrosMor: Now inside boleto check we send the email

PR #392, opened 2019-02-01
[PW-978] Fix for HPP not retrieving the correct merchantAccount

Aleffio:
**Fixed issue**: Hpp request was made with the merchantAccount configured on highest level, instead of specific store/storeview

### DIFF
--- a/Helper/PaymentMethods.php
+++ b/Helper/PaymentMethods.php
@@ -191,7 +191,7 @@ class PaymentMethods extends AbstractHelper
     protected function _fetchHppMethods($store, $country)
     {
         $skinCode = $this->_adyenHelper->getAdyenHppConfigData('skin_code');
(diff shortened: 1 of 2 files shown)


Helper/PaymentMethods.php
```diff
@@ -191,7 +191,7 @@ protected function _addHppMethodsToConfig($store, $country)
     protected function _fetchHppMethods($store, $country)
     {
         $skinCode = $this->_adyenHelper->getAdyenHppConfigData('skin_code');
-        $merchantAccount = $this->_adyenHelper->getAdyenAbstractConfigData('merchant_account');
+        $merchantAccount = $this->_adyenHelper->getAdyenAbstractConfigData('merchant_account', $store);
```
cyattilakiss: get the storeId from the store object.

PR #410, opened 2019-02-21
[PW-1019] add support for banktransfer_iban

rikterbeek:
**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Block/Checkout/Success.php
+++ b/Block/Checkout/Success.php
@@ -107,7 +107,9 @@ class Success extends \Magento\Framework\View\Element\Template
     public function getBankTransferData()
     {
         $result = [];
-        if (!empty($this->getOrder()->getPayment()->getAdditionalInformation('bankTransfer.owner'))) {
+        if (!empty($this->getOrder()->getPayment()) &&
+            !empty($this->getOrder()->getPayment()->getAdditionalInformation('bankTransfer.owner'))
+        ) {
             $result = $this->getOrder()->getPayment()->getAdditionalInformation();
         }
 
--- a/Block/Checkout/Success.php
+++ b/Block/Checkout/Success.php
@@ -76,9 +76,8 @@ class Success extends \Magento\Framework\View\Element\Template
                     'boleto_pdf_url' => $this->getBoletoPdfUrl()
                 ]
             );
-            return parent::_toHtml();
         }
-        return '';
+        return parent::_toHtml();
     }
 
     /**
@@ -103,6 +102,16 @@ class Success extends \Magento\Framework\View\Element\Template
             return $this->getOrder()->getPayment()->getAdditionalInformation('url');
         }
         return null;
+    }
+
+    public function getBankTransferData()
+    {
+        $result = [];
+        if (!empty($this->getOrder()->getPayment()->getAdditionalInformation('bankTransfer.owner'))) {
+            $result = $this->getOrder()->getPayment()->getAdditionalInformation();
+        }
+
+        return $result;
     }
 
     /**
--- a/Block/Info/Hpp.php
+++ b/Block/Info/Hpp.php
@@ -41,6 +41,31 @@ class Hpp extends AbstractInfo
     }
 
     /**
+     * Get all Banktransfer related data
+     *
+     * @return array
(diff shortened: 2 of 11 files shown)


Block/Checkout/Success.php
```diff
@@ -105,6 +104,16 @@ public function getBoletoPdfUrl()
         return null;
     }
 
+    public function getBankTransferData()
+    {
+        $result = [];
+        if (!empty($this->getOrder()->getPayment()->getAdditionalInformation('bankTransfer.owner'))) {
```
cyattilakiss: Can it happen that the getPayment() returns null? In that case the statement will fail so maybe it would be better to check if that is not empty first, then the additionalInformation

view/adminhtml/templates/info/adyen_hpp.phtml
```diff
@@ -35,27 +35,80 @@
 $_info = $this->getInfo();
 $_isDemoMode = $block->isDemoMode();
 ?>
-<?php if ($_pspReference = $_info->getAdditionalInformation('pspReference')):?>
+<?php if ($_pspReference = $_info->getAdditionalInformation('pspReference')): ?>
     <div>
-        <?php if($_isDemoMode): ?>
-            <?php echo __('Adyen PSP Reference: <a href="https://ca-test.adyen.com/ca/ca/accounts/showTx.shtml?pspReference=%1&txType=Payment" target="__blank">%1</a>', $block->escapeHtml($_pspReference), $block->escapeHtml($_pspReference)) ?>
+        <?php if ($_isDemoMode): ?>
+            <?php echo __('Adyen PSP Reference: <a href="https://ca-test.adyen.com/ca/ca/accounts/showTx.shtml?pspReference=%1&txType=Payment" target="__blank">%1</a>',
```
cyattilakiss: it is `target="_blank"` not `target="__blank"`, right? (two underscores)

view/adminhtml/templates/info/adyen_hpp.phtml
```diff
@@ -35,27 +35,80 @@
...
+                $block->escapeHtml($_pspReference), $block->escapeHtml($_pspReference)) ?>
         <?php else: ?>
-            <?php echo __('Adyen PSP Reference: <a href="https://ca-live.adyen.com/ca/ca/accounts/showTx.shtml?pspReference=%1&txType=Payment" target="__blank">%1</a>', $block->escapeHtml($_pspReference), $block->escapeHtml($_pspReference)) ?>
+            <?php echo __('Adyen PSP Reference: <a href="https://ca-live.adyen.com/ca/ca/accounts/showTx.shtml?pspReference=%1&txType=Payment" target="__blank">%1</a>',
```
cyattilakiss: the same

Block/Info/Hpp.php
```diff
@@ -40,6 +40,31 @@ public function isPaymentSelectionOnAdyen()
...
+    /**
+     * Get all Banktransfer related data
+     *
+     * @return array
+     * @throws \Magento\Framework\Exception\LocalizedException
+     */
+    public function getBankTransferData()
+    {
+        $result = [];
+        if(!empty($this->getInfo()->getOrder()->getPayment()->getAdditionalInformation('bankTransfer.owner'))) {
```
cyattilakiss: the same

view/adminhtml/templates/info/adyen_hpp.phtml
rikterbeek: yes correct. This was already in the code a long time but will change it in the PR.

view/adminhtml/templates/info/adyen_hpp.phtml
rikterbeek: yes correct. This was already in the code a long time but will change it in the PR.

Block/Info/Hpp.php
rikterbeek: order can't never have no payment in this stage but I will add it for extra defensive programming

Block/Checkout/Success.php
rikterbeek: order can't never have no payment in this stage but I will add it for extra defensive programming

PR #411, opened 2019-02-26
PW-1059: add check for additionalData

Aleffio:
**Fixed issue**:  If no values are sent from Adyen in additionalData, the additionalData object won't exist and the code throws an exception for undefined index.
This scenario applies only to 3dSecure flow


Controller/Process/Validate3d.php
```diff
@@ -167,10 +167,10 @@ public function execute()
 
 
                         // TODO: add better checks if variables are available in the result
```
rikterbeek: think you did exactly what this todo is saying. Can you remove the comment.

PR #424, opened 2019-03-19
[PW-1103][PW-1104][PW-1105][PW-1106] Implement 3DS 2.0 into Magento2

cyattilakiss:
**Description**
[PW-1103] Instead of using payment-information endpoint, the first request goes to set-payment-information which doesn't place the order therefore we can decide if we need to inject other api calls and communication with the frontend or not. After the 3DS2 flow is done we do the Magento payment-information call but skip the Adyen payments call since the response is already saved only the Adyen payments/details call is missing. For this a new plugin (Plugin/PaymentInformationManagement.php) is introduced which handles the Adyen payments call and does the whole request building 
[PW-1104] Render 3DS2 components in a Magento modal. Handling the response from AdyenThreeDS2Process if 3DS2 process is done or frontend action is required
[PW-1105] New endpoint: AdyenThreeDS2Process is able to handle ajax requests from frontend with 3DS2 data in the payload. Does the payments/details calls and decides if 3DS2 is finished or another frontend action is required
[PW-1106] Feature toggle (currently not hidden for test purposes) added to the Magento backend

TODO
Additional TODOs added to the internal ticket PW-1103


view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -173,9 +186,77 @@ define(
...
+
+                var threeDS2Node = document.getElementById('threeDS2Container');
+                
+                if (type == "IdentifyShopper") {
+                    self.threeDS2Component = self.checkout
+                        .create('threeDS2DeviceFingerprint', {
+                            fingerprintToken: token,
+                            onComplete: function(result) {
+                                self.processThreeDS2(result.data);
+                                $('#modal_content').modal("closeModal");
```
cyattilakiss: rename id #modal_content to #threeDS2Modal

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -173,9 +186,77 @@ define(
...
+                                console.log(result);
+                            }
+                        });
+                }
 
-                card.mount(cardNode);
+                self.threeDS2Component.mount(threeDS2Node);
+            },
+            /**
+             *
```
cyattilakiss: Add description

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -224,9 +310,48 @@ define(
                 }
 
                 if (this.validate() && additionalValidators.validate()) {
-                    this.isPlaceOrderActionAllowed(false);
+                    //this.isPlaceOrderActionAllowed(false);
```
cyattilakiss: uncomment it

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -224,9 +310,48 @@ define(
...
+
+                    fullScreenLoader.startLoader();
+
+                    //update payment method information if additional data was changed
+                    selectPaymentMethodAction(this.getData());
+                    setPaymentMethodAction(this.messageContainer).done(
+                        function (responseJSON) {
+                            fullScreenLoader.stopLoader();
```
cyattilakiss: add
`this.isPlaceOrderActionAllowed(true);`

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -224,9 +310,48 @@ define(
...
+                            self.validateThreeDS2OrPlaceOrder(responseJSON);
+                        });
+                    return false;
+                }
 
-                    this.getPlaceOrderDeferredObject()
+                return false;
+            },
+            /**
+             *
```
cyattilakiss: add description

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -224,9 +310,48 @@ define(
...
+             * @param responseJSON
+             */
+            validateThreeDS2OrPlaceOrder: function(responseJSON) {
+                var self = this;
+
+                console.log(responseJSON);
```
cyattilakiss: remove console log

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -193,8 +274,13 @@ define(
                         'expiryMonth': this.expiryMonth(),
                         'expiryYear': this.expiryYear(),
                         'holderName': this.creditCardOwner(),
-                        'store_cc': this.setStoreCc(),
-                        'number_of_installments': this.installment()
+                        'store_cc': this.setStoreCc,
+                        'number_of_installments': this.installment(),
+                        'java_enabled': navigator.javaEnabled().toString(),
```
rkewlani: please use already available 3DS2 helper functions. It will be easy and consistent across all plugins.
https://github.com/Adyen/adyen-3ds2-js-utils

Observer/AdyenCcDataAssignObserver.php
```diff
@@ -53,7 +58,12 @@ class AdyenCcDataAssignObserver extends AbstractDataAssignObserver
         self::ENCRYPTED_EXPIRY_MONTH,
         self::ENCRYPTED_EXPIRY_YEAR,
         self::HOLDER_NAME,
-        self::VARIANT
+        self::VARIANT,
+        self::JAVA_ENABLED,
```
rkewlani: We can use an object to add all browser related information similar to given in standard helper functions.
https://github.com/Adyen/adyen-3ds2-js-utils (refer BrowserInfo)

Helper/Data.php
```diff
@@ -35,8 +35,8 @@ class Data extends AbstractHelper
     const LIVE = 'live';
     const CHECKOUT_CONTEXT_URL_LIVE = 'https://checkoutshopper-live.adyen.com/checkoutshopper/';
     const CHECKOUT_CONTEXT_URL_TEST = 'https://checkoutshopper-test.adyen.com/checkoutshopper/';
-    const CHECKOUT_COMPONENT_JS_LIVE = 'https://checkoutshopper-live.adyen.com/checkoutshopper/sdk/2.1.0/adyen.js';
-    const CHECKOUT_COMPONENT_JS_TEST = 'https://checkoutshopper-test.adyen.com/checkoutshopper/sdk/2.1.0/adyen.js';
+    const CHECKOUT_COMPONENT_JS_LIVE = 'https://checkoutshopper-live.adyen.com/checkoutshopper/sdk/2.2.0/adyen.js';
```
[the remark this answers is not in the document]
Aleffio: Checked with Marc, there are no necessary changes apart from the checkbox for recurring, he suggests though to upgrade to 2.2.1 because of a small issue with BCMC

Plugin/PaymentInformationManagement.php
```diff
@@ -0,0 +1,168 @@
...
+        $request = $this->adyenRequestHelper->buildCustomerIpData($request, $shopperIp);
+
+        // AddressDataBuilder
+        $shippingAddress = $quote->getShippingAddress();
+        $request = $this->adyenRequestHelper->buildAddressData($request, $billingAddress, $shippingAddress);
+
+        // PaymentDataBuilder
+        $currencyCode = $quote->getQuoteCurrencyCode();
+        $amount = $quote->getGrandTotal();
+        $reference = $quote->getId();
```
Aleffio: $quote->getId() returns an int, so instead of having 000000007 as merchantReference we have 7, not sure if there is a way to get the orderNumber at this point of the flow

Observer/AdyenCcDataAssignObserver.php
```diff
@@ -40,6 +40,11 @@ class AdyenCcDataAssignObserver extends AbstractDataAssignObserver
     const ENCRYPTED_EXPIRY_YEAR = 'expiryYear';
     const HOLDER_NAME = 'holderName';
     const VARIANT = 'variant';
+    const JAVA_ENABLED = 'java_enabled';
+    const SCREEN_COLOR_DEPTH = 'screen_color_depth';
+    const SCREEN_WIDTH = 'screen_width';
+    const SCREEN_HEIGHT = 'screen_height';
+    const TIMEZONE_OFFSET = 'timezone_offset';
```
rkewlani: I am seeing some fields like channel and origin not present. Not sure if you will add it in next story. Just thought to mention so that we don't miss it.

Helper/Data.php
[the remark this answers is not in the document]
cyattilakiss: do we need to adjust our code to the change related to the BCMC? Or it is just a patch on their side for which we haven't implemented any workaround before?

Observer/AdyenCcDataAssignObserver.php
cyattilakiss: This one is for validating the parameters passed by the frontend in the additional_data array. In this case we follow the Magento 2 approach while creating these constants.

Observer/AdyenCcDataAssignObserver.php
cyattilakiss: Good one. Actually I retrieve the origin from the frontend and not from the frontend. Let's discuss if we should grab everything from the frontend or if it is more secure to get from backend. I have already had the functions to retrieve these details from the backend so it was also easier for me.

Plugin/PaymentInformationManagement.php
cyattilakiss: yeah true... the problem is that we don't have an order at this point of the flow
maybe there is a reserved order number or something for the quotes - let's check it because the notification won't work in the current case

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
cyattilakiss: Thank you for the heads up, I didn't know that there is this helper but definitely it would have been easier to use it :) I will just change the current functions to this.

Helper/Requests.php
```diff
@@ -0,0 +1,419 @@
...
+     * @param $billingAddress
+     * @return mixed
+     */
+    public function buildCustomerData($request = [], $customerId = 0, $billingAddress, \Magento\Quote\Model\Quote\Payment $payment)
+    {
+        if ($customerId > 0) {
+            $request['shopperReference'] = $customerId;
+        }
+
+        if (!empty($billingAddress)) {
```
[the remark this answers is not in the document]
cyattilakiss: remove it, prefill the fields if the field is required by adyen

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -193,13 +283,28 @@ define(
...
                         'holderName': this.creditCardOwner(),
-                        'store_cc': this.setStoreCc(),
-                        'number_of_installments': this.installment()
+                        'store_cc': this.storeCc,
+                        'number_of_installments': this.installment(),
+                        'java_enabled': browserInfo.javaEnabled,
+                        'screen_color_depth': browserInfo.colorDepth,
+                        'screen_width': browserInfo.screenWidth,
+                        'screen_height': browserInfo.screenHeight,
+                        'timezone_offset': browserInfo.timeZoneOffset
```
cyattilakiss: send the whole browserinfo from frontend to backend

Helper/Requests.php
```diff
@@ -0,0 +1,446 @@
...
+        $request['browserInfo']['screenWidth'] = $payment->getAdditionalInformation(AdyenCcDataAssignObserver::SCREEN_WIDTH);
+        $request['browserInfo']['screenHeight'] = $payment->getAdditionalInformation(AdyenCcDataAssignObserver::SCREEN_HEIGHT);
+        $request['browserInfo']['colorDepth'] = $payment->getAdditionalInformation(AdyenCcDataAssignObserver::SCREEN_COLOR_DEPTH);
+        $request['browserInfo']['timeZoneOffset'] = $payment->getAdditionalInformation(AdyenCcDataAssignObserver::TIMEZONE_OFFSET);
+        $request['browserInfo']['language'] = $this->adyenHelper->getCurrentLocaleCode($store);
+
+        if ($javaEnabled = $payment->getAdditionalInformation(AdyenCcDataAssignObserver::JAVA_ENABLED)) {
+            $request['browserInfo']['javaEnabled'] = $javaEnabled;
+        } else {
+            $request['browserInfo']['javaEnabled'] = false;
```
cyattilakiss: check if we send as an object from frontend if we still need this check

Helper/Data.php
```diff
@@ -1291,11 +1345,15 @@ public function getPoiId($storeId = null)
      * return the merchant account name defined in required settings.
      *
      * @param $paymentMethod
-     * @param int $storeId
+     * @param int|null $storeId
      * @return string
      */
-    public function getAdyenMerchantAccount($paymentMethod, $storeId)
+    public function getAdyenMerchantAccount($paymentMethod, $storeId = null)
```
maassenbas: Why $paymentMethod as parameter?

Plugin/PaymentInformationManagement.php
```diff
@@ -0,0 +1,207 @@
...
+ * Author: Adyen <magento@adyen.com>
+ */
+
+namespace Adyen\Payment\Plugin;
+
+use Adyen\Payment\Observer\AdyenHppDataAssignObserver;
+use Adyen\Payment\Observer\AdyenCcDataAssignObserver;
+use Magento\Vault\Model\Ui\VaultConfigProvider;
+
+class PaymentInformationManagement
```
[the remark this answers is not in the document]
cyattilakiss: The main idea is to always use the original/existing endpoints instead of exposing new ones and the set-payment-information endpoint looks just perfect for this usage (which saves the payment information but doesn't place the order) and in the flow we need to add the new 3ds2 functionality.

The other approach could be exposing a new endpoint and using that but the Magento community suggested to use the existing default ones

Gateway/Request/CcAuthorizationDataBuilder.php
```diff
@@ -24,91 +24,34 @@
...
+            // the payments response needs to be passed to the next process because after this point we don't have
+            // access to the payment object therefore to the additionalInformation array
+            $request = $response;
+            // Remove from additional data
+            $payment->unsAdditionalInformation("paymentsResponse");
 
-        if ($cardNumber = $payment->getAdditionalInformation(AdyenCcDataAssignObserver::ENCRYPTED_CREDIT_CARD_NUMBER)) {
-            $request['paymentMethod']['encryptedCardNumber'] = $cardNumber;
-        }
+            // TODO check if qoupte needs to be saved or not
```
maassenbas: Check the "TODO" in comments

Gateway/Validator/ThreeDS2ResponseValidator.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $isValid = true;
+        $errorMessages = [];
+
+        // validate result
+        if (!empty($response['resultCode'])) {
+            // 3DS2.0 should have IdentifyShopper or ChallengeShopper as a resultCode
+            switch ($response['resultCode']) {
+                case "IdentifyShopper":
+                    if (!empty($response['authentication']['threeds2.fingerprintToken'])) {
```
KadoBOT: Maybe this can be refactored a bit? The `if` inside a `switch` that is inside an `if` raises the complexity and makes it harder to reason about.

Helper/Requests.php
```diff
@@ -0,0 +1,446 @@
...
+     */
+    public function buildCustomerData($request = [], $customerId = 0, $billingAddress, \Magento\Quote\Model\Quote\Payment $payment)
+    {
+        if ($customerId > 0) {
+            $request['shopperReference'] = $customerId;
+        }
+
+        if (!empty($billingAddress)) {
+            // Openinvoice and afterpayTouch methods requires different request format
+            if ($this->adyenHelper->isPaymentMethodOpenInvoiceMethod(
```
KadoBOT: Can this condition be extracted to a variable? There is so much going on that It's a bit hard to understand what the condition really is.

view/frontend/web/js/threeds2-js-utils.js
```diff
@@ -0,0 +1 @@
...ndChild(i),a},getBrowserInfo:function(){var e=window&&window.screen?window.screen.width:"",n=window&&window.screen?window.screen.height:"",t=window&&window.screen?window.screen.colorDepth:"",r=window&&window.navigator?window.navigator.userAgent:"",o=!(!window||!window.navigator)&&navigator.javaEnabled(),a="";return window&&window.navigator&&(a=window.navigator.language?window.navigator.language:window.navigator.browserLanguage),{screenWidth:e,screenHeight:n,colorDepth:t,userAgent:r,timeZoneOffset:(new Date).getTimezoneOffset(),language:a,javaEnabled:o}},base64Url:{encode:function(e){var n=window.btoa(e).split("=")[0];return n=(n=n.replace("/+/g","-")).replace("///g","_")},decode:function(e){var n=e;switch((n=(n=n.replace("/-/g","+")).replace("/_/g","/")).length%4){case 0:break;case 2:n+="==";break;case 3:n+="=";break;default:window.console&&window.console.log&&window.console.log("### base64url::decodeBase64URL::  Illegal base64url string!")}try{return window.atob(n)}catch(e){throw new Error(e)}}},config:{challengeWindowSizes:o,validateChallengeWindowSize:a,getChallengeWindowSize:function(e){return o[a(e)]},THREEDS_METHOD_TIMEOUT:1e4,CHALLENGE_TIMEOUT:6e5}};n.default=i}]).default});
```
KadoBOT: Is this generated file or what? It seems like. Where this comes from?

cyattilakiss: yes it is: https://github.com/Adyen/adyen-3ds2-js-utils

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -95,20 +106,23 @@ define(
                 var allInstallments = self.getAllInstallments();
                 var cardNode = document.getElementById('cardContainer');
 
-                var checkout = new AdyenCheckout({
-                    locale: self.getLocale()
-                });
+                var enableStoreDetails = false;
+                if (self.canCreateBillingAgreement() && !self.isVaultEnabled())  {
+                    enableStoreDetails = true;
+                }
```
KadoBOT: I rather make this a function to make sure that `enableStoreDetails` is mutated. Also, `var` is hoisted which may cause bugs in the future.
I suggest to make something like:
```
var getStoreDetails = function() {
  return self.canCreateBillingAgreement() && !self.isVaultEnabled()
}
```
and in the object:
```
...
enableStoreDetails: getStoreDetails()
...
```

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -193,13 +283,28 @@ define(
...
                 };
                 this.vaultEnabler.visitAdditionalData(data);
                 return data;
             },
+            /**
+             * Get data for place order
+             * @returns {{method: *}}
+             */
+            getData: function() {
+                var data = {
```
KadoBOT: no need to assign to a variable, just return the object

view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
```diff
@@ -165,46 +170,50 @@ define(
...
+                         *
                          * @override
+                         *
+                         * @param data
+                         * @param event
+                         * @returns {boolean}
                          */
                         placeOrder: function (data, event) {
                             var self = this;
+                            console.log("placeorder");
```
KadoBOT: 🙈

view/frontend/web/js/threeds2-js-utils.js
KadoBOT: Can't we just use the library instead of copying the content?

PR #425, opened 2019-03-22
PW-1114: Update documentation links

Aleffio:


adyen-gc:
Checked each one, looks good. Thanks Aleffio

PR #428, opened 2019-04-10
PW-1200: Always pass storeid to initializeAdyenClient when available 

Aleffio:


Aleffio:
@cyattilakiss makes sense, we already have $storeManager in the helper

PR #430, opened 2019-04-16
Release 4.0.2

Aleffio:


Helper/Data.php
```diff
@@ -1343,6 +1343,10 @@ public function formatTerminalAPIReceipt($paymentReceipt)
     public function initializeAdyenClient($storeId = null, $apiKey = null)
     {
         // initialize client
+        if (empty($storeId)) {
```
cyattilakiss: can store id be 0 or starts with 1? Maybe better to change it to 
```suggestion
        if ($storeId === null) {
```

PR #434, opened 2019-04-25
PW-1085 House number parsing doesn't always work properly

cyattilakiss:
**Description**
Send house number separately only at klarna enabled countries

Tested case
Klarna: NL, DE, NO, SE


Block/Redirect/Redirect.php
```diff
@@ -382,20 +389,27 @@ protected function setBillingAddressData($formFields)
     protected function setShippingAddressData($formFields)
     {
         $shippingAddress = $this->_order->getShippingAddress();
-
+        $this->_adyenLogger->addAdyenDebug($shippingAddress->getCountryId());
```
Aleffio: can be removed?

PR #436, opened 2019-05-02
PW-1173: Ignore OFFER_CLOSED for already authorised orders, or for or…

Aleffio:
…ders that are not matching the paymentmethod of the notification
Set the ccType of boleto payments to the proper boleto variant.
reset order id on card placeorder

**Tested scenarios**
1) Credit card -> 3ds2 -> refresh page on challengeShopper ->  pay with boleto:
receive offerclosed for creditcard -> no action
receive offerclosed for boleto -> cancel order

2) Credit card -> 3ds2 -> refresh page on challengeShopper -> Credit card -> new order id so no action required

**Fixed issue**:

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -898,6 +898,41 @@ class Cron
                 }
                 break;
             case Notification::OFFER_CLOSED:
+                $previousSuccess = $this->_order->getData('adyen_notification_event_code_success');
+
+                // Order is already Authorised
+                if ($previousSuccess) {
+                    $this->_adyenLogger->addAdyenNotificationCronjob("Order is already authorised, skipping OFFER_CLOSED");
+                    break;
+                }
+
+                /*
+                 * For cards, it can be 'visa', 'maestro',...
+                 * For alternatives, it can be 'ideal', 'directEbanking',...
+                 */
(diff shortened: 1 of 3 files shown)


Model/Cron.php
```diff
@@ -898,6 +898,41 @@ protected function _processNotification()
                 }
                 break;
             case Notification::OFFER_CLOSED:
+                $previousSuccess = $this->_order->getData('adyen_notification_event_code_success');
+
+                // Order is already Authorised
+                if ($previousSuccess) {
```
rikterbeek: use !empty()

Model/Cron.php
```diff
@@ -1139,9 +1174,14 @@ protected function _setRefundAuthorized()
     protected function _authorizePayment()
     {
         $this->_adyenLogger->addAdyenNotificationCronjob('Authorisation of the order');
+
+        // Set adyen_notification_event_code_success to true so that we ignore a possible OFFER_CLOSED
+        if (strcmp($this->_success, 'true') == 0) {
+            $this->_order->setData('adyen_notification_event_code_success', 1);
```
rikterbeek: why setData instead of addiontionData ?

PR #450, opened 2019-05-28
[PW-1330] only log onError as onComplete will always be executed

rikterbeek:
**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
@@ -303,7 +303,8 @@ define(
                                     'screen_color_depth': browserInfo.colorDepth,
                                     'screen_width': browserInfo.screenWidth,
                                     'screen_height': browserInfo.screenHeight,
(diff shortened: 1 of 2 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
```diff
@@ -304,8 +303,7 @@ define(
                                     'screen_color_depth': browserInfo.colorDepth,
                                     'screen_width': browserInfo.screenWidth,
                                     'screen_height': browserInfo.screenHeight,
-                                    'timezone_offset': browserInfo.timeZoneOffset,
-                                    'language': browserInfo.language
```
cyattilakiss: we use this line of code with the browserinfo

rikterbeek: hmm seems to be a rebase issue will revert thanks.

rikterbeek: done

PR #453, opened 2019-06-07
PW-1273 Undefined index check in cc payment response PW-1349 getCurrentLocaleCode fallback bug

cyattilakiss:
**Description**
Fix for unhandled result in the query parameters
Fix for billing agreement undefined index in case of api response
misconfiguration
Fix undefined $store variable

### DIFF
--- a/Controller/Process/Result.php
+++ b/Controller/Process/Result.php
@@ -214,13 +214,18 @@ class Result extends \Magento\Framework\App\Action\Action
     {
         $result = false;
 
-        $this->_adyenLogger->addAdyenResult('Updating the order');
-
         if (!empty($response['authResult'])) {
 			$authResult = $response['authResult'];
 		} elseif (!empty($response['resultCode'])) {
 			$authResult = $response['resultCode'];
-		}
+        } else {
+            // In case the result is unknown we log the request and don't update the history
+            $this->_adyenLogger->addError("Unexpected result query parameter. Response: " . json_encode($response));
+
+            return $result;
+        }
+
+        $this->_adyenLogger->addAdyenResult('Updating the order');
 
         $paymentMethod = isset($response['paymentMethod']) ? trim($response['paymentMethod']) : '';
         $pspReference = isset($response['pspReference']) ? trim($response['pspReference']) : '';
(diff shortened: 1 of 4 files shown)


Model/Billing/Agreement.php
```diff
@@ -179,36 +179,41 @@ public function setCcBillingAgreement($contractDetail, $storeOneClick, $storeId)
             ->setMethodCode('adyen_oneclick')
             ->setReferenceId($contractDetail['recurring.recurringDetailReference']);
 
+        if (!isset($contractDetail['cardBin']) ||
+            !isset($contractDetail['cardHolderName']) ||
+            !isset($contractDetail['cardSummary']) ||
+            !isset($contractDetail['expiryDate']) ||
+            !isset($contractDetail['paymentMethod'])
+        ) {
+            $this->_errors[] = __('"In the Additional data in API response section, select: Card summary, Expiry Date, Cardholder name, Recurring details and Variant to create billing agreements immediately after the payment is authorized."');
```
rikterbeek: where is this being used ?

Controller/Process/Result.php
```diff
@@ -214,13 +214,18 @@ protected function _validateUpdateOrder($order, $response)
     {
         $result = false;
 
-        $this->_adyenLogger->addAdyenResult('Updating the order');
-
```
rikterbeek: don't we want to keep this log so we now the order updating process is started ?

cyattilakiss: I put it after the check for the result query param so it won't make any confusion that we are updating the order while we have already jumped out of it.

https://github.com/Adyen/adyen-magento2/pull/453/files#diff-bd7cef8f0429f77c78243ef5e37f68a5R228

Model/Billing/Agreement.php
cyattilakiss: We show it to the user here:
https://github.com/Adyen/adyen-magento2/pull/453/files#diff-735f70c1f948ebbd90915b1548d93432R1631

This is shown at the magento backend under the order

rikterbeek: thanks clear

PR #455, opened 2019-06-11
PW-1320 Force on latest patch version

cyattilakiss:
**Description**
force framework usage of the latest patches

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Inside Adyen toggle the following settings on inside the API and Responses secti
 * Variant
 
 ## Requirements
-This plugin supports Magento2 version 2.2 and higher.
+This plugin supports Magento2 version 2.2.8 and higher.
 
 ## Collaboration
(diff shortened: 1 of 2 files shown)


ebaschiera:
Hi guys,
I am not sure this is the best approach to ensure merchants have the safest environment.
There are new bugs at every Magento minor release, given the fact that Magento doesn't release security patches alone. In fact we tend to apply patches from https://github.com/magento/ece-tools/tree/develop/patches to close vulnerabilities and keep the bugs away.
This PR blocks our workflow, as we keep the original Magento version.
As side note, there's a bit of discussion about security patches going on here: https://twitter.com/mrloo/status/1139317642488160256

cyattilakiss:
Hi @ebaschiera ,

Thank you for the heads up, we are going to investigate it in the current sprint and will get back to you.

Best,
Attila
Adyen

PR #458, opened 2019-06-12
PW-1359 Bankaccount is missing on success page and backend

cyattilakiss:
**Description**
Add missing IBAN to the frontend and admin

### DIFF
--- a/view/adminhtml/templates/info/adyen_hpp.phtml
+++ b/view/adminhtml/templates/info/adyen_hpp.phtml
@@ -91,6 +91,13 @@ $_isDemoMode = $block->isDemoMode();
                 </tr>
             <?php endif; ?>
 
+            <?php if (!empty($banktranferData['bankTransfer.iban'])) : ?>
+                <tr>
+                    <th><?php echo $block->escapeHtml(__('IBAN')); ?></th>
+                    <td><?php echo $banktranferData['bankTransfer.iban'] ?></td>
+                </tr>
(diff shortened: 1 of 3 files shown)


view/adminhtml/templates/info/adyen_hpp.phtml
```diff
@@ -91,6 +91,13 @@ $_isDemoMode = $block->isDemoMode();
                 </tr>
             <?php endif; ?>
 
+            <?php if (!empty($banktranferData['bankTransfer.iban'])) : ?>
+                <tr>
+                    <th><?php echo $block->escapeHtml(__('IBAN')); ?></th>
+                    <td><?php echo $banktranferData['bankTransfer.iban'] ?></td>
+                </tr>
+            <?php endif; ?>
+
```
KadoBOT: Is it possible in PHP to import this chunk of code? I see that it is repeated three times, so my suggestion, if possible, is writing it once, and reusing it within those 3 files.

cyattilakiss: yeah these are magento specific phtml files, I think it would just add more complexity if we would start importing sub templates :(

KadoBOT: Gotcha!

rikterbeek:
@cyattilakiss let's make a hotfix release for this or a bigger one if we have more oustanding fixes/features.

PR #471, opened 2019-07-01
Update Requests.php

peterjaap:
**Description**

**Tested scenarios**

**Fixed issue**:


rikterbeek:
Hi Peter Jaap,

Thanks for the PR. This will impact all merchants integrations. I am not a fan of settings but I think we need to have this as an setting if you want to use street1 and street2 combined. As some merchants using street2 for the house number and this is send in the housenumber field. You don't want to send this twice that could lead to issues with openinvoice providers.

Regards,
Rik
Adyen

rikterbeek:
Hi Peter Jaap,

Thanks for the PR. We will look into this.

Regards,
Rik
Adyen

Issue #461, opened 2019-06-13
Refresh payment methods on country change


sanderjongsma:
**Magento version**: 2.3.1
**Plugin version**: 4.1.0
**Description**
The payment methods are rendered if you move from the shipping to the payment step. But if you move back to the shipping step from the payment step and change the country, the shipping methods are not refreshed. 

I think it is a bug because a lot of payment options are based on the country. Do you agree? Or is it supposed to work like this.

cyattilakiss:
Hi @sanderjongsma ,

Thank you for raising the issue.
We are going to test asap it and making the fix after being able to reproduce this.

Best,
Attila
Adyen

cyattilakiss:
Hi @sanderjongsma we are going to release the changes this week.

Best,
Attila

PR #478, opened 2019-07-15
Release 4.3.0

cyattilakiss:
**Description**

PW-1269 Possibility for multiple terminals
PW-1368 Upgrade components version to 3.0.0
PW-1385 Refresh payment methods on country change 
PW-1396 Add separate port support for origin key retrieval
PW-1405 Call to a member function getAdditionalInformation() on null
PW-1417 Check issuers in the whole response detail object
PW-1512 Update currency codes
PW-1518 Updated installment options up to 24x

**Fixed issue**:
#461
#471


Helper/PaymentMethods.php
```diff
@@ -390,4 +390,41 @@ protected function getCurrentShopperReference()
...
+
+        // In case the POS store id is set, provide in the request
+        if (!empty($this->adyenHelper->getPosStoreId($storeId))) {
+            $requestParams['store'] = $this->adyenHelper->getPosStoreId($storeId);
+        }
+
+        try {
+            $responseData = $service->getConnectedTerminals($requestParams);
+        } catch (\Adyen\AdyenException $e) {
+            $this->adyenLogger->error(
```
cyattilakiss: we should log the actual error message too

composer.json
```diff
@@ -14,7 +14,7 @@
     }
   ],
   "require": {
-    "adyen/php-api-library": ">=2.0.0",
+    "adyen/php-api-library": ">=2.1.0",
```
cyattilakiss:
```suggestion
    "adyen/php-api-library": "2.*",
```

cyattilakiss:
```suggestion
    "adyen/php-api-library": "~2.1",
```

rikterbeek:
@Aleffio we have 3 other PR open can we merge this and include it into the release ?

cyattilakiss:
> @Aleffio we have 3 other PR open can we merge this and include it into the release ?

merged them

cyattilakiss:
Let's merge first #486 into this release, then I think do the release because this is now pending for a long time. After the release we can include #483 into the next patch release after finishing testing.

othuress:
RE: PW-1385 Refresh payment methods on country change
After this update some payment methods doesn't work properly.
Adyen payment resets the data (selected bank in the bank dropdown and "Personal Number") every time you change something in the address or the address is validated:

1. Select some bank in the "Bankbetalning" payment or fill "Personal Number" in "Klarna Faktura" method.

2. Fill or change the street address line.

3. The payment form will reset itself in a second.

Steps to reproduce:

1. Select some bank in the "Bankbetalning" payment or fill "Personal Number" in "Klarna Faktura" method.

2. Run the following code in the browser console:

require(['uiRegistry'], function(registry) {

    registry

        .get('checkout.steps.shipping-step.shippingAddress')

        .validateShippingInformation()

});


---------------------------

checkout calls address validation when you press "Next" on the first page of the checkout and after each billing address update. (I need to re-enter all payment details again if I fill them first and then change billing address details.)

can this be addressed? its enough the payments are updated when country is changed

cyattilakiss:
Hi @othuress ,

Thank you for your comment. This looks like a valid issue so I would recommend you to open an issue under this repository. I just double checked it and unfortunately not the merge of PW-1385 ticket caused the issue but it was already there before, so we still have to figure out how to fix this issue.

Best,
Attila
Adyen

PR #483, opened 2019-07-19
Load quote from order for open invoice payments

JeroenVanLeusden:
**Description**
Since the upgrade To Magento v2.3.2 we were unable to use klarna because the line items were empty. Loading the quote from the checkout session resulted in a brand new quote without items.

This PR prevents that by loading the quote using quote id that is saved in the order.

**Tested scenarios**
1. Place order using klarna on Magento v2.3.2


cyattilakiss:
Hi @JeroenVanLeusden ,

Thank you for your PR (and also for the previous one :) )
Do you know if this fix breaks the support for previous versions? (2.2.8+ and 2.3.1+)
We will also give it a look at these versions and verify the PR.

Best,
Attila

JeroenVanLeusden:
Haven't tested any other Magento versions but don't expect any issues. Please let me know if you find anything!

PR #486, opened 2019-07-26
Fix issues with popup in live

rikterbeek:
**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
@@ -246,23 +246,39 @@ define(
                     self.threeDS2ChallengeComponent = self.checkout
                         .create('threeDS2Challenge', {
                             challengeToken: token,
+                            size: '05',
                             onComplete: function (result) {
-                                popupModal.modal("closeModal");
+                                self.closeModal(popupModal);
+
                                 fullScreenLoader.startLoader();
                                 threeds2.processThreeDS2(result.data).done(function (responseJSON) {
                                     self.validateThreeDS2OrPlaceOrder(responseJSON);
                                 }).error(function () {
-                                    popupModal.modal("closeModal");
                                     self.isPlaceOrderActionAllowed(true);
                                     fullScreenLoader.stopLoader();
                                 });
                             },
                             onError: function (error) {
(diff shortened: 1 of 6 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -246,24 +246,40 @@ define(
...
                     self.threeDS2ChallengeComponent.mount(threeDS2Node);
                 }
             },
+            /**
+             * This method is a workaround to close the modal in the right way and reconstruct the threeDS2Modal.
+             * This will solve issues when you cancel the 3DS2 challenge and retry the payment
+             */
+            closeModal: function (popupModal) {
+                popupModal.modal("closeModal");
+                $('.threeDS2Modal').remove();
```
cyattilakiss: Isn't it an id?
```suggestion
                $('#threeDS2Modal').remove();
```

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -246,24 +246,40 @@ define(
...
                             onError: function (error) {
+                                self.closeModal(popupModal);
                                 console.log(JSON.stringify(error));
                             }
                         });
                     self.threeDS2ChallengeComponent.mount(threeDS2Node);
                 }
             },
+            /**
+             * This method is a workaround to close the modal in the right way and reconstruct the threeDS2Modal.
```
cyattilakiss: There supposed to be a function for the checkout component which destroys the component itself and makes it available to mount other components to the div in the DOM. I like that approach more than removing then adding the elements with JS to the DOM but I would say let's merge it and then I will give it a look later.

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
rikterbeek: no it creates an class for this by the modal

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
rikterbeek: This is not related to this. But the only workaround we found to destroy the modal completly.

PR #487, opened 2019-07-28
PW-1527 Fix cc_type saving for vault

cyattilakiss:
**Description**
Save ccType in every case when it is available - fix for vault payments

**Tested scenarios**
Vault payment

### DIFF
--- a/Gateway/Validator/CheckoutResponseValidator.php
+++ b/Gateway/Validator/CheckoutResponseValidator.php
@@ -33,6 +33,11 @@ class CheckoutResponseValidator extends AbstractValidator
     private $adyenLogger;
 
     /**
+     * @var \Adyen\Payment\Helper\Data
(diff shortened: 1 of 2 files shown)


Gateway/Validator/CheckoutResponseValidator.php
```diff
@@ -80,6 +87,13 @@ public function validate(array $validationSubject)
                         }
                     }
 
+                    // Save cc_type if available in the response
+                    if (!empty($response['additionalData']['paymentMethod'])) {
+                        $ccType = $this->adyenHelper->getMagentoCreditCartType($response['additionalData']['paymentMethod']);
+                        $payment->setAdditionalInformation('cc_type', $ccType);
```
rikterbeek: Do we log when ccType is null ?

PR #495, opened 2019-08-08
PW-1510 Add instalments for Terminal API

cyattilakiss:
**Description**
Add configuration for POS instalments
Add POS instalments
Refactor instalments on the frontend (credit cards, pos cloud)
If no terminal is available show message

### DIFF
--- a/etc/adminhtml/system/adyen_pos_cloud.xml
+++ b/etc/adminhtml/system/adyen_pos_cloud.xml
@@ -78,6 +78,28 @@
             <config_path>payment/adyen_pos_cloud/recurring_type</config_path>
             <comment>If you want to store the credentials for ONECLICK or RECURRING or combination of these two you need to contact support@adyen.com and ask if they can enable to store cards on company level. Changing this setting without doing this will lead to failed payments!</comment>
         </field>
+        <field id="enable_installments" translate="label" type="select" sortOrder="219" showInDefault="1"
+               showInWebsite="1" showInStore="1">
+            <label>Enable Installments</label>
+            <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+            <tooltip>Enable installments for each credit card type.</tooltip>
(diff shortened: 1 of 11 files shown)


etc/adminhtml/system/adyen_pos_cloud.xml
```diff
@@ -78,6 +78,28 @@
             <config_path>payment/adyen_pos_cloud/recurring_type</config_path>
             <comment>If you want to store the credentials for ONECLICK or RECURRING or combination of these two you need to contact support@adyen.com and ask if they can enable to store cards on company level. Changing this setting without doing this will lead to failed payments!</comment>
         </field>
+        <field id="enable_installments" translate="label" type="select" sortOrder="219" showInDefault="1"
```
rikterbeek: Do we need this field, can't we just apply instalments if there are configured?

view/frontend/web/template/payment/pos-cloud-form.html
```diff
@@ -50,38 +50,69 @@
...
+
         <div class="actions-toolbar">
             <div class="primary">
                 <button class="action primary checkout"
                         type="submit"
                         data-bind="
                             click: initiate,
                             attr: {title: $t('Place Order')},
-                            enable: getConnectedTerminals().length > 0
+                            enable: isButtonActive()
```
rikterbeek: should we not check if there is any connected terminals to select before making the button active ?
What will happen if there are no connected terminals available ?

etc/adminhtml/system/adyen_pos_cloud.xml
cyattilakiss: We could but this is a nice feature just to enable and disable them without losing the instalments configuration

view/frontend/web/template/payment/pos-cloud-form.html
cyattilakiss: indeed, I will put this check in the isButtonActive function

PR #496, opened 2019-08-08
PW-390 Pay with Google for Web

Aleffio:
**Description**
Add Pay with Google for Web payment method inside M2

**Tested scenarios**
- Authorisation
- Saved billing agreement through notification
- Saved billing agreement through synchronous response
- Saved vault token through synchronous response


Block/Form/GooglePay.php
```diff
@@ -0,0 +1,62 @@
...
+//     */
+//    public function getApplePayShippingTypes()
+//    {
+//        $applePayShippingTypes = $this->_adyenHelper->getApplePayShippingTypes();
+//        $types = [];
+//        foreach ($applePayShippingTypes as $applePayShippingType) {
+//            $types[$applePayShippingType['value']] = $applePayShippingType['label'];
+//        }
+//        return $types;
+//    }
```
cyattilakiss: Remove commented part

Helper/Data.php
```diff
@@ -288,9 +288,14 @@ public function formatAmount($amount, $currency)
                 $format = 2;
         }
 
+        if ($returnFormat) {
```
cyattilakiss: I would outsource it to another function and would call that function in this function for formatting

Model/Billing/Agreement.php
```diff
@@ -190,7 +193,10 @@ public function setCcBillingAgreement($contractDetail, $storeOneClick, $storeId)
         }
         // Billing agreement is CC
 
-        $ccType = $contractDetail['paymentMethod'];
+        $ccType = $variant = $contractDetail['paymentMethod'];
```
cyattilakiss: why do we need to vars for the same value?

Aleffio: ccType will be replaced in line 203 and will be the Magento card type, we still need the adyen type to save in 252

Helper/Data.php
Aleffio: good idea, maybe better to do it in the PHP library and call it here

Block/Form/GooglePay.php
Aleffio: oops

Helper/Data.php
cyattilakiss: yeah since we do the same in every plugin, okay then in the library :)

Gateway/Request/GooglePayAuthorizationDataBuilder.php
```diff
@@ -0,0 +1,78 @@
...
+        $paymentDataObject = \Magento\Payment\Gateway\Helper\SubjectReader::readPayment($buildSubject);
+        $payment = $paymentDataObject->getPayment();
+        $token = $payment->getAdditionalInformation('token');
+
+        $request['paymentMethod']['type'] = 'paywithgoogle';
+        // get payment data
+        if ($token) {
+            $parsedToken = json_decode($token);
+            try {
+				$request['paymentMethod']['paywithgoogle.token'] = $parsedToken;
```
rikterbeek: looks like some weird formatting or is it just only on github like this ?

Helper/Data.php
rikterbeek: not the responsibility of this method indeed. you can just split the responsibility. Are you adding this in the library?
Woul be great to add a task for it and update this to the library when we support al v49 PMs integration.

etc/adminhtml/system/adyen_google_pay.xml
```diff
@@ -0,0 +1,55 @@
...
+  ~ Adyen Payment Module
+  ~
+  ~ Copyright (c) 2019 Adyen B.V.
+  ~ This file is open source and available under the MIT license.
+  ~ See the LICENSE file for more info.
+  ~
+  ~ Author: Adyen <magento@adyen.com>
+  -->
+<include xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Config:etc/system_include.xsd">
+    <group id="adyen_google_pay" translate="label" type="text" sortOrder="460" showInDefault="1" showInWebsite="1" showInStore="1">
```
rikterbeek: Did you asked grant for help on the fields ?

rikterbeek:
@Aleffio is the token by default visible=false for google pay transaction so it is not visible  for the shopper ?

Gateway/Request/GooglePayAuthorizationDataBuilder.php
Aleffio: just github apparently

Aleffio:
> @Aleffio is the token by default visible=false for google pay transaction so it is not visible for the shopper ?

added

etc/adminhtml/system/adyen_google_pay.xml
Aleffio: meeting him in the afternoon

Gateway/Request/GooglePayAuthorizationDataBuilder.php
msilvagarcia: This line in indented with tabs instead of spaces, and GitHub expands those to 8 characters

view/frontend/layout/vault_cards_listaction.xml
```diff
@@ -28,7 +28,7 @@
         <referenceContainer name="content">
             <referenceBlock name="vault.cards.list">
                 <block class="Adyen\Payment\Block\Customer\CardRenderer" name="adyen.card.renderer"
-                       template="Magento_Vault::customer_account/credit_card.phtml"/>
+                       template="Magento_Vault::customer_account/credit_card.phtml" cacheable="false"/>
```
msilvagarcia: Isn't this an overkill?

PR #499, opened 2019-08-12
Treating notification issue

rqsakai:
**Description**
We were having issues where a few (less than 5) notification failed to integrate for multiple reasons, and this killed the entire batch that got locked with processing = 1.

We added an exception validation that rolls back the status fo the failed transaction


rikterbeek:
Hi @rqsakai ,

Many thanks for this PR! We will validate it and merge it into the code.

Regards,
Rik
Adyen

PR #501, opened 2019-08-14
Expose last 4 digits of API keys

msilvagarcia:
To help debug issues, expose the last 4 digits of API keys.

### DIFF
--- a/Test/Unit/Comment/ApiKeyEndingTest.php
+++ b/Test/Unit/Comment/ApiKeyEndingTest.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Adyen\Payment\Test\Comment;
+
+use Adyen\Payment\Model\Comment\ApiKeyEnding;
+use Magento\Framework\Encryption\Encryptor;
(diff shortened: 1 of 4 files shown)


Test/Unit/Comment/ApiKeyEndingTest.php
```diff
@@ -0,0 +1,33 @@
...
+            ['4321', '1234'],
+            ['asdfasdfasdf', 'qwerqwerqwer']
+        ];
+        $encryptor->method('decrypt')
+            ->will($this->returnValueMap($map));
+
+        $this->apiKeyEndingComment = new ApiKeyEnding($encryptor);
+    }
+
+    public function testSomething()
```
cyattilakiss: pleeease rename it "toSomethingBetter" ;) so when we read the result we are going to have a clue :D

msilvagarcia: Thank you! I meant to rename that before pushing 😆

PR #503, opened 2019-08-20
PW-1634 Retrieve shopper email during guest checkout wth virtual product

cyattilakiss:
**Description**
Add guest email address to card payments when only that email address is
available

### DIFF
--- a/Helper/Requests.php
+++ b/Helper/Requests.php
@@ -67,12 +67,15 @@ class Requests extends AbstractHelper
     }
 
     /**
-     * @param $request
+     * @param array $request
      * @param int $customerId
      * @param $billingAddress
-     * @return mixed
+     * @param $storeId
+     * @param null $payment
+     * @param null $payload
+     * @return array
      */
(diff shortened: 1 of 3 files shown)


Helper/Requests.php
```diff
@@ -67,12 +67,15 @@ public function buildMerchantAccountData($request = [], $paymentMethod, $storeId
...
      * @param int $customerId
      * @param $billingAddress
-     * @return mixed
+     * @param $storeId
+     * @param null $payment
+     * @param null $payload
+     * @return array
      */
-    public function buildCustomerData($request = [], $customerId = 0, $billingAddress, $storeId, $payment = null)
+    public function buildCustomerData($request = [], $customerId = 0, $billingAddress, $storeId, $payment = null, $payload = null)
```
AlexandrosMor: Is the payment parameter needed?

msilvagarcia: It seems to be used in line 85

cyattilakiss: yes we need it currently

PR #511, opened 2019-08-27
Fix Adyen\DataTest

msilvagarcia:
This one is a bit more complex. DataTest depends on some external classes from Magento, so I set up a config and a bootstrap files for testing. While I was at this, I also made the dependencies a bit more strict.


composer.json
```diff
@@ -16,7 +16,11 @@
   "require": {
     "adyen/php-api-library": "~2.1",
     "magento/framework": ">=101.0.8 <102 || >=102.0.1",
-    "magento/module-vault": "101.*"
+    "magento/module-vault": "101.*",
+    "php": "~7.1.3||~7.2.0||~7.3.0"
```
rikterbeek: As we support Magento 2.2.8 and higher the minimum php version we need to support is 7.0.13 https://github.com/magento/magento2/blob/2.2.8/composer.json#L11

composer.json
```diff
@@ -30,5 +34,11 @@
     "psr-4": {
       "Adyen\\Payment\\Tests\\": "tests/"
     }
+  },
+  "scripts": {
+    "test": [
+      "Composer\\Config::disableProcessTimeout",
+      "vendor/bin/phpunit -c Test/phpunit.xml"
```
rikterbeek: what are we testing here?

msilvagarcia: The test suites are configured in the phpunit.xml file. It includes everything under the Test directory.

composer.json
msilvagarcia: Just pushed a commit to match with Magento 2.2.8 :)

PR #517, opened 2019-09-05
Fix cartebancaire_small.png non existent

NickdeK:
**Description**
Add missing `cartebancaire_small.png` image that can be requested in https://github.com/Adyen/adyen-magento2/blob/develop/Model/Ui/TokenUiComponentProvider.php#L68

If the `cartebancaire_small.png` image is missing it will result in the following error
```
Exception #0 (Magento\Framework\View\Asset\File\NotFoundException): Unable to resolve the source file for 'frontend/Elephant/vanmoof/en_US/Adyen_Payment/images/logos/cartebancaire_small.png'
```

**Tested scenarios**
1. Go to the checkout
2. Login as a user that has previously payed with cartebancaire
3. Checkout should load normally


cyattilakiss:
Hi @NickdeK ,

Thank you for the PR, we are going to merge it soon.

Best,
Attila
Adyen

PR #521, opened 2019-09-05
Fix admin config backend model

asrar7787:
**Description**
The updated code fixes bug which throws stack trace (in developer mode) in Magento 2 admin area on a fresh Magento2 installation due to empty config value in `core_config_data`:
`Exception #0 (InvalidArgumentException): Unable to unserialize value.`

The affected admin config field id `installments`.

**Tested scenarios**
Once this PR is merged:
1. Go to Admin > Stores > Settings > Configuration > Sales > Payment Methods
2. The page should load as normal


Aleffio:
Hi @asrar7787 ,

Thank you for your PR! we are going to review and merge it.
I wasn't able to reproduce your issue even by removing the fields in the core_config_data table in the database in developer mode, which version of Magento did you experience this issue on?

Thanks,
Alessio
Adyen

Model/Config/Backend/Installments.php
```diff
@@ -112,8 +112,8 @@ public function beforeSave()
     protected function _afterLoad()
     {
         $value = $this->getValue();
-        $value = $this->serializer->unserialize($value);
-        if (is_array($value)) {
+        if (!empty($value) && is_array($value)) {
```
cyattilakiss: the is_array check should happen after the unserialize because the $value which is serialized will always be a string.

cyattilakiss:
Hi @asrar7787 ,

Could you please rebase your branch to the latest develop?

Best,
Attila
Adyen

asrar7787:
hey @Aleffio,

This has been happening on Magento Commerce 2.3.2.

Model/Config/Backend/Installments.php
asrar7787: yes this now make sense but when $value returns null there’s no empty check for it.

cyattilakiss: I agree with the empty check. This is what I meant:

```
if (!empty($value)) {
    $value = $this->serializer->unserialize($value);
    if (is_array($value)) {
        $value = $this->encodeArrayFieldValue($value);
        $this->setValue($value);
    }
}
```

asrar7787: Correct - That should do!
However that will increase your method’s, eventually class’ cognitive complexities over all.

May be - we can return if $value is empty in the first place and focus on executing other statements?

cyattilakiss: I'm okay with that.

msilvagarcia:
Thanks for your help, @asrar7787 .

Cheers,
Marcos
Adyen

PR #528, opened 2019-09-30
Remove extra greater character

phaelfp:
**Description**
I noticed that on the order detail page, when the payment method is bill of exchange, the bill of thumb print link has a "greater" sign before the link, I think it is an extra character and should be removed.

**Tested scenarios**

**Fixed issue**:


rikterbeek:
Hi @phaelfp,

Thanks for spotting this error. We will merge it.

Regards,
Rik
Adyen

PR #534, opened 2019-10-02
Do not remove credit card method when using api retrieve-adyen-paymen…

jimmy-lion:
…t-methods endpoint

**Description**
When retrieving payment methods using /V1/guest-carts/:cartId/retrieve-adyen-payment-methods, the credit card payment method is filtered at this point:

https://github.com/Adyen/adyen-magento2/blob/develop/Helper/PaymentMethods.php#L219

This is the reason that Adyen support gave us for this:

_You are right, the backend will filter out a few payment methods, because in the default implementation we have a setting for credit cards separately and we render it on the frontend if we have the credit card payment method enabled._

The proposed solution is a new parameter $includeSchemeType which is set to true when retrieving payment methods array from api.

### DIFF
--- a/Api/AdyenPaymentProcessInterface.php
+++ b/Api/AdyenPaymentProcessInterface.php
@@ -27,7 +27,7 @@ namespace Adyen\Payment\Api;
 interface AdyenPaymentProcessInterface
 {
     /**
-     * @param string $payload
+     * @param mixed $payload
      * @return string
      */
     public function initiate($payload);
--- a/Helper/PaymentMethods.php
+++ b/Helper/PaymentMethods.php
@@ -129,13 +129,13 @@ class PaymentMethods extends AbstractHelper
     /**
      * {@inheritDoc}
      */
-    public function getPaymentMethods($quoteId, $country = null)
+    public function getPaymentMethods($quoteId, $country = null, $includeSchemeType = false)
     {
         // get quote from quoteId
         $quote = $this->quoteRepository->getActive($quoteId);
         $store = $quote->getStore();
 
-        $paymentMethods = $this->addHppMethodsToConfig($store, $country);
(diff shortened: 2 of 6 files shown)


cyattilakiss:
HI @jimmy-lion ,

Thank you for the pull request we really appreciate it.
We have already looked into it but the proposed solution would break a bit the single responsibility of the API. The approach that we would like to follow is removing the filtering from the backend and do the filtering only on the frontend, therefore headless and default implementations can also use the same API the same way.

The filtering is done in this section and also in the fetchAlternativeMethods function which needs to be moved to the frontend: https://github.com/Adyen/adyen-magento2/blob/7c6eec150561fc0b257091a1af5e368ba1830434/Helper/PaymentMethods.php#L157-L171

The filtering could be done here on the frontend: https://github.com/Adyen/adyen-magento2/blob/15c30cbe84b36cceaa2b30dae8037323b1c5ad2c/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js#L95

Please let me know what do you think.

Best,
Attila
Adyen

jimmy-lion:
ok, now I understand what you want.  I will try to make a new pull request for it this weekend!

thanks

msilvagarcia:
Hey @jimmy-lion 

Thank you for putting effort in this issue. It's always great to see community involvement.

However, it seems like we are failing in putting the message across. With that in mind, we decided to prioritize this issue higher in our backlog and put it out in a release soon.

We will try to keep you in the loop, either here or through Zendesk. But I also ask you to keep an eye open in our release notes.

We understand you're in a pressure, but please understand that we also need to maintain this in the future.

Kind regards,
Marcos
Adyen

rikterbeek:
Hi @jimmy-lion,

We solved the issue with this PR:
https://github.com/Adyen/adyen-magento2/pull/591

Regards,
Rik
Adyen

PR #536, opened 2019-10-09
Fix for IE11 and older browsers

Aleffio:


view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -141,7 +141,7 @@ define(
                 // retrieve new payment methods if country code changed
                 if (shippingAddressCountryCode != currentShippingAddressCountryCode) {
                     fullScreenLoader.startLoader();
-                    adyenPaymentService.retrieveAvailablePaymentMethods();
+                    adyenPaymentService.retrieveAvailablePaymentMethods(null);
```
KadoBOT: I rather not pass anything and do `if(callback)` instead of `if(callback != null)`

rikterbeek: @cyattilakiss can you change that ?

Issue #523, opened 2019-09-23
Most of the translations are not working.


laurakezbere:
**Magento version**: 2.3.2
**Plugin version**: 4.5.1
**Description**

Not all translations are working for me. I wrote a mixin for adyen-cc-method.js, rewrote initialize function and replaced this:
```
this.checkout = new AdyenCheckout({
    locale: this.getLocale(),
    risk: {
        enabled: false
    }
});
```
with this:
```
translations[this.getLocale()] = {
    'creditCard.numberField.placeholder': $.mage.__('Card number'),
    'creditCard.holderName.placeholder': $.mage.__('Name on card'),
    'creditCard.cvcField.placeholder': $.mage.__('CVC'),
    'creditCard.expiryDateField.month.placeholder': $.mage.__('Month'),
    'creditCard.expiryDateField.year.placeholder': $.mage.__('Year')
};

// initialize adyen component for general use
this.checkout = new AdyenCheckout({
    locale: this.getLocale(),
    risk: {
        enabled: false
    },
    translations: translations
});
```
To do this I was referencing this documentation - https://docs.adyen.com/checkout/web-sdk/customization/localization Checkout fields section.
So the issue is that this worked only for `creditCard.holderName.placeholder` field and I'm not sure if this is an issue on my side or the documentation has provided wrong keys.

cyattilakiss:
Hi @laurakezbere ,

Thank you for raising this issue. Indeed, we are currently using the checkout component version 3.0.0 and this issue that you mentioned was fixed in the v.3.2.0.
We are going to implement that version asap.

Best,
Attila
Adyen

cyattilakiss:
Hi @laurakezbere ,

In the linked PR above we bumped the version of the checkout component to 3.2.0. You can use this branch or wait for the next release when this will be released as well.

Best,
Attila
Adyen

cyattilakiss:
Hi @laurakezbere ,

We have released version 5.0.0 including the fix for your issue.
Let me know if you face any issue regarding to this.

Best,
Attila
Adyen

PR #541, opened 2019-10-10
[PW-1746] Upgrade checkout component to 3.2.0

cyattilakiss:
**Description**
Update checkout component's version to 3.2.0
Update UI
Load checkout component js in head instead of body

**Fixed issue**:  
#523

### DIFF
--- a/view/frontend/web/css/styles.css
+++ b/view/frontend/web/css/styles.css
@@ -206,10 +206,15 @@
 /* General styling for checkout */
 
 .payment-method-content .field input, .payment-method-content .field select {
-    max-width: 168px;
+    max-width: 250px;
 }
 
 .adyen-checkout__link adyen-checkout__link--more-information {
+    margin-bottom: 10px;
+}
+
+.checkout-component-dock {
+    max-width: 250px;
     margin-bottom: 10px;
 }
 
@@ -217,7 +222,6 @@
     margin-top: 10px;
     display: none;
 }
-
 
 /* Checkout component Adyen v3.2.0 styling start */
 .adyen-checkout__field{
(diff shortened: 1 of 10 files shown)


view/frontend/web/css/styles.css
```diff
@@ -206,19 +206,23 @@
...
 
 .adyen-checkout__link adyen-checkout__link--more-information {
     margin-bottom: 10px;
 }
 
+.checkout-component-dock {
+    max-width: 250px;
+    margin-bottom: 10px;
+}
+
```
msilvagarcia: Shouldn't this be responsibility of the merchant to design?

view/frontend/web/template/payment/google-pay-form.html
```diff
@@ -51,7 +51,7 @@
             <!--/ko-->
         </div>
         <div class="field number googlePay">
-            <div afterRender="renderGooglePay()" data-bind="attr: { id: 'googlePay'}">
+            <div class="checkout-component-dock" afterRender="renderGooglePay()" data-bind="attr: { id: 'googlePay'}">
```
rikterbeek: what is the class checkout-component-dock doing ?

cyattilakiss: Since now the component is 100% wide we need a container div to control the size of the component. The div was already there so I just added a class for it to be able to style it.

view/frontend/web/css/styles.css
cyattilakiss: Yes it is, but we usually make it just a bit nicer so merchants still get a nice out of the box solution

msilvagarcia: But it also makes it harder for them to customize it, no?

Issue #508, opened 2019-08-23
Adyen recording same billing and shipping address as shipping address when they are different in Magento


sskharate:
**Magento version**: Magento Commerce Cloud 2.2.9 
**Plugin version**: 4.2.1
**Description**
We are seeing some cases where - On Adyen, the billing and shipping address are both same as the order shipping address when they are different in Magento Order.

Is this a known issue? and is there a fix available?

Thanks
Sagar

cyattilakiss:
Hi @sskharate ,

We are retrieving the billing address and shipping address from the quote but it can be that it is still not updated when we retrieve the data in the backend. We are going to look into it and inform you. Meanwhile if you have any solution we really appreciate the pull request :)

Best,
Attila
Adyen

sskharate:
Hi @cyattilakiss 

Is there a fix for this now?

Thanks
Sagar

cyattilakiss:
Hi @sskharate ,

Not yet, but we are going to pick this item up in the next sprint. and hopefully going to fix it quick.

Best,
Attila
Adyen

cyattilakiss:
Hi @sskharate ,

I will close this PR since it is already fixed in the version 5 and above. See my comment in the related PR: https://github.com/Adyen/adyen-magento2/pull/543#issuecomment-547443754

Have a nice day!
Best,
Attila
Adyen

PR #543, opened 2019-10-14
[PW-1667] Adyen recording same billing and shipping address as shipping address when they are different in Magento

cyattilakiss:
**Description**
Update billing address before sending the payments request to Adyen

**Fixed issue**:  
#508

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
@@ -37,9 +37,11 @@ define(
         'Magento_Paypal/js/action/set-payment-method',
         'Magento_Checkout/js/action/select-payment-method',
         'Adyen_Payment/js/threeds2-js-utils',
-        'Adyen_Payment/js/model/threeds2'
+        'Adyen_Payment/js/model/threeds2',
+        'Adyen_Payment/js/action/set-billing-address',
+        'Magento_Ui/js/model/messageList'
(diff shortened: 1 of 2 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -335,17 +337,19 @@ define(
                     event.preventDefault();
                 }
 
-                if (this.validate() && additionalValidators.validate()) {
-                    fullScreenLoader.startLoader();
-                    self.isPlaceOrderActionAllowed(false);
+                setBillingAddressAction(globalMessageList).then(function() {
```
rikterbeek: the placeOrder does already setting the billingAddress does this not execute this logic twice ?
This will decrease performance as 2 api calls are needed on the place order button. Would it not be better to pass the billingAddress into the API call ? as well maybe better for supporting PWA apps ?

cyattilakiss: Yes the original placeOrder flow will update it again but only after the adyen /payments call is done with the old billing address. Why I choose this solution is because even if we get redirect (3ds1) back from Adyen we save the billing address in the quote and if something went wrong the user will still have the new billing address already saved. 

What I can do to optimise it is to update the quote in our custom controller before making the /payments call and then we have 1 less ajax call but the same functionality. 
I will do the changes soon.

Issue #537, opened 2019-10-09
Urgent | Broken 3d secure due to requestOptions


ruthgeridema:
**Magento version**: 2.3.2
**Plugin version**: latest
**Description**
[PW-1737] Add idempotency key (#531) breaks 3D secure payments. It returns '3d secure not valid' while the 3d secure IS valid.

Logs are not helping to solve this issue. If we remove the additional $requestOptions the checkout works as expected

cyattilakiss:
Hi @ruthgeridema ,

Thank you for raising the issue, I'm checking it immediately.
Can you help me with providing, which test case breaks?
3DS1 or 3DS2?

Thanks,
Attila
Adyen

ruthgeridema:
Hi @cyattilakiss ,

[2019-10-09 11:13:07] AdyenLoggerTest.ADYEN_RESULT: 3D secure is active {"is_exception":false} []
[2019-10-09 11:13:07] AdyenLoggerTest.ADYEN_RESULT: Customer was redirected to bank for 3D-secure validation. {"is_exception":false} []

cyattilakiss:
Thanks.
We found the issue, working on the hotfix right now.

Best,
Attila
Adyen

cyattilakiss:
Hi @ruthgeridema 

You can find the hotfix in the https://github.com/Adyen/adyen-magento2/releases/tag/4.5.4
Thank you again for raising the issue.

Best,
Attila
Adyen

ryanpalmerweb:
Hi Everyone,

Just to follow up on this, we encountered this issue and applied the fix, however a number of orders are still failing and it looks like this is because the failed order increment id's which have been logged within Adyen are being reused.

We are currently in the process of manually correcting quotes with a reserved increment id which has failed

Thanks

cyattilakiss:
Hi @ryanpalmerweb ,

The order id which has failed during payment should be skipped and a new increment id should be used for each new payment attempt. I just tested this specific scenario and is working on my test instance.

What do you think how it can be that Magento assigns old order ids as new increment id to new orders?

Maybe I miss something, please let me know if you have more information or further questions.

Best,
Attila
Adyen

ryanpalmerweb:
Hi Attila,

I think this is specifically for orders which had the broken 3D secure issue, I'm not sure on the exact audit trail but we can confirm that the order ids which were failing yesterday have been re-used by Magento, however the database has not been rolled back or anything like that

[...]

PR #548, opened 2019-10-16
[PW-1782] Use idempotency key only for redirect payment methods

cyattilakiss:
**Description**
Only add idempotency key if payment method is adyen_hpp

**Tested scenarios**
1
refused card payment 
authorised hpp payment (ideal)
2
refused card payment
authorised card payment
3
refused hpp payment (ideal)
authorised hpp payment (ideal)
4
refused direct alternative payment method (afterpay)
authorised direct alternative payment method (afterpay)

**Fixed issue**:
#537

### DIFF
--- a/Gateway/Http/Client/TransactionAuthorization.php
+++ b/Gateway/Http/Client/TransactionAuthorization.php
@@ -70,7 +70,12 @@ class TransactionAuthorization implements ClientInterface
     {
         $request = $transferObject->getBody();
 
-        $requestOptions['idempotencyKey'] = $request['reference'];
+        $requestOptions = [];
+
+        if (!empty($request['idempotencyKeyRequired']) && $request['idempotencyKeyRequired'] === true) {
+            $requestOptions['idempotencyKey'] = $request['reference'];
+            unset($requestOptions['idempotencyKey']);
+        }
         
         // call lib
         $service = new \Adyen\Service\Payment($this->_client);
--- a/Gateway/Http/Client/TransactionAuthorization.php
+++ b/Gateway/Http/Client/TransactionAuthorization.php
@@ -69,12 +69,12 @@ class TransactionAuthorization implements ClientInterface
     public function placeRequest(\Magento\Payment\Gateway\Http\TransferInterface $transferObject)
     {
         $request = $transferObject->getBody();
+        $headers = $transferObject->getHeaders();
 
         $requestOptions = [];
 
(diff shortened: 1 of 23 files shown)


Gateway/Http/Client/TransactionAuthorization.php
```diff
@@ -70,7 +70,12 @@ public function placeRequest(\Magento\Payment\Gateway\Http\TransferInterface $tr
     {
         $request = $transferObject->getBody();
 
-        $requestOptions['idempotencyKey'] = $request['reference'];
+        $requestOptions = [];
+
+        if (!empty($request['idempotencyKeyRequired']) && $request['idempotencyKeyRequired'] === true) {
+            $requestOptions['idempotencyKey'] = $request['reference'];
+            unset($requestOptions['idempotencyKey']);
```
rikterbeek: why do we first set it and then unset ? probably you want to unset idempotencyKeyRequired right ?  I don't think it is nice to add it into the request pure to know if it is needed is there no elegance way to do this ?

Gateway/Http/Client/TransactionPayment.php
```diff
@@ -65,7 +65,13 @@ public function placeRequest(\Magento\Payment\Gateway\Http\TransferInterface $tr
         $client = $this->adyenHelper->initializeAdyenClient();
 
         $service = new \Adyen\Service\Checkout($client);
-        $requestOptions['idempotencyKey'] = $request['reference'];
+
+        $requestOptions = [];
+
+        if (!empty($request['idempotencyKeyRequired']) && $request['idempotencyKeyRequired'] === true) {
+            $requestOptions['idempotencyKey'] = $request['reference'];
+            unset($requestOptions['idempotencyKey']);
```
rikterbeek: same here ?

Gateway/Http/Client/TransactionAuthorization.php
cyattilakiss: Yes indeed, by mistake I unset the wrong one. I will change that.
I think there is a bit more elegant way. I can use the TransferInterface's header property instead of the body to pass it since at the end it is a header setting.

cyattilakiss: Made the changes

Gateway/Http/Client/TransactionPayment.php
cyattilakiss: Made the changes

PR #550, opened 2019-10-16
Shopper ip address fall back for getRemoteIp()

cyattilakiss:
**Description**
Order->getXForwardedFor() can return null, in this case fall back for
Order->getRemoteIp()

**Tested scenarios**
afterpay where shopper ip is required

### DIFF
--- a/Gateway/Request/CustomerIpDataBuilder.php
+++ b/Gateway/Request/CustomerIpDataBuilder.php
@@ -39,7 +39,8 @@ class CustomerIpDataBuilder implements BuilderInterface
      *
      * @param \Adyen\Payment\Helper\Requests $adyenRequestsHelper
      */
-    public function __construct(\Adyen\Payment\Helper\Requests $adyenRequestsHelper)
+    public function __construct(
+        \Adyen\Payment\Helper\Requests $adyenRequestsHelper)
     {
         $this->adyenRequestsHelper = $adyenRequestsHelper;
     }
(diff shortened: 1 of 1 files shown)


Gateway/Request/CustomerIpDataBuilder.php
```diff
@@ -52,7 +53,12 @@ public function build(array $buildSubject)
     {
         /** @var \Magento\Payment\Gateway\Data\PaymentDataObject $paymentDataObject */
         $paymentDataObject = \Magento\Payment\Gateway\Helper\SubjectReader::readPayment($buildSubject);
-        $shopperIp = $paymentDataObject->getPayment()->getOrder()->getXForwardedFor();
+        $order = $paymentDataObject->getPayment()->getOrder();
+        $shopperIp = $order->getXForwardedFor();
+
+        if (empty($shopperIp)) {
+            $shopperIp = $order->getRemoteIp();
```
rikterbeek: are we sure that this does not result in a server IP ? because if that is the case it is better to not send a value!

cyattilakiss: Without ip address sent, afterpay results in an internal server error so looks like shopperIp is a required field.
Can we maybe send like a default value in this case, does it make sense?

Gateway/Request/CustomerIpDataBuilder.php
```diff
@@ -52,7 +53,12 @@ public function build(array $buildSubject)
     {
         /** @var \Magento\Payment\Gateway\Data\PaymentDataObject $paymentDataObject */
         $paymentDataObject = \Magento\Payment\Gateway\Helper\SubjectReader::readPayment($buildSubject);
-        $shopperIp = $paymentDataObject->getPayment()->getOrder()->getXForwardedFor();
+        $order = $paymentDataObject->getPayment()->getOrder();
+        $shopperIp = $order->getXForwardedFor();
```
msilvagarcia:
```suggestion
        $shopperIp = $order-> getRemoteIp();
```
`getRemoteIp` actually uses the `$_SERVER['REMOTE_ADDR']` server variable, which is more reliable as a client IP address.

Gateway/Request/CustomerIpDataBuilder.php
msilvagarcia:
```suggestion
            $shopperIp = $order-> getXForwardedFor();
```

`getXForwardedFor` uses the `$_SERVER['HTTP_X_FORWARDED_FOR']` server variable, which can be overwritten by the client and should not be trusted.

PR #552, opened 2019-10-18
Add a stopLoader when the validateThreeDS2OrPlaceOrder method fails

maurosempere:
**Description**
When we make a payment with a saved card and the CVC is not correct, the full-screen loader is not removed.

**Tested scenarios**
1. Configure Stored Payment Methods
2. Place an order, paying with a 3ds card and saving it for later payments
3. Re-order and pay with your saved card
4. Insert a wrong CVC/CVV
5. Complete the order to force the rejected payment message
6. Here will have the loading screen permanently

image-capture


olivermontes:
nice hunt @maurosempere :)

PR #579, opened 2019-11-12
[PW-1682] order management ui

AlexandrosMor:
**Description**

**Tested scenarios**

**Fixed issue**:


rikterbeek:
Hi @AlexandrosMor please change the xml naming so it is corresponding to the UI section.

PR #580, opened 2019-11-12
Add afterSavePaymentInformationAndPlaceOrder plugin for guest checkout

cyattilakiss:
**Description**
3DS2 challenge is not shown for guest users because the plugin only covers logged in users but not the guest users.

### DIFF
--- a/Plugin/GuestPaymentInformationManagement.php
+++ b/Plugin/GuestPaymentInformationManagement.php
@@ -44,7 +44,8 @@ class GuestPaymentInformationManagement
     protected $adyenHelper;
 
     /**
-     * PaymentInformationManagement constructor.
+     * GuestPaymentInformationManagement constructor.
+     *
      * @param \Magento\Sales\Api\OrderRepositoryInterface $orderRepository
      * @param \Adyen\Payment\Logger\AdyenLogger $adyenLogger
      * @param \Adyen\Payment\Helper\Data $adyenHelper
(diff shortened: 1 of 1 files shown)


Plugin/GuestPaymentInformationManagement.php
```diff
@@ -0,0 +1,92 @@
...
+    public function afterSavePaymentInformationAndPlaceOrder(
+        \Magento\Checkout\Api\GuestPaymentInformationManagementInterface $subject,
+        $result
+    ) {
+        try {
+            $order = $this->orderRepository->get($result);
+            $payment = $order->getPayment();
+
+            if ($payment->getMethod() === AdyenCcConfigProvider::CODE ||
+                $payment->getMethod() === AdyenOneclickConfigProvider::CODE
```
Aleffio:
```suggestion
```

Plugin/GuestPaymentInformationManagement.php
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public function afterSavePaymentInformationAndPlaceOrder(
+        \Magento\Checkout\Api\GuestPaymentInformationManagementInterface $subject,
+        $result
+    ) {
+        try {
+            $order = $this->orderRepository->get($result);
+            $payment = $order->getPayment();
+
+            if ($payment->getMethod() === AdyenCcConfigProvider::CODE ||
```
Aleffio:
```suggestion
            if ($payment->getMethod() === AdyenCcConfigProvider::CODE 
```

PR #582, opened 2019-11-13
3ds2 fix guest checkout add sortOrder

cyattilakiss:


rikterbeek:
Let's make it 90

cyattilakiss:
Just reread the documentation and if we want to run the after method of the plugin the last among other plugins we have to use a small number so I will just keep it as it is. The 10 is a best practise so if anyone still wants to run their plugin after ours they can still go below.

PR #583, opened 2019-11-13
PW-1688: UI Alternative Payment Methods => Local Payment Methods

Aleffio:


etc/adminhtml/system/adyen_hpp.xml
```diff
@@ -33,44 +33,55 @@
...
-            <label>Klarna\RatePay\Afterpay Settings</label>
+        <group id="adyen_hpp_store_payment_details" translate="label" showInDefault="1" showInWebsite="1" showInStore="1" sortOrder="100">
+            <label>Store payment details options</label>
+            <comment>Adyen can securely store the payment details for select local payment methods of your shoppers, allowing you to accept recurring payments.</comment>
             <frontend_model>Magento\Config\Block\System\Config\Form\Fieldset</frontend_model>
-            <field id="ratepay_id" translate="label" type="text" sortOrder="50" showInDefault="1" showInWebsite="1" showInStore="1">
-                <label>RatePAY Device Ident SId</label>
-                <tooltip>Unique RatePAY Id provided by RatePAY integration consultant</tooltip>
-                <config_path>payment/adyen_hpp/ratepay_id</config_path>
+            <field id="enable_recurring" translate="label" type="select" sortOrder="20" showInDefault="1" showInWebsite="1"
```
cyattilakiss: is this sortOrder correct?

Model/Config/Source/SplitPaymentRefundStrategy.php
```diff
@@ -44,8 +44,8 @@ public function getSplitPaymentRefundStrategies()
     {
         return [
             self::REFUND_FIRST_PAYEMENT_FIRST => __('Refund from first payment first'),
-            self::REFUND_LAST_PAYEMENT_FIRST => 'Refund from last payment first',
-            self::REFUND_ON_RATIO => __('refund based on ratio')
+            self::REFUND_LAST_PAYEMENT_FIRST => ('Refund from last payment first'),
```
msilvagarcia:
```suggestion
            self::REFUND_LAST_PAYEMENT_FIRST => __('Refund from last payment first'),
```

etc/adminhtml/system/adyen_hpp.xml
Aleffio: there is only one field in this group, so doesn't really matter

Model/Config/Source/YesNoBa.php
```diff
@@ -0,0 +1,40 @@
...
+ * Copyright (c) 2019 Adyen B.V.
+ * This file is open source and available under the MIT license.
+ * See the LICENSE file for more info.
+ *
+ * Author: Adyen <magento@adyen.com>
+ */
+
+namespace Adyen\Payment\Model\Config\Source;
+
+class YesNoBa extends \Magento\Config\Model\Config\Source\YesNo
```
msilvagarcia: "Ba" is not a very descriptive name.

rikterbeek: YesNoBillingAgreements

rikterbeek: As this is just a YesNO should we not keep it generic so we can re-use it ?

PR #586, opened 2019-11-15
[PW-1686] Payment methods > Credit cards

AlexandrosMor:
**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/etc/adminhtml/system/adyen_payment_methods.xml
+++ b/etc/adminhtml/system/adyen_payment_methods.xml
@@ -1,0 +1,102 @@
+<?xml version="1.0"?>
+<!--
+/**
+ *                       ######
+ *                       ######
+ * ############    ####( ######  #####. ######  ############   ############
+ * #############  #####( ######  #####. ######  #############  #############
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen Payment module (https://www.adyen.com/)
+ *
+ * Copyright (c) 2015 Adyen BV (https://www.adyen.com/)
+ * See LICENSE.txt for license details.
+ *
+ * Author: Adyen <magento@adyen.com>
+ */
+-->
+<include xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Config:etc/system_include.xsd">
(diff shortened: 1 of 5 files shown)


etc/adminhtml/system/adyen_payment_methods.xml
```diff
@@ -0,0 +1,102 @@
...
+            <field id="allowspecific" translate="label" type="allowspecific" sortOrder="20" showInDefault="1"
+                   showInWebsite="1" showInStore="0">
+                <label>Payment from Applicable Countries</label>
+                <source_model>Magento\Payment\Model\Config\Source\Allspecificcountries</source_model>
+                <config_path>payment/adyen_cc/allowspecific</config_path>
+            </field>
+            <field id="specificcountry" translate="label" type="multiselect" sortOrder="30" showInDefault="1"
+                   showInWebsite="1" showInStore="0">
+                <label>Payment from Specific Countries</label>
+                <source_model>Magento\Directory\Model\Config\Source\Country</source_model>
```
Aleffio: You can use the <depends> tag here, check https://github.com/Adyen/adyen-magento2/blob/b5cb60d258810ae41dd714592a0cd83519cc9021/etc/adminhtml/system/adyen_hpp.xml#L82

etc/adminhtml/system/adyen_payment_methods.xml
```diff
@@ -0,0 +1,105 @@
...
+            <field id="active" translate="label" type="select" sortOrder="10" showInDefault="1" showInWebsite="1"
+                   showInStore="1">
+                <label>Enabled</label>
+                <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                <config_path>payment/adyen_cc/active</config_path>
+            </field>
+            <field id="title" translate="label" type="text" sortOrder="20" showInDefault="1" showInWebsite="1"
+                   showInStore="1">
+                <label>Title</label>
+                <config_path>payment/adyen_cc/title</config_path>
```
Aleffio: Not sure if the path needs to be changed to adyen_payment_methods

cyattilakiss: Alexandros please check if you change the value and save it if it still works, I think it will without any problem. If you would change the config_path you would need to write a migration script because it would change the setting in the db as well.

etc/adminhtml/system/adyen_payment_methods.xml
```diff
@@ -0,0 +1,105 @@
...
+                   showInStore="1">
+                <label>Store one-click contracts</label>
+                <tooltip>"To complete the transaction, the shopper needs to enter their CVC/CVV"</tooltip>
+                <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                <config_path>payment/adyen_abstract/enable_oneclick</config_path>
+            </field>
+            <field id="store_recurring" translate="label" type="select" sortOrder="20" showInDefault="1"
+                   showInWebsite="1"
+                   showInStore="1">
+                <label>Sotre recurring contracts</label>
```
msilvagarcia:
```suggestion
                <label>Store recurring contracts</label>
```

AlexandrosMor: @msilvagarcia

PR #591, opened 2019-11-22
[PW-1835] paymentMethods endpoint returns all payment methods. Filter…

rikterbeek:
…ing happens in front-end validation and removed unused helpers.

**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
@@ -43,6 +43,7 @@ define(
         var paymentMethod = ko.observable(null);
         var messageComponents;
         var shippingAddressCountryCode = quote.shippingAddress().countryId;
+        var unsupportedPaymentMethods = ['scheme', 'boleto', 'bcmc_mobile_QR', 'wechatpay'];
         /**
          * Shareble adyen checkout component
          * @type {AdyenCheckout}
@@ -70,7 +71,7 @@ define(
                         'bankLocationId'
                     ]);
                 return this;
-            },initialize: function () {
+            }, initialize: function () {
 
                 var self = this;
                 this._super();
@@ -87,9 +88,9 @@ define(
 
                 // reset variable:
                 adyenPaymentService.setPaymentMethods();
-                
-                adyenPaymentService.retrieveAvailablePaymentMethods(function() {
-                    let paymentMethods = adyenPaymentService.getAvailablePaymentMethods();
+
+                adyenPaymentService.retrieveAvailablePaymentMethods(function () {
+                    var paymentMethods = adyenPaymentService.getAvailablePaymentMethods();
                     if (JSON.stringify(paymentMethods).indexOf("ratepay") > -1) {
                         var ratePayId = window.checkoutConfig.payment.adyenHpp.ratePayId;
                         var dfValueRatePay = self.getRatePayDeviceIdentToken();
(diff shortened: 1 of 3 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -145,14 +146,19 @@ define(
 
                 var paymentMethods = adyenPaymentService.getAvailablePaymentMethods();
 
-                var paymentList = _.map(paymentMethods, function (value) {
+                var paymentList = _.reduce(paymentMethods, function (acc, value) {
```
msilvagarcia:
```suggestion
                var paymentList = _.reduce(paymentMethods, function (accumulator, value) {
```

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -145,14 +146,19 @@ define(
...
+
+                    if (!self.isPaymentMethodSupported(value.type)) {
+                        return acc;
```
msilvagarcia:
```suggestion
                        return accumulator;
```

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -502,11 +508,24 @@ define(
                         result.bankLocationId = ko.observable(null);
                     }
 
-                    return result;
-                });
+                    acc.push(result);
```
msilvagarcia:
```suggestion
                    accumulator.push(result);
```

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -502,11 +508,24 @@ define(
...
+                    return acc;
```
msilvagarcia:
```suggestion
                    return accumulator;
```

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -502,11 +508,24 @@ define(
...
+                if (paymentMethod == 'wechatpayWeb') {
+                    return true;
+                }
+                for (var i = 0; i < unsupportedPaymentMethods.length; i++) {
+                    var match =  paymentMethod.match(unsupportedPaymentMethods[i]);
+                    if(match) {
+                        return false;
+                    }
+                }
+                return true;
```
msilvagarcia: We can make this easier to understand by replacing this block with:

```js
return unsupportedPaymentMethods.indexOf(paymentMethod) !== -1;
```

KadoBOT: This would match the whole word only. "boleto" and "boletobancario" needs to be treated as the same and would fail in your scenario.

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
KadoBOT: I like `acc` more tbh, and I see others using `acc` more often as well.

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
msilvagarcia: Ah, I see. That makes sense, but I'd provide some indication of why this is being done then.

rikterbeek: @msilvagarcia fair point added some comments.

ScottPowell14:
Hi @rikterbeek @msilvagarcia can you explain the use case of this PR? Doesn't really seem to make sense that card payment method should be returned if card component will be mounted

msilvagarcia:
Hey @ScottPowell14 

While I agree this is poorly documented in the Pull Request itself, there's some documentation in the YouTrack issue. The reason for this Pull Request was to fix an issue with PWA installations.

I removed your comment to get rid of sensitive information.

Cheers!
Marcos
Adyen

PR #593, opened 2019-11-26
PW-1887: fix for pms(afterpaytouch) with checkoutIssuerDirectLinkEnabled

Aleffio:


Gateway/Validator/CheckoutResponseValidator.php
```diff
@@ -151,6 +151,10 @@ public function validate(array $validationSubject)
                         $paymentData = $response['paymentData'];
                     }
 
+                    if (!empty($response['details'])) {
+                        $details = $response['details'];
+                    }
```
rikterbeek: why doing this here ? I only see this being used in the else can we combine it with that if ?

Controller/Process/Result.php
```diff
@@ -389,7 +389,21 @@ protected function validatePayloadAndReturnResponse($response)
...
+            $key = array_search('returnUrlQueryString', $details[0]);
+
+            if ($key !== false) {
+                $request['details']['returnUrlQueryString'] = http_build_query($response);
+            } else {
+                $request["details"] = $response;
+            }
+        } else {
+            $request["details"] = $response;
+        }
```
rikterbeek: a bit concerned about this flow, is this something our merchants need ?

cyattilakiss: Yes unfortunately for checkout directIssuerPages we need to implement this extra step - so currently this is a quick fix for those who use this setting but in the future this account setting can become default and can be that the API will also adjust.
For now I would just use it like this and let's follow up with the checkout team if they are going to make it more generic later.

Gateway/Validator/CheckoutResponseValidator.php
```diff
@@ -184,6 +184,9 @@ public function validate(array $validationSubject)
                         // Flag to show we are in the checkoutAPM flow
                         $payment->setAdditionalInformation('checkoutAPM', true);
 
+                        if (!empty($details)) {
```
rikterbeek: where is this variable declared ?

Controller/Process/Result.php
rikterbeek: We can declare $request["details"] = $response; first and then override it with the if statement. That results in 2 else statements we can remove and cleaner code. What do you think ?

Issue #590, opened 2019-11-21
 savePaymentInformationAndPlaceOrder plugins incorrectly changing return type


ryanpalmerweb:
**Magento version**: 2.3.2-p2
**Plugin version**: 5.0.2
**Description**

This issue manifested in the checkout as the place order button showing various errors such as "The entity that was requested doesn't exist. Verify the entity and try again" and "No such entity with cartId = " - however we found that the order was being placed and a refresh of the checkout would instead show no items in the shopping cart.

<img width="818" alt="Screen Shot 2019-11-21 at 16 08 54" src="https://user-images.githubusercontent.com/7669058/69356051-bf1b4400-0c7a-11ea-8845-3ad986a9dbb3.png">

A stack trace on the exception revealed that another module was throwing a NoSuchEntityException with the order repository being unable to retrieve the entity by order id.

Various third party modules intercept the savePaymentInformationAndPlaceOrder method which is part of the \Magento\Checkout\Api\PaymentInformationManagementInterface class

The original class contains the following definition in the PHP file - take note of the return type

```
    /**
     * Set payment information and place order for a specified cart.
     *
     * @param int $cartId
     * @param \Magento\Quote\Api\Data\PaymentInterface $paymentMethod
     * @param \Magento\Quote\Api\Data\AddressInterface|null $billingAddress
     * @throws \Magento\Framework\Exception\CouldNotSaveException
     * @return int Order ID // RETURN TYPE INT / ORDERID
     */
      public function savePaymentInformationAndPlaceOrder
```

This means that any modules extending this class (including your own!) rely on the method to return an integer containing the new order id.

However, the plugins AdyenPaymentInformationManagementAddPaymentInfo and AdyenGuestPaymentInformationManagementAddPaymentInfo can change the return type to an encoded JSON string, breaking plugins further down the chain.

```
    /**
     * @param \Magento\Checkout\Api\GuestPaymentInformationManagementInterface $subject
     * @param $result
     * @return string
     * @throws \Magento\Framework\Exception\LocalizedException
     */
    public function afterSavePaymentInformationAndPlaceOrder(
        \Magento\Checkout\Api\GuestPaymentInformationManagementInterface $subject,
        $result
    ) {
        try {
            $order = $this->orderRepository->get($result);
            $payment = $order->getPayment();

            if ($payment->getMethod() === AdyenCcConfigProvider::CODE) {
                //THIS CHANGED THE RETURN TYPE
[...]

PR #600, opened 2019-12-12
B: Change sort order of PaymentInformationManagement plugin

Maikel-Koek:
So it always loads last, since it changes the return type and value,
workaround for https://github.com/Adyen/adyen-magento2/issues/590

**Description**

**Tested scenarios**

**Fixed issue**:


cyattilakiss:
Hi @Maikel-Koek ,

Thank you for the pull request.
In the issue you referenced #590 I have an answer for this issue:

> The ones with a version where the sort order works as expected this interceptor should run the last.

So if we change it in the plugin then for those who has an older version where the interceptor sort order works we will break their integration. Therefore I cannot merge this pull request now.
Thank you for your understanding.

Best,
Attila
Adyen

Issue #599, opened 2019-12-11
"3DS2.0 Enabled" configuration setting is not used


galdiuz:
**Magento version**: 2.3.2
**Plugin version**: 5.0.2
**Description**:
The "3DS2.0 Enabled" setting found under "CreditCard API integration" in Magento system config is never used since v5.0.0. The only usage of the setting in the code is in `Adyen\Payment\Helper\Data::isCreditCardThreeDS2Enabled`, but that function is never used.

```
➜  magento git:(master) grep -rni "threeds2_enabled" vendor/adyen
vendor/adyen/module-payment/etc/adminhtml/system/adyen_cc.xml:66:            <config_path>payment/adyen_cc/threeds2_enabled</config_path>
vendor/adyen/module-payment/etc/config.xml:64:                <threeds2_enabled>1</threeds2_enabled>
vendor/adyen/module-payment/Helper/Data.php:1821:        return $this->getAdyenCcConfigDataFlag('threeds2_enabled', $storeId);
➜  magento git:(master) grep -rni "isCreditCardThreeDS2Enabled" vendor/adyen
vendor/adyen/module-payment/Helper/Data.php:1819:    public function isCreditCardThreeDS2Enabled($storeId = null)
```

It used to be used in Adyen\Payment\Model\AdyenPaymentProcess, which was [removed in v5.0.0](https://github.com/Adyen/adyen-magento2/commit/01997eee61bbed0f60fb8ae5cf07ed36c2bb366d#diff-1e83b8f13c9c4345dfc50a69c1718572L167).

While I'm not fully aware of how the payment flow works, one suggestion of where this perhaps should be used is in [`Adyen\Payment\Helper\Requests::buildThreeDS2Data`](https://github.com/Adyen/adyen-magento2/blob/5.0.2/Helper/Requests.php#L307), like so:
```diff
diff --git a/Helper/Requests.php b/Helper/Requests.php
index 9111572..afe9294 100644
--- a/Helper/Requests.php
+++ b/Helper/Requests.php
@@ -304,7 +304,7 @@ class Requests extends AbstractHelper
      */
     public function buildThreeDS2Data($request = [], $additionalData)
     {
-        $request['additionalData']['allow3DS2'] = true;
+        $request['additionalData']['allow3DS2'] = $this->adyenHelper->isCreditCardThreeDS2Enabled();
         $request['origin'] = $this->adyenHelper->getOrigin();
         $request['channel'] = 'web';
[...]

PR #604, opened 2019-12-17
Use 3DS2 toggle configuration fix

cyattilakiss:
**Description**
The 3DS2 toggle setting is ignored since version 5.0.0 because of a
major change. Add the check back in the Request buildThreeDS2Data()

**Fixed issue**:  
#599

**Tested scenarios**
3DS2 enabled
3DS2 disabled


rikterbeek:
@cyattilakiss can you rebase so we can run the integration tests ?

PR #613, opened 2020-01-09
[PW-1693]: Updating configuration UI for Payment methods > Boleto

acampos1916:
* Removed comment/subtitle from fieldset.
* Updated Delivery Days field label.
* Removed tooltip from Delivery Days field.
* Updated default values for title and Delivery Days.


etc/adminhtml/system/adyen_boleto.xml
```diff
@@ -50,8 +50,7 @@
             <config_path>payment/adyen_boleto/boletotypes</config_path>
         </field>
         <field id="delivery_days" translate="label" type="text" sortOrder="50" showInDefault="1" showInWebsite="1" showInStore="0">
-            <label>Days for Delivery</label>
-            <tooltip>How many days to be added to the current date for delivery. Only numbers allowed.</tooltip>
+            <label>Days for delivery and payment</label>
             <frontend_class>validate-number</frontend_class>
             <config_path>payment/adyen_boleto/delivery_days</config_path>
         </field>
```
Aleffio: can you please add in "specificcountry" the dependency on "allowspecific"(lines 69-70)?
like we do here:
https://github.com/Adyen/adyen-magento2/blob/b5cb60d258810ae41dd714592a0cd83519cc9021/etc/adminhtml/system/adyen_hpp.xml#L82-L84
so we show one less field if "allowspecific" is disabled

Helper/Data.php
```diff
@@ -213,9 +213,9 @@ public function __construct(
     public function getRecurringTypes()
     {
         return [
-            \Adyen\Payment\Model\RecurringType::ONECLICK => 'ONECLICK',
-            \Adyen\Payment\Model\RecurringType::ONECLICK_RECURRING => 'ONECLICK,RECURRING',
-            \Adyen\Payment\Model\RecurringType::RECURRING => 'RECURRING'
+            \Adyen\Payment\Model\RecurringType::ONECLICK => 'One-click',
+            \Adyen\Payment\Model\RecurringType::ONECLICK_RECURRING => 'One-click,Recurring',
+            \Adyen\Payment\Model\RecurringType::RECURRING => 'Recurring'
```
rikterbeek: This will break the implementation. These are not labels these are send into the request. We need to send the values ONECLICK, ONECLICK,RECURRING or RECURRING.

Model/RecurringType.php
```diff
@@ -27,9 +27,9 @@ class RecurringType
 {
 
     const NONE = '';
-    const ONECLICK = 'ONECLICK';
-    const ONECLICK_RECURRING = 'ONECLICK,RECURRING';
-    const RECURRING = 'RECURRING';
+    const ONECLICK = 'One-click';
+    const ONECLICK_RECURRING = 'One-click,Recurring';
+    const RECURRING = 'Recurring';
```
rikterbeek: same as above. Would be better to keep one place where we define these constants this would be the best place.

etc/config.xml
```diff
@@ -35,14 +35,18 @@
...
+                <fraud_manual_review_status>pending</fraud_manual_review_status>
+                <send_email_bank_sepa_on_pending>No</send_email_bank_sepa_on_pending>
                 <group>adyen</group>
             </adyen_abstract>
             <adyen_cc>
-                <active>1</active>
+                <active>0</active>
                 <model>AdyenPaymentCcFacade</model>
-                <title>Credit Card</title>
+                <title>Adyen Credit Card</title>
```
rikterbeek: I believe this field will be visible by default to the shoppers. We don't want any adyen branding to shoppers so better leave this as it is.

etc/config.xml
```diff
@@ -173,11 +177,11 @@
             <adyen_boleto>
                 <active>0</active>
                 <model>AdyenPaymentBoletoFacade</model>
-                <title>Boleto</title>
+                <title>Adyen Boleto</title>
```
rikterbeek: Same here as mentioned above.

Issue #610, opened 2020-01-07
The order cannot be confirmed even though the notification arrives


dbolufer:
**Magento version**: 2.2.8
**Plugin version**: 5.0
**Description**

On the production environment we have an order in payment_review status, this order is payed and Adyen has sent the notification.

When the adyen cron process the notification, it raise a "**Unique constraint violation found**" because there is a row on the table `adyen_order_payment` with the same pspreference and the due to UNIQUE KEY restriction the cronjob the raise an exception.

We've solved the error manually:
* DELETE FROM adyen_order_payment where pspreference = 'XXXX'
* UPDATE adyen_notification set done = 0 where pspreference = 'XXXX'
* Wait the cron job execution
* Change order.status to processing because the order.status = payment_authorized

Please take a look, this is the second time in a month with this bug.

**Update** We thought we had a couple of bad orders but looking for info we had realized that we had 21 orders payment_review en magento y processing en adyen.

We are using this query to check this missed orders:
```sql
SELECT * FROM adyen_notification an 
LEFT JOIN adyen_order_payment aop on aop.pspreference = an.pspreference
LEFT JOIN sales_order_payment sop on aop.payment_id = sop.entity_id
LEFT JOIN sales_order so on so.entity_id = sop.parent_id
 WHERE event_code = "AUTHORISATION" AND  so.status = "payment_review";
 ```

This is the log for the error:

```[2020-01-07 16:23:04] AdyenLoggerTest.ADYEN_NOTIFICATION_CRONJOB: Processing notification 22635 {"is_exception":false} []
[2020-01-07 16:23:04] AdyenLoggerTest.ADYEN_NOTIFICATION_CRONJOB: Array
(
    [entity_id] => 22635
    [pspreference] => XXXX
    [merchant_reference] => XXXX
    [event_code] => AUTHORISATION
    [success] => true
    [payment_method] => mc
    [amount_value] => 15800
    [amount_currency] => EUR
    [reason] => 370529:1014:06/2021
    [live] => true
    [additional_data] => a:6:{s:10:"expiryDate";s:7:"06/2021";s:8:"authCode";s:6:"370529";s:11:"cardSummary";s:4:"1014";s:14:"cardHolderName";s:17:"XXXX";s:24:"recurringProcessingModel";s:10:"CardOnFile";s:13:"hmacSignature";s:44:"XXXX=";}
    [done] => 0
    [processing] => 1
    [created_at] => 2020-01-02 17:18:19
)
[...]

Issue #603, opened 2019-12-16
Exception on "sales_order_invoice_pay" event makes Order to be stuck on "Payment Review"


viniciusfabri:
**Magento version**: 2.3.2
**Plugin version**: 4.5.1
**Description**
Steps to reproduce: 
- Create an Observer on _sales_order_invoice_pay_ that throws an exception (\Magento\Framework\Exception\CouldNotSaveException in our case)
- Create an order and pay with Adyen

Expected results:
- Notification is not processed correctly, on the catch of the exception in the Adyen module the database changes are reverted 

Actual results:
- Notification is not processed correctly, but _adyen_order_payment_ entry is still in there, so the next time the cron job tries to process that notification, we get an "Unique constraint violation" in that table because the pspreference is already there

Adyen\Model\Cron::_createInvoice() triggers that event, and a third party module has an observer that sends the Order to a third party service that does some business logic, but throws an exception if it fails (in this case it seems the service was down). I don't think this is the best solution, but I still don't think the Adyen module should put itself in a stuck position if the notification process fails for some reason

cyattilakiss:
Hi @viniciusfabri ,

Thank you for raising the issue. I picked up this item into our backlog to find a solution.

Best,
Attila
Adyen

cyattilakiss:
Hi @viniciusfabri ,

Just merged a PR which will introduce a fix for this issue. We are going to release it in the next release cycle.

Best,
Attila
Adyen

PR #614, opened 2020-01-09
[PW-1993] Fix adyen_order_payment new entry creation issue during cron

cyattilakiss:
**Description**
Add try catch into cron process when new adyen_order_payment entity is added

Turned out that other plugins can intercept into invoice creation and
these are potential sources of exceptions. After inserting a new
adyen_order_payment entity we start the invoice creation process and if
something goes wrong during that process we reprocess the notification in
the next batch. Therefore the adyen_order_payment is already saved and the
next run will violate the unique constraint setting of the table.

This is a quick solution to prevent the process to stop and the order to
get stuck because of this exception.

**Fixed issue**: 
Resolves GitHub issues: #603, #610

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -1351,20 +1351,28 @@ class Cron
         // validate if amount is total amount
         $orderCurrencyCode = $this->_order->getOrderCurrencyCode();
         $amount = $this->_adyenHelper->originalAmount($this->_value, $orderCurrencyCode);
-
-        // add to order payment
-        $date = new \DateTime();
-        $this->_adyenOrderPaymentFactory->create()
-            ->setPspreference($this->_pspReference)
-            ->setMerchantReference($this->_merchantReference)
-            ->setPaymentId($paymentObj->getId())
-            ->setPaymentMethod($this->_paymentMethod)
-            ->setAmount($amount)
-            ->setTotalRefunded(0)
-            ->setCreatedAt($date)
-            ->setUpdatedAt($date)
-            ->save();
-
+        
+        try {
+            // add to order payment
+            $date = new \DateTime();
+            $this->_adyenOrderPaymentFactory->create()
+                ->setPspreference($this->_pspReference)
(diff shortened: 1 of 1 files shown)


Model/Cron.php
```diff
@@ -1351,20 +1351,28 @@ protected function _prepareInvoice()
...
+                ->setTotalRefunded(0)
+                ->setCreatedAt($date)
+                ->setUpdatedAt($date)
+                ->save();
+        } catch (\Exception $e) {
+            $this->_adyenLogger->addError(
+                'While processing a notification an exception occured. The payment has already been saved in the
+                 adyen_order_payment table but something went went wrong later. Please check your logs for potential
+                 error messages regarding the merchant reference (order id): "' . $this->_merchantReference . '" and PSP
+                 reference: "' . $this->_pspReference . '"'
```
msilvagarcia: Won't these line breaks make the log terrible to read?

Model/Cron.php
```diff
@@ -1351,20 +1351,28 @@ protected function _prepareInvoice()
...
+                ->setPaymentMethod($this->_paymentMethod)
+                ->setAmount($amount)
+                ->setTotalRefunded(0)
+                ->setCreatedAt($date)
+                ->setUpdatedAt($date)
+                ->save();
+        } catch (\Exception $e) {
+            $this->_adyenLogger->addError(
+                'While processing a notification an exception occured. The payment has already been saved in the
+                 adyen_order_payment table but something went went wrong later. Please check your logs for potential
```
Aleffio:
```suggestion
                 adyen_order_payment table but something went wrong later. Please check your logs for potential
```

Model/Cron.php
cyattilakiss: Concatenated them line by line

Model/Cron.php
cyattilakiss: TY, removed the extra "went"

PR #615, opened 2020-01-09
PW-1998: filter out bcmc from alternative payment methods

Aleffio:
**Description**
BCMC is returned in the paymentMethods call outside of the "scheme" payment method:
"name" : "Bancontact card", "supportsRecurring" : true, "type" : "bcmc"

However, this payment method doesn't work outside of the credit card method.
Filter it out from the alternative paymentmethods call

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
@@ -43,7 +43,7 @@ define(
         var paymentMethod = ko.observable(null);
         var messageComponents;
         var shippingAddressCountryCode = quote.shippingAddress().countryId;
-        var unsupportedPaymentMethods = ['scheme', 'boleto', 'bcmc_mobile_QR', 'wechatpay'];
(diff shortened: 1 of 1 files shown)


arnoudhgz:
Would it not be better to fix this in the result of the API?

cyattilakiss:
@arnoudhgz I will give you the background of this change, I hope it will help to clarify it.
The API sends back all the payment methods available for a merchant account after some filtering based on the additional parameters sent. Although in the checkout page of Magento 2 we want to use the Adyen checkout web components. The Generic card component is able to handle different schemes including the BCMC and this is the preferred solution from our end. This means that instead of rendering the Generic card component and a BCMC redirect payment we just simply hide the BCMC redirect payment method from our checkout.
If you have any further question please let us know we are happy to help.

Best,
Attila
Adyen

stefandoorn:
This change also filters `bcmc_mobile_app` in `String.match(unsupportedPaymentMethod[i])`, therefore also removing 'Bancontact Mobile' payment method from the checkout. 

Changing the `bcmc` to `/^bcmc$/` seems to solve this issue for us, showing `BCMC Mobile` again in the checkout.

PR #619, opened 2020-01-10
Add configuration for Stale bot

rikterbeek:
**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,19 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 14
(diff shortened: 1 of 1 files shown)


cyattilakiss:
Are the labels "Bug report" and "Enhancement" going to be added automatically to the issues when created or we need to mark them after reviewing the issue?

Is it possible to add automatically a label "new" to all the newly created issues and add it as well to the "exemptLabels" list, because it can be that we are inactive and we don't label the new issues.

Is it possible to add a new label "waiting for more information" and even though an issue is labeled as "Bug report" but we have the "waiting for more information" tag as well then the bot considers it to stale? An example use case: New bug report is created, we label it as bug report and ask for more info, the more info never arrives and we have an issue open which is never going to be staled.

PR #625, opened 2020-01-15
Imrankln patch 1

imrankln:
**Description**

Calling of `array $data = []` in the middle of constructor throws exception, "class not found".
If we call the above code at the end of constructor argument, everything works fine.
Following two files need modification:
Block/Redirect/Redirect.php
Block/Info/PaymentLink.php

**Tested scenarios**

**Fixed issue**:


rikterbeek:
Hi @imrankln,

Thanks for the PR! It is indeed best practise to have this at last. We will review test it and merge when there are no issues.

Regards,
Rik
Adyen

Issue #629, opened 2020-01-15
Under some circumstances, it's possible to make a payment but no Order created


viniciusfabri:
**Describe the bug**
If Persistent Cart and Guest Checkout are both enabled, if the customer's session expires with an on-going cart, the user is able to make all the way through to the payment step, submit their credit card info, Adyen processes the notification, but an exception is thrown in Magento that blocks the order from being created (but the customer's card was charged already)

**To Reproduce**
1. Enable Persistent Cart and Guest Checkout, with Adyen Credit Card as payment method
2. Create customer account
3. Login to customer account
4. Add product to cart
5. Go to checkout
6. Edit PHPSESID cookie and set a different value; or delete the cookie. Main idea is to get a different session (as if your current session expired for various reasons), meaning you will no longer be logged in but persistence will keep your current customer data.
7. Refresh page. If redirected to empty cart use minicart to go back to back to checkout. May have to try multiple times, sometimes it does not work from the first try.
8. Fill checkout form data and place order

**Expected behavior**
User is able to place Order

**Actual behavior**
User is shown "Invalid State Change Requested" exception message, Order is not created, but their card was charged (if payment is authorised immediately)

**Screenshots**
Below is a screenshot of our exception.log in Production:
![image](https://user-images.githubusercontent.com/11316647/72433626-05cc7c80-379a-11ea-9f47-8afc1f4c5d48.png)
Each exception is one submission of the payment form, meaning that this particular customer was charged more than 5 times.
This patterns repeats every couple of days on a small/medium sized e-commerce, so I imagine it's not that rare

**Additional context**
I did some digging on this. I initially thought it was a Magento bug, but I could not reproduce it there. Then I noticed it was happening only with Adyen as the payment method. When the scenario above occurs (guest customer trying to finish an order from a quote that belonged to a customer), Magento would "anonymize" the Quote on _vendor/magento/module-persistent/Model/Checkout/GuestPaymentInformationManagementPlugin.php_, which basically removes the customer ID, customer, etc from the Quote object and then saves it.

The Adyen module, however, gets the quote from the Session [here](https://github.com/Adyen/adyen-magento2/blob/develop/Gateway/Validator/InstallmentValidator.php#L80), which then Magento fills the customer information back in, making the check in vendor/magento/module-quote/Model/ChangeQuoteControl.php::isAllowed() fail

[...]

PR #630, opened 2020-01-16
[PW-2015]: Fix orders not being created but card getting charged when Persistent Cart and Guest Checkout enabled

acampos1916:
**Description**
Replacing quote fetch method in Installment Validator so the order flow is interrupted organically when the session has been unset. Instead of relying on session a new repository query has been introduced.

**Tested scenarios**
Placing order with Persistent Cart and Guest Checkout enabled, after manually unsetting session cookie. Credit Card payment method.

**Fixed issue**:  #629

### DIFF
--- a/Gateway/Validator/InstallmentValidator.php
+++ b/Gateway/Validator/InstallmentValidator.php
@@ -40,11 +40,6 @@ class InstallmentValidator extends AbstractValidator
     private $adyenHelper;
 
     /**
-     * @var \Magento\Checkout\Model\Session
-     */
-    private $session;
-
-    /**
      * @var \Magento\Framework\Serialize\SerializerInterface
      */
     private $serializer;
@@ -59,7 +54,6 @@ class InstallmentValidator extends AbstractValidator
      * @param \Magento\Payment\Gateway\Validator\ResultInterfaceFactory $resultFactory
      * @param \Adyen\Payment\Logger\AdyenLogger $adyenLogger
      * @param \Adyen\Payment\Helper\Data $adyenHelper
-     * @param \Magento\Checkout\Model\Session $session
      * @param \Magento\Framework\Serialize\SerializerInterface $serializer
      * @param \Magento\Quote\Model\QuoteRepository $quoteRepository
      */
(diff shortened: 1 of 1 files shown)


Gateway/Validator/InstallmentValidator.php
```diff
@@ -77,7 +78,9 @@ public function validate(array $validationSubject)
         $isValid = true;
         $fails = [];
         $payment = $validationSubject['payment'];
-        $quote = $this->session->getQuote();
+        $quoteId = $payment->getOrderId();
+        //This validator also runs for other payments that don't necesarily have a quoteId
+        $quote = $quoteId?$this->quoteRepository->get($quoteId):false;
```
acampos1916: Not sure if this is the best way to check if a payment contains quoteId. 

In my tests when placing the order this validate() function ran twice, once for sales_order_payment and once for sales_quote_payment event prefixes, one of those do not contain quoteId and a NoSuchEntityException is thrown. This line prevents that. Feedback would be appreciated.

cyattilakiss:
```suggestion
        if ($quoteId) {
            $quote = $this->quoteRepository->get($quoteId);
        } else {
            $quote = false;
        }
```

Gateway/Validator/InstallmentValidator.php
```diff
@@ -77,7 +78,13 @@ public function validate(array $validationSubject)
...
         $fails = [];
         $payment = $validationSubject['payment'];
-        $quote = $this->session->getQuote();
+        $quoteId = $payment->getQuoteId();
+        //This validator also runs for other payments that don't necesarily have a quoteId
+        if ($quoteId) {
+            $quote = $this->quoteRepository->get($quoteId);
+        } else {
+            $quote = false;
+        }
```
rikterbeek: if quote is empty we can ignore it right. Not better to do if(!$quoteId) { return; } ?

acampos1916: Goed idee. The validate function returns `$this->createResult($isValid, $fails)`. Discussing with Attila we think that we could make it return that result if there's no quoteId but seems like unnecessarily duplicating code.

PR #637, opened 2020-01-28
Add scope param when building the redirect url

geryjuhasz:
**Description**
Adding the scope param will make sure we don't end up with weird redirects when there are multiple store views and the store code is being appended to the url. We spotted this bug in action on a website.

**Tested scenarios**
When doing checkout on /se storeview, the redirect path that got generated was on /no storeview, resulting a https://site.com/se/https://site.com/no/adyen/process/redirect/ url, which was leading to a 404 and the purchase was not being completed

**Fixed issue**:  none?!

### DIFF
--- a/Model/Ui/AdyenCcConfigProvider.php
+++ b/Model/Ui/AdyenCcConfigProvider.php
@@ -111,6 +111,7 @@ class AdyenCcConfigProvider implements ConfigProviderInterface
      */
     public function getConfig()
     {
+        $storeId = $this->storeManager->getStore()->getId();
         // set to active
         $config = [
             'payment' => [
@@ -118,7 +119,7 @@ class AdyenCcConfigProvider implements ConfigProviderInterface
                     'vaultCode' => self::CC_VAULT_CODE,
                     'isActive' => true,
                     'redirectUrl' => $this->_urlBuilder->getUrl(
-                        'adyen/process/redirect/', ['_secure' => $this->_getRequest()->isSecure()])
+                        'adyen/process/redirect/', ['_secure' => $this->_getRequest()->isSecure(), '_scope' => $storeId])
                 ]
             ]
         ];
--- a/Model/Ui/AdyenOneclickConfigProvider.php
+++ b/Model/Ui/AdyenOneclickConfigProvider.php
@@ -108,13 +108,14 @@ class AdyenOneclickConfigProvider implements ConfigProviderInterface
      */
     public function getConfig()
     {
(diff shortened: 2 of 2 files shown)


cyattilakiss:
Hi @geryjuhasz ,

Thank you for the PR we are going to look into it soon.

Best,
Attila
Adyen

cyattilakiss:
Hi @geryjuhasz ,

I just tested the latest version without your changes in a multi store setup and all the redirects were working fine. Could you please provide us the steps / configurations to reproduce the issue that you are facing?

We us the Url class to build the redirect urls without the scope parameter not just in the pointed out 2 cases but more. If we need to add the scope then we will need to make sure to add the parameter to all the occurrences.

Thank you in advance.
Best,
Attila
Adyen

geryjuhasz:
> Hi @geryjuhasz ,
> 
> I just tested the latest version without your changes in a multi store setup and all the redirects were working fine. Could you please provide us the steps / configurations to reproduce the issue that you are facing?
> 
> We us the Url class to build the redirect urls without the scope parameter not just in the pointed out 2 cases but more. If we need to add the scope then we will need to make sure to add the parameter to all the occurrences.
> 
> Thank you in advance.
> Best,
> Attila
> Adyen

Did you have the "add store code to url" option enabled in magento admin ?
Our case was with this option activated, we were testing a checkout on sitename.com/se (note the se storecode in the url), and after submitting the checkout form, we ended up on sitename.com/se/https://sitename.com/no/adyen/process/redirect . I var_dumpped the result of getUrl method and it was indeed building a url with the "no" storecode in it, while I was on the "se" storecode at that moment.

cyattilakiss:
Yes we also have that setting enabled.
Are you maybe testing on a PWA integration?

Just to confirm that I get it right:
Instead of 
"https://sitename.com/se/adyen/process/redirect"
you got
"https://sitename.com/se/https://sitename.com/no/adyen/process/redirect"
In this case I assume there is more then just a scope issue.

Please let me know if you have any other config or step in your mind that can help us reproduce the issue.

Best,
Attila
Adyen

geryjuhasz:
> Please let me know if you have any other config or step in your mind that can help us reproduce the issue.

no pwa.
exactly.
Adding the scope as I did in the commit fixed the issue though, it's all working fine with those changes.

cyattilakiss:
Hi @geryjuhasz ,

I tried to verify the changes in our latest release as well, but I didn't succeed :(
I will close this PR for now, but thank you for the contribution!

Best,
Attila
Adyen

PR #638, opened 2020-02-05
Only load 'return_path' config data if it's required.

fahu:
**Description**
It doesn't make sense to retrieve the configuration for failure on each request.
It's enough to only do it on failure.

### DIFF
--- a/Controller/Process/Result.php
+++ b/Controller/Process/Result.php
@@ -99,8 +99,6 @@ class Result extends \Magento\Framework\App\Action\Action
         $response = $this->getRequest()->getParams();
         $this->_adyenLogger->addAdyenResult(print_r($response, true));
 
-        $failReturnPath = $this->_adyenHelper->getAdyenAbstractConfigData('return_path');
-
         if ($response) {
             $result = $this->validateResponse($response);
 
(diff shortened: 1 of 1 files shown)


Controller/Process/Result.php
```diff
@@ -110,10 +108,12 @@ public function execute()
                 $this->_redirect('checkout/onepage/success', ['_query' => ['utm_nooverride' => '1']]);
             } else {
                 $this->_cancel($response);
+                $failReturnPath = $this->_adyenHelper->getAdyenAbstractConfigData('return_path');
                 $this->_redirect($failReturnPath);
             }
         } else {
             // redirect to checkout page
+            $this->_adyenHelper->getAdyenAbstractConfigData('return_path');
```
cyattilakiss:
```suggestion
            $failReturnPath = $this->_adyenHelper->getAdyenAbstractConfigData('return_path');
```

cyattilakiss:
Hi @fahu ,

Thank you for opening the PR, we really appreciate it.
I added a suggestion as well.

Best,
Attila
Adyen

fahu:
Thanks for the notice. I changed it.

Issue #640, opened 2020-02-10
Google Pay not correctly set up in production mode


ebaschiera:
**Describe the bug**
Hi,
when configuring Google Pay to work in production mode, at checkout Google Pay says "Google Pay is not available".
It looks like the environment variable is not correctly initialized.

**To Reproduce**
1. Configure Adyen moduel for production mode
2. Activate and enable Google Pay

**Expected behavior**
At checkout the Google Pay button should appear. Instead the button is not visible and there's an error message: "Google Pay is not available"

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: Chrome 
 - Version 79

**Additional context**
The problem IMHO is at line https://github.com/Adyen/adyen-magento2/blob/develop/view/frontend/web/js/view/payment/method-renderer/adyen-google-pay-method.js#L104 where Google Pay expects "PRODUCTION" but it gets "LIVE".

Thanks!

cyattilakiss:
Hi @ebaschiera,

Thank you for raising this issue. We are going to test it against your steps in our instances and in case we can reproduce the issue we are going to make a bug fix asap.
Also if you could provide a pull request with the fix we would be grateful review and release that.

Best,
Attila
Adyen

cyattilakiss:
Hi @ebaschiera ,

Indeed, I verified that we do propagate the wrong value.

Best,
Attila

Aleffio:
Hi @ebaschiera,

This issue has been fixed in the latest(5.4.0) release,
closing the issue,

Kind Regards,
Alessio
Adyen

PR #641, opened 2020-02-10
[PW-2091] Fix Google Pay environment value in PRODUCTION

cyattilakiss:
**Description**

Google pay expects `PRODUCTION` and not `LIVE` in case of production
environment
Add getGooglePayEnvironment() to retrieve environment value to Google
pay
Replace the getCheckoutEnvironment() with getGooglePayEnvironment() on
the frontend for Google pay

**Fixed issue**:  
#640


cyattilakiss:
@rikterbeek changed the PR as discussed in private, minimising the risk of removing this in the future when the component releases a fix for this.

PR #643, opened 2020-02-11
[PW 2022] Fix payment-information interceptor to work with other plugins

cyattilakiss:
**Description**

Because we have to handle 3DS2 communication (identifyShopper and challengeShopper) we changed the afterSavePaymentInformationAndPlaceOrder() return value from int to a JSON. This is unfortunately a confirmed bad practise because with this approach we break all the other plugins which tries to intercept into the process since they expect an int as a parameter and not a JSON. Also the frontend can break because of the same reason.

**Fixed issue**:  
#590


Api/AdyenOrderPaymentStatusInterface.php
```diff
@@ -0,0 +1,34 @@
...
+ * Author: Adyen <magento@adyen.com>
+ */
+
+namespace Adyen\Payment\Api;
+
+interface AdyenOrderPaymentStatusInterface
+{
+    /**
+     * @param string $orderId
+     * @return mixed
```
msilvagarcia: Maybe change to string

view/frontend/web/js/model/adyen-payment-service.js
```diff
@@ -65,6 +65,26 @@ define(
                         self.setPaymentMethods([]);
                     }
                 )
+            },
+            getOrderPaymentStatus: function (orderId, callback) {
+                var self = this;
```
msilvagarcia: This can be removed

view/frontend/web/js/model/adyen-payment-service.js
```diff
@@ -65,6 +65,26 @@ define(
...
+                var serviceUrl = urlBuilder.createUrl('/adyen/orders/:orderId/payment-status', {
+                    orderId: orderId
+                });
+
+                storage.get(
+                    serviceUrl
+                ).done(
```
msilvagarcia: This can go in a single line

view/frontend/web/js/model/adyen-payment-service.js
```diff
@@ -65,6 +65,26 @@ define(
...
+                    function (response) {
+                        if (callback) {
+                            callback(response);
+                        }
+                    }
```
msilvagarcia: This can be replaced with `callback`

view/frontend/web/js/model/adyen-payment-service.js
```diff
@@ -65,6 +65,26 @@ define(
...
+                ).fail(
+                    function (response) {
+                       console.log(response);
+                    }
```
msilvagarcia: This means that the shopper will never know if an error occurs

rikterbeek: We should not do this indeed and log the error on server level.

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -338,7 +339,7 @@ define(
                         ).done(
                         function (response) {
                             self.afterPlaceOrder();
-                            self.validateThreeDS2OrPlaceOrder(response);
+                            adyenPaymentService.getOrderPaymentStatus(response, self.validateThreeDS2OrPlaceOrder.bind(self));
```
rikterbeek: not sure why we need to do self.validateThreeDS2OrPlaceOrder.bind(self). I would suggest to implement a promis.

PR #644, opened 2020-02-11
PW-2100 filter out applepay, paywithgoogle from alternatives

Aleffio:
**Description**
filter out applepay, paywithgoogle from alternatives


Aleffio:
> Is this checking if the payment method is already enabled in Magento 2 admin?

It is not needed, since those payment methods will not work here(adyen-hpp) anyway.

PR #645, opened 2020-02-12
[PW 1949]: Notifications error_count and error_message

acampos1916:
**Description**
Added an error count and error message columns to adyen_notification table and Notifications Overview. UpgradeSchema has been modified for this.

**Tested scenarios**
Placed and order, received notification and it now shows in Notifications Overview. Thrown an exception during the notification processing and the exception is appended to error_message, error_count increased by 1.

**Fixed issue**:  PW-1949


etc/module.xml
```diff
@@ -24,7 +24,7 @@
 -->
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">
 
-    <module name="Adyen_Payment" setup_version="4.5.4">
+    <module name="Adyen_Payment" setup_version="5.3.1">
```
acampos1916: Is this the correct usage of setup_version according to our release convention? https://hub.is.adyen.com/our-company/organization/workstreams/mid-market-solution/plugins-and-partners/internal-repo/release-steps/magento2/#release-on-github

Model/Notification.php
```diff
@@ -49,6 +49,7 @@ class Notification extends \Magento\Framework\Model\AbstractModel implements Not
     const REPORT_AVAILABLE = "REPORT_AVAILABLE";
     const ORDER_CLOSED = "ORDER_CLOSED";
     const OFFER_CLOSED = "OFFER_CLOSED";
+    const MAX_ERROR_COUNT = 5;
```
acampos1916: Hard limit for now

Model/Notification.php
```diff
@@ -359,6 +360,69 @@ public function setDone($done)
...
+     * Sets Processing flag.
+     *
+     * @param bool $processing
+     * @return $this
+     */
+    public function setProcessing($processing)
+    {
+        return $this->setData(self::PROCESSING, $processing);
+    }
+
```
acampos1916: These getter and setter for processing are not needed for this PR but I noticed that they were missing

Model/Notification.php
```diff
@@ -359,6 +360,69 @@ public function setDone($done)
...
+    /**
+     * Sets Processing flag.
+     *
+     * @param bool $processing
+     * @return $this
+     */
+    public function setProcessing($processing)
+    {
+        return $this->setData(self::PROCESSING, $processing);
+    }
```
acampos1916: These getter and setter for processing are not needed for this PR but I noticed that they were missing

etc/module.xml
msilvagarcia: Yes, I believe so.

cyattilakiss: it's gonna be a feature release so I would go for ->

```suggestion
    <module name="Adyen_Payment" setup_version="5.4.0">
```

PR #639, opened 2020-02-07
[PW-1942] Notifications Overview

acampos1916:
**Description**
Adds a notifications overview page (System > Adyen > Notifications Overview) that displays the notifications stored in DB for easy reference.

**Tested scenarios**
List live and test notifications stored in DB with matching and unmatching merchant_reference. In case that the merchant_reference is not found in Magento then no link to order is provided.

**Fixed issue**:  PW-1942

**WIP**: Add new exception column for failed cron process.

PR #646, opened 2020-02-17
Release 5.4.0

cyattilakiss:
**Description**

 - Fix marketplace validation error
 - Add Fashioncheque logo
 - #638 Only load 'return_path' config data if it's required.
 - #641 [PW-2091] Fix Google Pay environment value in PRODUCTION
 - #644 [PW-2100] filter out applepay, paywithgoogle from alternatives
 - #639 [PW-1942] Notifications Overview 
 - #645 [PW 1949]: Notifications error_count and error_message 

**Fixed issue**:


composer.json
```diff
@@ -17,7 +17,7 @@
     "adyen/php-api-library": "~4.1",
```
Aleffio: please update "version" to "5.4.0"

msilvagarcia: Opened #647 for this

PR #650, opened 2020-02-24
[PW-2022] changes to get it working

rikterbeek:
**Description**

Changes to get PW-2022 working
**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Model/AdyenOrderPaymentStatus.php
+++ b/Model/AdyenOrderPaymentStatus.php
@@ -24,6 +24,7 @@
 
 namespace Adyen\Payment\Model;
 
+use Magento\Framework\Exception\LocalizedException;
 use Magento\Framework\Exception\NoSuchEntityException;
 use Adyen\Payment\Model\Ui\AdyenCcConfigProvider;
 
@@ -66,6 +67,7 @@ class AdyenOrderPaymentStatus implements \Adyen\Payment\Api\AdyenOrderPaymentSta
      * {@inheritDoc}
      * @throws MagentoNoSuchEntityException
      * @throws AdyenException
+     * @throws LocalizedException
      */
     public function getOrderPaymentStatus($orderId)
     {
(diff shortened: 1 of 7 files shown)


Model/AdyenOrderPaymentStatus.php
```diff
@@ -55,37 +55,30 @@ public function __construct(
...
+        ) {
+            $additionalInformation = $payment->getAdditionalInformation();
+            return $this->adyenHelper->buildThreeDS2ProcessResponseJson(
+                $additionalInformation['threeDSType'],
+                $additionalInformation['threeDS2Token']
+            );
         }
-
-        return $result;
+        return true;
```
cyattilakiss: Do we handle the boolean true return?

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -339,7 +357,10 @@ define(
                         ).done(
                         function (response) {
                             self.afterPlaceOrder();
-                            adyenPaymentService.getOrderPaymentStatus(response, self.validateThreeDS2OrPlaceOrder.bind(self));
+                            adyenPaymentService.getOrderPaymentStatus(response)
+                                .done(function (responseJSON) {
```
cyattilakiss: this way is nicer!

view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
```diff
@@ -184,6 +207,7 @@ define(
                          * @returns {boolean}
                          */
                         placeOrder: function (data, event) {
+                            debugger;
```
cyattilakiss: Is it WIP?

rikterbeek: it was but it is working so now removed it.

Model/AdyenOrderPaymentStatus.php
rikterbeek: Yes it will continue to the success page.

Model/AdyenOrderPaymentStatus.php
```diff
@@ -55,37 +55,30 @@ public function __construct(
...
-            throw new \Magento\Framework\Exception\LocalizedException(__('This order no longer exists.'));
+        if ($payment->getMethod() === AdyenCcConfigProvider::CODE ||
+            $payment->getMethod() === AdyenOneclickConfigProvider::CODE
+        ) {
+            $additionalInformation = $payment->getAdditionalInformation();
+            return $this->adyenHelper->buildThreeDS2ProcessResponseJson(
+                $additionalInformation['threeDSType'],
+                $additionalInformation['threeDS2Token']
+            );
         }
```
acampos1916: $this->orderRepository->get($orderId); can still throw an exception if the orderId supplied does not exist. Shouldn't we keep the try/catch block to handle it?

rikterbeek: @acampos1916 I can indeed but the front-end is not able to catch this exception as this time and as well quite a lot of work to implement as it should never happen. I believe this should give a server error so the merchant during development will notice that they probably build a plugin that changed the response causing this to fail. I am open for any other suggestion but believe most of them will need quite a lot of development for it to make sure the front-end will understand this error.

Issue #494, opened 2019-08-08
Wrong store id used for config settings, wrong api url is getting loaded. Related credit memo sync


govereem:
**Magento version**: 2.2.8
**Plugin version**: 4.2.1
**Description**

When creating a credit memo, Adyen is not receiving them from our Magento store.
We have 2 merchant accounts in our webshop, one is working fine. The other one fails. Payment methods are working fine, for both stores.

AdyenException: status 403, type:security Invalid Merchant Account code 901.

`AdyenLoggerTest.INFO: Request url to Adyen: shows te wrong url. The url showing is for our main store. Not the store where the order / credit memo was created. Zo this means the storeid is 1, the default and not using the config settings from the right store.

```
vendor/magento/module-sales/Model/Order/Payment.php Line: 675 the store id is correct.

But in vendor/adyen/module-payment/Gateway/Http/Client/TransactionRefund.php 
$this->_client = $this->_adyenHelper->initializeAdyenClient();

There is no storeId set. Method placeRequest will always use storeId 1
```

cyattilakiss:
Hi @govereem ,

I created a branch where I uploaded a fix that I think should solve this issue. Can you please verify if this helps? 
https://github.com/Adyen/adyen-magento2/commit/70af7744889be38ee17c35695177abaed9908f3c

Thank you in advance,
Attila
Adyen

govereem:
Hi Attila,

Thank you for trying to solve this problem.

Unfortunately this solution does not solve the problem, because
$storeManager->getStore()->getId(); wil return the default store, not the
store from the order.

With kind regards,
Gideon

Op do 8 aug. 2019 om 16:20 schreef cyattilakiss <notifications@github.com>:

cyattilakiss:
Hi @govereem ,

I will dive a bit deeper then into this issue. I thought that we simply miss the storeId but indeed it needs to be retrieved with another approach.
If you can contribute and create a PR if you have a solution that would be appreciated :)

Best,
Attila

govereem:
For now i fixed it by retrieving the order increment from reference, but
for workaround its fine, but not the best solution.

vendor/adyen/module-payment/Gateway/Http/Client/TransactionRefund.php


        \Magento\Sales\Model\OrderRepository $orderRepository,
        \Magento\Framework\Api\SearchCriteriaBuilder $searchCriteriaBuilder,

    public function
placeRequest(\Magento\Payment\Gateway\Http\TransferInterface
$transferObject)
[...]

PR #653, opened 2020-02-25
[PW-2073]Pass the correct store id from order for the modification requests re…

AlexandrosMor:
Description
Wrong store id is used for multi store merchants. When creating a credit memo, capture or cancel an order and there are multiple stores, the storeid is added to the request.

Tested scenarios

From a second store with different merchant account making the following actions
Capture
Cancel
Refund

Fixed issue: #494

### DIFF
--- a/Gateway/Request/CaptureDataBuilder.php
+++ b/Gateway/Request/CaptureDataBuilder.php
@@ -70,7 +70,8 @@ class CaptureDataBuilder implements BuilderInterface
         $requestBody = [
             "modificationAmount" => $modificationAmount,
             "reference" => $payment->getOrder()->getIncrementId(),
-            "originalReference" => $pspReference
+            "originalReference" => $pspReference,
+            "storeId" => $payment->getOrder()->getStoreId()
         ];
 
         $brandCode = $payment->getAdditionalInformation(
--- a/Gateway/Request/CaptureDataBuilder.php
+++ b/Gateway/Request/CaptureDataBuilder.php
@@ -40,10 +40,14 @@ class CaptureDataBuilder implements BuilderInterface
      * CaptureDataBuilder constructor.
      *
      * @param \Adyen\Payment\Helper\Data $adyenHelper
+     * @param \Adyen\Payment\Logger\AdyenLogger $adyenLogger
+
      */
-    public function __construct(\Adyen\Payment\Helper\Data $adyenHelper)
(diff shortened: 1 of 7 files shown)


Gateway/Request/RefundDataBuilder.php
```diff
@@ -158,7 +159,8 @@ public function build(array $buildSubject)
                     "modificationAmount" => $modificationAmount,
                     "reference" => $payment->getOrder()->getIncrementId(),
                     "originalReference" => $pspReference,
-                    "merchantAccount" => $merchantAccount
+                    "merchantAccount" => $merchantAccount,
+                    "storeId" => $payment->getOrder()->getStoreId()
```
rikterbeek: I know this is an easy way to do it but it is not very nice. Is there a way we can extend the transferObject and add here a new attribute storeId so we don't have to do this into the payload and unset it. @cyattilakiss remembered you did something like that before ?

AlexandrosMor: Can you check again ? @rikterbeek

Gateway/Request/CaptureDataBuilder.php
```diff
@@ -40,10 +40,14 @@ class CaptureDataBuilder implements BuilderInterface
      * CaptureDataBuilder constructor.
      *
      * @param \Adyen\Payment\Helper\Data $adyenHelper
+     * @param \Adyen\Payment\Logger\AdyenLogger $adyenLogger
+
      */
-    public function __construct(\Adyen\Payment\Helper\Data $adyenHelper)
+    public function __construct(\Adyen\Payment\Helper\Data $adyenHelper,\Adyen\Payment\Logger\AdyenLogger $adyenLogger)
```
cyattilakiss: logger can be removed ;)

Gateway/Request/CaptureDataBuilder.php
```diff
@@ -81,13 +85,11 @@ public function build(array $buildSubject)
             $openInvoiceFields = $this->getOpenInvoiceData($payment);
             $requestBody["additionalData"] = $openInvoiceFields;
         }
-
         $request['body'] = $requestBody;
-
+        $request['clientConfig']=[ "storeId" => $payment->getOrder()->getStoreId()];
```
cyattilakiss: please format the code

PR #656, opened 2020-02-28
[PW-2033]: Update Magento2 to latest library version dependency (6.0.0)

acampos1916:
**Description**
This PR upgrades the PHP API Lib version from ~4.1 to ~6.0. Boleto and Multibanco now respond back with "PresentToShopper" resultCode instead of "Received" and the information to be processed is in the "action" index.

When the Checkout Component is upgraded from 3.2.0 we'll be able to pass this "action" object.

By the time of doing this PR Multibanco requests to Adyen test have been failing for 24h and thorough testing was not possible, this has to be tested afterwards but the "action" object is still expected.

**Tested scenarios**
Boleto payment
CC payment
CC 3DS2 payment

**Fixed issue**:  PW-2033


Gateway/Validator/CheckoutResponseValidator.php
```diff
@@ -89,6 +89,7 @@ public function validate(array $validationSubject)
                             }
                         }
                     } elseif (!empty($response['additionalData']['comprafacil.entity'])) {
+                        //Multibanco resultCode has changed after checkout v49 and comprafacil.entity is not received anymore
```
rikterbeek: it is now called entity without the prefix ? I see below that $multibancoData['entity']  is used.

Block/Checkout/Success.php
```diff
@@ -96,10 +104,10 @@ public function isBoletoPayment()
     /**
      * @return null|\string[]
      */
-    public function getBoletoPdfUrl()
+    public function getBoletoData()
     {
         if ($this->isBoletoPayment()) {
-            return $this->getOrder()->getPayment()->getAdditionalInformation('url');
+            return $this->getOrder()->getPayment()->getAdditionalInformation();
```
cyattilakiss: I think we should specify the additional data we want to get instead of everything
```suggestion
            return $this->getOrder()->getPayment()->getAdditionalInformation('action');
```

Block/Checkout/Success.php
```diff
@@ -130,7 +138,7 @@ public function getMultibancoData()
     {
         $result = [];
         if (!empty($this->getOrder()->getPayment()) &&
-            !empty($this->getOrder()->getPayment()->getAdditionalInformation('comprafacil.entity'))
+            strcmp($this->getOrder()->getPayment()->getAdditionalInformation('paymentMethodType'), 'multibanco') === 0
```
cyattilakiss: Check if the `paymentMethodType` is not empty
```suggestion
            !empty($this->getOrder()->getPayment()->getAdditionalInformation('paymentMethodType')) &&
            strcmp($this->getOrder()->getPayment()->getAdditionalInformation('paymentMethodType'), 'multibanco') === 0
```

Gateway/Validator/CheckoutResponseValidator.php
```diff
@@ -105,32 +106,7 @@ public function validate(array $validationSubject)
...
-                        }
-
-                        if (isset($outputDetails['boletobancario.url'])) {
-                            $payment->setAdditionalInformation(
-                                'url',
-                                $outputDetails['boletobancario.url']
-                            );
-                        }
-                    }
+                    $payment->setAdditionalInformation('action', $response['action']);
```
cyattilakiss: Check if $response['action'] is not empty before assigning in the additionalInformation
Also a question for the pspReference, why was it removed from the additionalInformation? Don't we use it anymore?

```suggestion
                    $payment->setAdditionalInformation('pspReference', $response['pspReference']);
                    if (!empty($response['action']) {
                        $payment->setAdditionalInformation('action', $response['action']);
                    }
```

composer.json
```diff
@@ -14,7 +14,7 @@
     }
   ],
   "require": {
-    "adyen/php-api-library": "~4.1",
+    "adyen/php-api-library": "~6.0",
```
cyattilakiss: Also enough to specify it like this
```suggestion
    "adyen/php-api-library": "~6",
```

Block/Checkout/Success.php
```diff
@@ -44,22 +44,30 @@ class Success extends \Magento\Framework\View\Element\Template
      */
     protected $_orderFactory;
 
+    /**
+     * @var \Magento\Framework\Pricing\Helper\Data
+     */
+    public $priceHelper;
```
cyattilakiss: Where do you use this public priceHelper?

Gateway/Validator/CheckoutResponseValidator.php
```diff
@@ -105,32 +106,7 @@ public function validate(array $validationSubject)
                     $payment->setAdditionalInformation('pspReference', $response['pspReference']);
                     break;
                 case "PresentToShopper":
-                    $payment->setAdditionalInformation('pspReference', $response['pspReference']);
```
acampos1916: add pspReference if and assign data

Gateway/Validator/CheckoutResponseValidator.php
acampos1916: Yes, the index is now 'entity' with no prefix https://docs.adyen.com/payment-methods/multibanco/api-only#-payments-response

Block/Checkout/Success.php
acampos1916: In the template file, in order to format the value that we get in the response using the Magento configured locale https://github.com/Adyen/adyen-magento2/pull/656/files#diff-b47a2330059e98c2cdb5ce9b576ab116R120

Gateway/Validator/CheckoutResponseValidator.php
acampos1916: Added !empty check on action and pspReference and assigned both variables

PR #657, opened 2020-02-28
[PW-2150] Fix 3DS2 challenge popup styling

cyattilakiss:


view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -245,6 +245,8 @@ define(
                     var popupModal = $('#threeDS2Modal').modal({
                         // disable user to hide popup
                         clickableOverlay: false,
+                        responsive: true,
+                        innerScroll: false,
```
Aleffio: do we need the same in adyen-oneclick-method.js?

PR #663, opened 2020-03-06
[PW-2005] Fix klarna discount line and logo

AlexandrosMor:
**Description**
Refactor discount lines according to [Klarna standards: ](https://docs.adyen.com/payment-methods/klarna/invoice-lines#discount-line-incl-vat)

**Tested scenarios**

Klarna transaction is successful.
Check that there are no correction lines in logsearch

Merchant Requests:
**Fixed issue**:


Gateway/Request/CheckoutDataBuilder.php
```diff
@@ -241,14 +241,14 @@ protected function getOpenInvoiceData($order): array
         // Discount cost
         if ($discountAmount != 0) {
 
-            $description = __('Total Discount');
-            $itemAmount = $this->adyenHelper->formatAmount($discountAmount, $currency);
+            $description = __('Discount');
+            $itemAmount = '-'.$this->adyenHelper->formatAmount($discountAmount, $currency);
```
rikterbeek: as this will be of type int shalle we do $number = -$this->adyenHelper->formatAmount($discountAmount, $currency); instead ?

cyattilakiss: I think it's not really readable this way.
Maybe like this?
```suggestion
            $itemAmount = $this->adyenHelper->formatAmount(-1 * $discountAmount, $currency);
```

AlexandrosMor: Yes it is working with the minus in front of the int

AlexandrosMor: I used the first approach

Issue #664, opened 2020-03-09
[BUG] Payment Methods do not match cart total amount


GGMCarl:
**Describe the bug**
Endpoint: rest/V1/guest-carts/XXX/retrieve-adyen-payment-methods

If payment amount is greater than XXX Eur for example payment methods do not change. Invoice like Klarna should not be possible to select and therefore be hidden in the response. Value should be detected by submitted cart id of magento. Currently the Module is broken at this point.. .also Adyen Rest-API Does always return same payment methods...

GGMCarl:
Maybe releated to hist https://github.com/Adyen/adyen-magento2/issues/546

cyattilakiss:
Hi @GGMCarl ,

Thank you for raising this issue, we are going to look into it and create the necessary follow up steps.

Best,
Attila
Adyen

cyattilakiss:
Hi @GGMCarl ,

The fix has just been released in the version 6.0.0
Please note that this is a major release including breaking changes for backwards compatibility. (You can read more about the changes in the release notes)

I will close this issue for now but please feel free to reopen it if you still experience the issue after the upgrade.

Best,
Attila
Adyen

Issue #546, opened 2019-10-15
Incorrect grand total amount when using REST API


sergy-gzh:
**Magento version**: 2.3.3
**Plugin version**: 4.5.4
**Description**
When calling `retrieve-adyen-payment-methods` API for both `customer` and `guest` Adyen magento module is using session to fetch quote info.
As a result incorrect quote info (empty) is achieved and default value of (10) is communicated.
**Expected**
Rest API will load quote data based on the provided `cart_id` not session data.
https://github.com/Adyen/adyen-magento2/blob/develop/Helper/PaymentMethods.php#L383

cyattilakiss:
Hi @sergy-gzh ,

Thank you for raising the issue, we are aware of the problem and going to work on it as soon as possible.

Best,
Attila
Adyen

sergy-gzh:
Wow, easy stale bot! Issue is relevant.

cyattilakiss:
@sergy-gzh sorry for the bot :) We just installed it and haven't set all the necessary tags yet but we are on it.

cyattilakiss:
Hi @sergy-gzh , 

Thank you for the patience. The fix has just been released in the version 6.0.0 but you can also give a look to the pull request with the fix #667

Please note that for PWA integrations version 6.0.0 includes breaking changes and the updated documentation for PWA integration is coming soon on our docs page.

I will close this issue now, but please feel free to reopen it if you still experience the issue after the upgrade.

Best,
Attila
Adyen

PR #667, opened 2020-03-17
[PW-2183] Use quote by id instead of sessions at retrieve-adyen-payment-methods

cyattilakiss:
**Description**

Use the quote retrieved by the quote id and not by the session.

**Tested scenarios**
Retrieve payment methods by using the API without any sessions
Value is changing along with the quote value

**Fixed issue**:  
#546 
#664

### DIFF
--- a/Helper/PaymentMethods.php
+++ b/Helper/PaymentMethods.php
@@ -86,6 +86,11 @@ class PaymentMethods extends AbstractHelper
     protected $themeProvider;
 
     /**
+     * @var \Magento\Quote\Model\Quote
+     */
+    protected $quote;
+
+    /**
      * PaymentMethods constructor.
      *
      * @param \Magento\Quote\Api\CartRepositoryInterface $quoteRepository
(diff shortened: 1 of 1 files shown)


Helper/PaymentMethods.php
```diff
@@ -133,41 +138,51 @@ public function getPaymentMethods($quoteId, $country = null)
...
         $quote = $this->quoteRepository->getActive($quoteId);
-        $store = $quote->getStore();
 
-        $paymentMethods = $this->fetchAlternativeMethods($store, $country);
+        // If quote cannot be found early return the empty paymentMethods array
+        if (empty($quote)) {
+            return [];
+        }
+
+        $this->setQuote($quote);
```
rikterbeek: why adding a setQuote() method as we can easily do $this->quote = $quote ?

Helper/PaymentMethods.php
```diff
@@ -133,41 +138,51 @@ public function getPaymentMethods($quoteId, $country = null)
...
     /**
-     * @param $store
      * @param $country
      * @return array
      */
-    protected function fetchAlternativeMethods($store, $country)
+    protected function fetchAlternativeMethods($country)
     {
-        $merchantAccount = $this->adyenHelper->getAdyenAbstractConfigData('merchant_account');
+        $quote = $this->getQuote();
```
rikterbeek: why not doing a $this->quote instead ?

Helper/PaymentMethods.php
cyattilakiss: There was already a getter created and by generally talking getters and setters are good to use because you can have better control over your property. In the current case we don't get these benefits because the $quote, the getQuote() and setQuote() are all protected but can be useful for later.

Helper/PaymentMethods.php
cyattilakiss: The getQuote() getter was already in the code and used several times. Instead of replacing it with the $this->quote I decided to keep using it and add the setter as well.

PR #668, opened 2020-03-17
PW-2069 Change default values for Order Status

msilvagarcia:
As an improvement for merchants installing the plugin, we're changing the default values for the **Magento order management** section.

### DIFF
--- a/etc/config.xml
+++ b/etc/config.xml
@@ -28,7 +28,6 @@
             <adyen_abstract>
                 <active>0</active>
                 <model>AdyenPaymentGenericFacade</model>
-                <order_status>pending</order_status>
                 <demo_mode>1</demo_mode>
                 <debug>1</debug>
                 <title_renderer>title_image</title_renderer>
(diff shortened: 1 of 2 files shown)


etc/config.xml
```diff
@@ -38,6 +37,16 @@
...
                 <enable_recurring>0</enable_recurring>
                 <group>adyen</group>
+                <!-- etc/adminhtml/system/order_management/order_management.xml -->
+                <order_status>pending</order_status>
+                <payment_pre_authorized>pending</payment_pre_authorized>
+                <payment_authorized>processing</payment_authorized>
+                <payment_authorized_virtual>complete</payment_authorized_virtual>
+                <payment_cancelled>canceled</payment_cancelled>
+                <fraud_manual_review_status>pending</fraud_manual_review_status>
+                <fraud_manual_review_accept_status>processing</fraud_manual_review_accept_status>
```
acampos1916: @msilvagarcia I see that in the prototype this field is set to Pending and not to Processing, but Pending is not an option on the source model we're using for it (Magento\Sales\Model\Config\Source\Order\Status\Processing). Seems like a mistake on the prototype. Maybe it is a good idea to state this in the ticket so the discrepancy is documented?

acampos1916: Done!

PR #669, opened 2020-03-18
[PW-2128]: Migration script for setting terminal_selection based on pos_store_id

acampos1916:
**Description**
A new config dropdown has been introduced to hide/show the pos_store_id text field. This terminal_selection config will now be used to check if the terminals to be queried are merchant account level or store level.

The new config usage has already been merged to the feature branch here https://github.com/Adyen/adyen-magento2/blob/c80a0332dce9a4a2a81e3aad204a8c50d25492e6/Helper/PaymentMethods.php#L376

**Tested scenarios**
* setup:upgrade for scopes with no pos_store_id set: No changes are made

* setup:upgrade for scopes with pos_store_id set: config is created for terminal_selection=store_level

* setup:upgrade for scopes with pos_store_id empty: config is created for terminal_selection=merchant_account_level

**Fixed issue**:  PW-2128

### DIFF
--- a/Setup/UpgradeData.php
+++ b/Setup/UpgradeData.php
@@ -66,6 +66,10 @@ class UpgradeData implements UpgradeDataInterface
             $this->updateSchemaVersion244($setup);
         }
 
+        if (version_compare($context->getVersion(), '6.0.0', '<')) {
+            $this->updateDataVersion600($setup);
+        }
+
         $setup->endSetup();
     }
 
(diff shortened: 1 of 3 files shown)


Setup/UpgradeData.php
```diff
@@ -66,6 +66,10 @@ public function upgrade(ModuleDataSetupInterface $setup, ModuleContextInterface
             $this->updateSchemaVersion244($setup);
         }
 
+        if (version_compare($context->getVersion(), '6.0.0', '<')) {
```
rikterbeek: How did you test this as you did not upgrade the setup version so this will never be triggered. What is your take on this are we adding all the scripts under 1 version upgrade this could break some development environments switching between branches and versions.

Setup/UpgradeData.php
```diff
@@ -158,4 +162,58 @@ public function updateSchemaVersion244(ModuleDataSetupInterface $setup)
...
+            ->where('value <> "" AND value IS NOT NULL');
+        $configPosStoreIdValues = $connection->fetchAll($select);
+        foreach ($configPosStoreIdValues as $configPosStoreIdValue) {
+            $scope = $configPosStoreIdValue['scope'];
+            $scopeId = $configPosStoreIdValue['scope_id'];
+            $this->configWriter->save(
+                $pathTerminalSelection,
+                'store_level',
+                $scope,
+                $scopeId
```
rikterbeek: just wondering if we take into account that a setting can be in multiple stores. Do we include storeId into the read an save operation ?

Setup/UpgradeData.php
acampos1916: I manually changed the setup version in my environment in order to test it, but did not commit it as we can do that in a release merge. Do you think we should change the setup version in this feature branch?

I do believe that all migration scripts related to this config UI feature should be included under the same version upgrade because some configs are interdependent. This one in particular will not break environments when changing to older versions or branches because it only adds a new config row when needed, that row won't be used on envs after a downgrade. I'll keep an eye to keep new changes backwards compatible where possible.

Setup/UpgradeData.php
acampos1916: Correct, the $scope and $scopeId variables localize the config value to website, store or storeview. If there's a pos_store_id setting for a specific storeview we'll apply the terminal_selection in that specific scope as well.

Setup/UpgradeData.php
acampos1916: Now using alpha versioning for this branch https://github.com/Adyen/adyen-magento2/pull/669/commits/5678b920e1ca79f6f5df2637e329123ac2cc452d

PR #671, opened 2020-03-20
PW-2068 Change shopper locale config input type

msilvagarcia:
Improves the UI of the admin menu.


acampos1916:
@msilvagarcia Are we merging this directly into develop and not the feature branch? I don't see a problem but just checking

PR #674, opened 2020-03-25
[PW-2124] Replace ISO Country Code config field with Shopper Country dependent dropdown

AlexandrosMor:
**Description**
The previous text config field "ISO Country Code" is being replaced with a country select field that is dependent on a Yes/No type of dropdown.

Migration script must be created to upgrade from the country code as text to the dropdown value. Then the backend logic must be modified where needed to use this new value.

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1901,4 +1901,21 @@ class Data extends AbstractHelper
         );
 
     }
+
+    /**
+     * Gives back CountryCode configuration values
+     *
+     * @param null $storeId
+     * @return mixed
+     */
+    public function getAdyenCountryCodeConfigData($storeId = null)
+    {
+        $shopperLocale = trim($this->_adyenHelper->getAdyenHppConfigData('shopper_locale', $storeId));
+        $countryCode = trim($this->_adyenHelper->getAdyenHppConfigData('country_code', $storeId));
+        if ($countryCode == false && $shopperLocale == 1) {
+            return $this->getConfigData($countryCode, 'adyen_hpp', $storeId);
+        } else {
+            return "";
+        }
+    }
 }
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1910,10 +1910,10 @@ class Data extends AbstractHelper
      */
     public function getAdyenCountryCodeConfigData($storeId = null)
     {
-        $shopperLocale = trim($this->_adyenHelper->getAdyenHppConfigData('shopper_locale', $storeId));
-        $countryCode = trim($this->_adyenHelper->getAdyenHppConfigData('country_code', $storeId));
-        if ($countryCode == false && $shopperLocale == 1) {
-            return $this->getConfigData($countryCode, 'adyen_hpp', $storeId);
+        $shopperCountry = trim($this->getConfigData('shopper_country', 'adyen_hpp', $storeId));
+        $countryCode = trim($this->getConfigData('country_code', 'adyen_hpp', $storeId));
+        if ($shopperCountry == 1) {
+            return $countryCode;
         } else {
             return "";
         }
--- a/Model/Config/Source/ShopperCountryMode.php
+++ b/Model/Config/Source/ShopperCountryMode.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ *                       ######
+ *                       ######
+ * ############    ####( ######  #####. ######  ############   ############
+ * #############  #####( ######  #####. ######  #############  #############
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen Payment module (https://www.adyen.com/)
+ *
+ * Copyright (c) 2020 Adyen BV (https://www.adyen.com/)
+ * See LICENSE.txt for license details.
+ *
+ * Author: Adyen <magento@adyen.com>
+ */
+
+namespace Adyen\Payment\Model\Config\Source;
+
+class ShopperCountryMode implements \Magento\Framework\Option\ArrayInterface
+{
+    const OPTIONS = [
+        [
+            'value' => 0,
+            'label' => 'Based on address'
+        ],
+        [
+            'value' => 1,
+            'label' => 'Costum'
+        ],
+    ];
+
+    /**
+     * @return array
+     */
+    public function toOptionArray()
+    {
+        return self::OPTIONS;
+    }
+}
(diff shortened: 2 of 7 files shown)


Model/Config/Source/ShopperCountryMode.php
```diff
@@ -0,0 +1,46 @@
...
+class ShopperCountryMode implements \Magento\Framework\Option\ArrayInterface
+{
+    const OPTIONS = [
+        [
+            'value' => 0,
+            'label' => 'Based on address'
+        ],
+        [
+            'value' => 1,
+            'label' => 'Costum'
```
acampos1916:
```suggestion
            'label' => 'Custom'
```

etc/adminhtml/system/payment_methods/local_payment_methods/advanced_options.xml
```diff
@@ -26,32 +26,42 @@
...
     </field>
-    <field id="ratepay_id" translate="label" type="text" sortOrder="20" showInDefault="1" showInWebsite="1" showInStore="1">
+    <field id="shopper_specific_country" translate="label" type="select" sortOrder="20" showInDefault="1" showInWebsite="1" showInStore="0">
+        <label>Custum shopper country</label>
+        <depends>
+            <field id="shopper_country">1</field>
+        </depends>
+        <source_model>Magento\Directory\Model\Config\Source\Country</source_model>
+        <can_be_empty>1</can_be_empty>
+        <config_path>payment/adyen_hpp/countrycode</config_path>
```
acampos1916:
```suggestion
        <config_path>payment/adyen_hpp/country_code</config_path>
```

acampos1916: This is the correct config path so it works with previous installations

etc/adminhtml/system/payment_methods/local_payment_methods/advanced_options.xml
```diff
@@ -26,32 +26,42 @@
...
     <field id="shopper_country" translate="label" type="select" sortOrder="10" showInDefault="1" showInWebsite="1" showInStore="1">
         <label>Shopper country</label>
-        <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
-        <!--config_path></config_path--> <!--TODO: New field? Yes-->
+        <source_model>Adyen\Payment\Model\Config\Source\ShopperCountryMode</source_model>
+        <config_path>payment/adyen_hpp/shoppercountry</config_path>
     </field>
-    <field id="ratepay_id" translate="label" type="text" sortOrder="20" showInDefault="1" showInWebsite="1" showInStore="1">
+    <field id="shopper_specific_country" translate="label" type="select" sortOrder="20" showInDefault="1" showInWebsite="1" showInStore="0">
+        <label>Custum shopper country</label>
```
Aleffio:
```suggestion
        <label>Custom shopper country</label>
```

Setup/UpgradeData.php
```diff
@@ -306,4 +308,37 @@ private function setKarCaptureMode($connection)
...
+            ->where('value <> "" AND value IS NOT NULL');
+        $configCountryCodeValues = $connection->fetchAll($select);
+
+        foreach ($configCountryCodeValues as $configCountryCodeValue) {
+            $scope = $configCountryCodeValue['scope'];
+            $scopeId = $configCountryCodeValue['scope_id'];
+
+            if (isset($configCountryCodeValue)) {
+                $shoppercountry = 1;
+            }
```
acampos1916: $shoppercountry contains the path, not the value.
And I think this IF won't be necessary to assign the value to a variable, the SQL query already returns all rows where country_code is not empty nor NULL, so it will always be true

Setup/UpgradeData.php
```diff
@@ -306,4 +308,37 @@ private function setKarCaptureMode($connection)
...
+            $this->configWriter->save(
+                $shoppercountry,
+                'store_level',
```
acampos1916:
```suggestion
                1,
```

acampos1916: You want to make shoppercountry=1 for every row that is iterated because country_code is not empty nor NULL

Helper/Data.php
```diff
@@ -1901,4 +1901,21 @@ public function getPspReferenceSearchUrl($pspReference, $liveEnvironment)
...
+     * Gives back CountryCode configuration values
+     *
+     * @param null $storeId
+     * @return mixed
+     */
+    public function getAdyenCountryCodeConfigData($storeId = null)
+    {
+        $shopperLocale = trim($this->_adyenHelper->getAdyenHppConfigData('shopper_locale', $storeId));
+        $countryCode = trim($this->_adyenHelper->getAdyenHppConfigData('country_code', $storeId));
+        if ($countryCode == false && $shopperLocale == 1) {
```
acampos1916: @AlexandrosMor from [this line](https://github.com/Adyen/adyen-magento2/blob/102e262d6bc7c73c25c90c76354d3a18d9db0b92/Gateway/Command/PayByMailCommand.php#L155) it looks like country_code could be empty instead of FALSE. Have you tested generating a PayByMail link or PaymentMethods call with the new shopper_country=1 and shopper_country=0? I think this IF would never evaluate to true and the country_code is not sent

etc/adminhtml/system/payment_methods/local_payment_methods/advanced_options.xml
```diff
@@ -26,32 +26,42 @@
...
+        <config_path>payment/adyen_hpp/shopper_country</config_path>
     </field>
-    <field id="ratepay_id" translate="label" type="text" sortOrder="20" showInDefault="1" showInWebsite="1" showInStore="1">
+    <field id="shopper_specific_country" translate="label" type="select" sortOrder="20" showInDefault="1" showInWebsite="1" showInStore="0">
+        <label>Custom shopper country</label>
+        <depends>
+            <field id="shopper_country">1</field>
+        </depends>
+        <source_model>Magento\Directory\Model\Config\Source\Country</source_model>
+        <can_be_empty>1</can_be_empty>
```
acampos1916:
```suggestion
        <validate>required-entry</validate>
```

acampos1916: If the country_code dropdown shows we can validate that it has a selection with this line

etc/adminhtml/system/payment_methods/local_payment_methods/advanced_options.xml
```diff
@@ -26,32 +26,42 @@
...
-        <source_model>Magento\Directory\Model\Config\Source\Country</source_model>
-        <can_be_empty>1</can_be_empty>
-        <config_path>payment/adyen_hpp/specificcountry</config_path>
+    <field id="specificcountry" translate="label" type="multiselect" sortOrder="60" showInDefault="1" showInWebsite="1" showInStore="0">
+    <label>Payment from specific countries</label>
+    <depends>
+        <field id="allowspecific_local_payment_methods">1</field>
+    </depends>
+    <source_model>Magento\Directory\Model\Config\Source\Country</source_model>
+    <can_be_empty>1</can_be_empty>
```
acampos1916: If the country_code dropdown shows we can validate that it has a selection with this line
```suggestion
    <validate>required-entry</validate>
```

Issue #488, opened 2019-07-30
Order status pending when payment cancelled and later authorized


evs-xsarus:
**Magento version**: 2.2.7
**Plugin version**: 3.0.0
**Description**

I have an order with this payment history:

* 25 jul. 2019 10:53:37 Pending Customer Notified / Order created
* 25 jul. 2019 10:55:13 Customer Not Notified / Order status is set to 'Canceled' and order items are cancelled.
    > Adyen Result URL response: 
    > authResult: CANCELLED 
    > pspReference: 
    > paymentMethod: ideal
* 25 jul. 2019 11:00:05 Canceled Customer Not Notified / Order status is set to Pending but order items stay cancelled.
    > Adyen HTTP Notification(s): 
    > eventCode: AUTHORISATION 
    > pspReference: xxxxxxxxxxxxxxxxx 
    > paymentMethod: ideal 
    > success: true 
    > reason:ideal
* 25 jul. 2019 11:00:05 Pending Customer Not Notified / Order stays on pending but will never get an invoice because the items are still cancelled.
 > Payment is authorised waiting for capture

I believe that the order items must also be uncancelled by this Adyenplugin since it's weird to have a pending order with only cancelled items.

ryanpalmerweb:
Hi Adyen,

We've noticed this problem in our stores aswell, we are using plugin version 4.0.1

When the customer returns from a failed payment, it looks like the order ID is not being changed, and they are being allowed to go back to Adyen to make a second attempt.

As the original order is cancelled, the success notification is unable to change the order to a processing state.

I would expect the behaviour here to be that a new order number was issued, or that the module undoes the cancellation.

I've attached the access log so that you can see the customer flow, with any identifying information removed.

> (customer ip address removed) - - [21/Aug/2019:17:03:32 +0100] "GET /adyen/process/redirect/ HTTP/2.0" 200 1998 "https://www.ourwebsite.co.uk/checkout/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.100 Safari/537.36"

[...]

PR #679, opened 2020-03-31
[PW-2174]: Removing sync order cancellation for 3DS1 failure response

acampos1916:
**Description**
When 3DS payments fail the plugin should cancel the order only on notification processing (Authorization Success False). This fixes issues where orders get conflicting synchronous responses.

**Tested scenarios**
3DS1 failed payments do not cancel the newly created order.
Authorization Success False notifications do cancel the order

**Fixed issue**:  PW-2174
#488

### DIFF
--- a/Controller/Process/Redirect.php
+++ b/Controller/Process/Redirect.php
@@ -224,7 +224,7 @@ class Redirect extends \Magento\Framework\App\Action\Action
 
 						$this->_redirect('checkout/onepage/success', ['_query' => ['utm_nooverride' => '1']]);
 					} else {
-						$order->addStatusHistoryComment(__('3D-secure validation was unsuccessful. This order will be cancelled when the relevant notification has been processed.'))->save();
+						$order->addStatusHistoryComment(__('3D-secure validation was unsuccessful. This order will be cancelled when the related notification has been processed.'))->save();
 
 						// reactivate the quote
 						$session = $this->_getCheckout();
--- a/Controller/Process/Redirect.php
+++ b/Controller/Process/Redirect.php
@@ -224,6 +224,11 @@ class Redirect extends \Magento\Framework\App\Action\Action
 
 						$this->_redirect('checkout/onepage/success', ['_query' => ['utm_nooverride' => '1']]);
 					} else {
+					    /*
+					     * Since responseCode!='Authorised' the order could be cancelled immediately,
+					     * but redirect payments can have multiple conflicting responses.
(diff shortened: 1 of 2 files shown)


Model/AdyenThreeDS2Process.php
```diff
@@ -145,13 +145,6 @@ public function initiate($payload)
         if($result['resultCode'] != 'Authorised') {
             $this->checkoutSession->restoreQuote();
 
-            // Always cancel the order if the paymenth has failed
-            if (!$order->canCancel()) {
-                $order->setState(\Magento\Sales\Model\Order::STATE_NEW);
-            }
-
-            $order->cancel()->save();
-
```
cyattilakiss: This is a synchronous response so the cancellation is valid

acampos1916: Restoring this cancellation because it isn't redirect

Controller/Process/Redirect.php
```diff
@@ -224,12 +224,7 @@ public function execute()
 
 						$this->_redirect('checkout/onepage/success', ['_query' => ['utm_nooverride' => '1']]);
 					} else {
-						$order->addStatusHistoryComment(__('3D-secure validation was unsuccessful.'))->save();
-
-						// Move the order from PAYMENT_REVIEW to NEW, so that can be cancelled
-						$order->setState(\Magento\Sales\Model\Order::STATE_NEW);
-						$this->_adyenHelper->cancelOrder($order);
-						$this->messageManager->addErrorMessage("3D-secure validation was unsuccessful");
+						$order->addStatusHistoryComment(__('3D-secure validation was unsuccessful. This order will be cancelled when the related notification has been processed.'))->save();
```
rikterbeek, line 227: I would quickly explain here why we don't cancel it immediately. So that there is a possibility that there is a successful authentication as well.

acampos1916: @rikterbeek added a comment to explain it, if you think the wording is ok we can use it: https://github.com/Adyen/adyen-magento2/blob/d1e6d13c2c4abd584f5272e5cf704c68567dd368/Controller/Process/Redirect.php#L227-L231

Beagon:
Hi,

I seem to still encounter this issue. I've seen this on 2 yesterday one if which was a PayPal order and the other one iDeal. We're running version 6.4.0 but are upgrading to 6.6.0 but I don't feel like this has to do with 6.4.0 since this should be fixed in 6.2.

Sincerely,

Robin

PR #681, opened 2020-04-01
[WIP][PW-2138]Update components version to 3.6.1

AlexandrosMor:
**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-google-pay-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-google-pay-method.js
@@ -100,7 +100,7 @@ define(
                     }
                 });
                 var googlepay = self.checkoutComponent.create('paywithgoogle', {
-                    showPayButton: true,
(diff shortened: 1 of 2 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-google-pay-method.js
```diff
@@ -100,7 +100,7 @@ define(
                     }
                 });
                 var googlepay = self.checkoutComponent.create('paywithgoogle', {
-                    showPayButton: true,
+                    showPayButton: false,
```
cyattilakiss, line 103: for Google pay we need to show the pay button - sorry that is an exception
The setting you are looking for is in the cc and hpp method js files.
Probably you will need to create the `showPayButton` configuration because it was by default false so we didn't need to do anything with it until now

cyattilakiss:
I am going to pick this up in the PW-2281

PR #683, opened 2020-04-01
Fix order payment status on non-3DS orders

msilvagarcia:
If an order is made with Credit Card but without 3DS information required, the indexes `threeDSType` and `threeDS2Token` might not exist. To prevent PHP throwing a notice, we need to check if these values exist before trying to read them.


cyattilakiss:
@msilvagarcia I will close this ticket - please see my answer on mm
(Just don't want anyone to review it :) )

Issue #662, opened 2020-03-05
Alternative payment methods allow user to bypass required payment address


acjbizar:
**Describe the bug**
On the one-page checkout, the alternative payment methods of Adyen (in our case iDEAL and Klarna), seem unconcerned with whether the payment address is filled in correctly (or at all).

Native Magento payment methods only activate the payment button when the payment address is filled in correctly. The Adyen iDEAL payment method only requires a vendor/bank to be selected, and the Adyen Klarna payment method does not require any action.

**To Reproduce**
1. Go to `/checkout`
2. Click on on one of the Adyen (alternative) payment methods
3. Leave the payment address form empty
4. Watch the payment button get activated.

**Expected behavior**
I would expect the payment button to remain disabled until the required payment address form is filled in correctly, as appears to be the case with the native Magento payment methods.

**Screenshots**
![image](https://user-images.githubusercontent.com/411726/75979100-17035100-5ee0-11ea-80b1-551486df912b.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 80

**Additional context**
I see a difference in the `data-bind` of the payment button in comparison with the Magento native payment methods.

Magento (native) payment methods:
```
    click: placeOrder,
    attr: {'title': $t('Place Order')},
    enable: (getCode() == isChecked()),
    css: {disabled: !isPlaceOrderActionAllowed()}
```

Adyen (alternative) payment methods:
```
    click: $parent.continueToAdyenBrandCode, enable: placeOrderAllowed() && (value == $parent.isBrandCodeChecked())
```

I am guessing this causes it to miss some of the required conditions.

cyattilakiss:
Hi @acjbizar ,

Thank you for raising this issue.
I just checked it and indeed, we let the user place the order when they disable the 'My billing and shipping address are the same' checkbox but since they don't update the address with the empty fields - because Magento by default does validation on it - we still use the billing address which is still there so the payment will go through.
I picked up this ticket into our backlog.

Best,
Attila
Adyen

acjbizar:
Hi @cyattilakiss,

Thanks for your reply. The thing is though, this website does not use a shipping address, since their products are not shipped. (Instead, the user selects a pick-up location.) So this actually becomes a business critical issue for us.

[...]

PR #685, opened 2020-04-06
Add extra check to submit button

acjbizar:
**Description**
Some webshops work with only a billing address because products need to be picked up from a location. If this is the case, the billing address should be filled in, and valid before a payment method can be used. This is already the case with payment methods native to Magento 2, but not with this Adyen module. Because of this, orders could be placed using Adyen payment methods, without adding an address at all. This extra check fixes this.

**Fixed issue**:  #662.


cyattilakiss:
Hi @acjbizar ,

Thank you for opening the pull request. We are going to test your solution and review the code this week.

Best,
Attila
Adyen

acjbizar:
Hi! It appears my branch is failing this automated `continuous-integration/travis-ci/pr` build, but I don’t think this is related to my code.

cyattilakiss:
Hi @acjbizar ,

Don't worry about that, we still need to figure out how the external pull requests can be validated by our automated tests.
We just merged your code, it will be released in our next release soon.
Thanks for the contribution :)

Best,
Attila
Adyen

PR #692, opened 2020-04-15
[PW-2199]: Add configuration to toggle order cancellation on notification processing

acampos1916:
**Description**
We've added a config value "notifications_can_cancel" as a check to avoid order cancellation on notification processing. Defaults to 1 or true.

A field in system.xml is not being added. To change the config value the config.xml or system.xml files must be extended.

**Tested scenarios**
Processing notifications with event_code CAPTURE_FAILED, CANCELLATION, CANCELLED, CANCEL_OR_REFUND, OFFER_CLOSED. Order is not cancelled and a new message is logged in cronjob.log: "Order can not be canceled based on the plugin configuration"

**Fixed issue**:  PW-2199


Model/Cron.php
```diff
@@ -829,6 +829,12 @@ protected function _retrieveLast4DigitsFromReason($reason)
      */
     protected function _holdCancelOrder($ignoreHasInvoice)
     {
+        $canCancel = $this->adyenHelper->getAdyenAbstractConfigData('notifications_can_cancel');
+        if(!$canCancel){
+            $this->_adyenLogger->addAdyenNotificationCronjob('Order can not be canceled based on the plugin configuration');
+            return;
+        }
+
```
Aleffio, line 837: for OFFER_CLOSED, before calling **_holdCancelOrder**, in case the order cannot be cancelled, we set the state to STATE_NEW (line 1003).
Can we add the same check there for _notifications_can_cancel_ ? If we don't want to cancel the order, we also don't want to set its state to new

acampos1916: Good catch. Since this also applies for ORDER_CLOSED we decided to create a new helper class for Config values. Now we're checking the notifications_can_cancel on ORDER_CLOSED and OFFER_CLOSED before changing the state to NEW.

Helper/Config.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        $path = implode("/", [self::XML_PAYMENT_PREFIX, $xmlPrefix, $field]);
+
+        if (!$flag) {
+            return $this->scopeConfig->getValue($path, \Magento\Store\Model\ScopeInterface::SCOPE_STORE, $storeId);
+        } else {
+            return $this->scopeConfig->isSetFlag($path, \Magento\Store\Model\ScopeInterface::SCOPE_STORE, $storeId);
+        }
+    }
+}
```
cyattilakiss, line 87: add an empty line and auto format PSR2 if anything missing

Helper/Config.php
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Retrieve information from payment configuration
+     *
+     * @param string $field
+     * @param string $xmlPrefix
+     * @param int $storeId
+     * @param bool|false $flag
+     * @return bool|mixed
+     */
+    private function getConfigData($field, $xmlPrefix, $storeId, $flag = false)
```
msilvagarcia, line 77: Where is the `$flag` parameter used?

msilvagarcia: I mean, the only possible call to this method sets it to `true`....

acampos1916: I wanted to include that function just as it was on Data.php because we are going to outsource more and more of the methods already there, and some set the flag to false.

So it is not being used right now but will be.

Helper/Config.php
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * Copyright (c) 2020 Adyen BV (https://www.adyen.com/)
+ * See LICENSE.txt for license details.
+ *
+ * Author: Adyen <magento@adyen.com>
+ */
+
+namespace Adyen\Payment\Helper;
+
+use Magento\Framework\App\Helper\AbstractHelper;
```
rikterbeek, line 26: I don't think we need to extend from AbstractHelper as this has many imports in the constructor we don't need for this simple helper.

acampos1916: Agreed, we can remove it for now.

Helper/Config.php
acampos1916: @msilvagarcia let me know if you think we should simplify this class further. Rik proposed to remove the AbstractHelper dependency which is now done

PR #695, opened 2020-04-17
[PW-2264]: New data builder for LPMs to include storePaymentMethod flag based on configuration

acampos1916:
**Description**
For the new stored_local_payments_active we're implementing a new data builder that includes the storePaymentMethod in the /payments request based on the value of that config field.

**Tested scenarios**
Payment using alternative payment methods with the stored_local_payments_active enabled/disabled sends a payment request with storePaymentMethod true/false

**Fixed issue**:  PW-2264

### DIFF
--- a/Helper/Requests.php
+++ b/Helper/Requests.php
@@ -373,7 +373,7 @@ class Requests extends AbstractHelper
      * @param $storeId
      * @param $payment
      */
-    public function buildRecurringAlternativePaymentMethodsData($request = [], $areaCode, int $storeId, $additionalData)
+    public function buildRecurringAlternativePaymentMethodsData($request = [], $areaCode, int $storeId)
     {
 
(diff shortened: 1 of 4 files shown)


Helper/Requests.php
```diff
@@ -367,6 +367,30 @@ public function buildRecurringData($request = [], $areaCode, int $storeId, $addi
         return $request;
     }
 
+    /**
+     * @param $request
+     * @param $areaCode
+     * @param $storeId
+     * @param $payment
+     */
+    public function buildRecurringAlternativePaymentMethodsData($request = [], $areaCode, int $storeId, $additionalData)
```
Aleffio, line 376:
```suggestion
    public function buildRecurringAlternativePaymentMethodsData($request = [], $areaCode, int $storeId)
```
not used, right?

acampos1916: Right. Removed it from the call and definition.

PR #713, opened 2020-05-15
[PW-2387]: IP validation implementation 

acampos1916:


Helper/IpAddress.php
```diff
@@ -0,0 +1,133 @@
...
+     * @param bool $fullCacheKeyUpdate
+     * @return bool
+     */
+    public function isIpAddressValid($ipAddresses)
+    {
+        if (empty($ipAddresses)) {
+            return false;
+        }
+
+        $cachedIpsArray = $this->getIpAddressesFromCache();
```
rikterbeek, line 85: Add some validation here. If cache key is empty what can happen if the cron is not running. Let's retrieve it here. Maybe it is even better to remove the full cron and always do it here with a cache key that we define for one day.

etc/adminhtml/system/adyen_security.xml
```diff
@@ -0,0 +1,42 @@
...
+    <group id="adyen_security" translate="label" type="text" sortOrder="90" showInDefault="1" showInWebsite="1" showInStore="1">
+        <label><![CDATA[Security]]></label>
+        <frontend_model>Magento\Config\Block\System\Config\Form\Fieldset</frontend_model>
+        <field id="notifications_ip_hmac_check" translate="label" type="select" sortOrder="10" showInDefault="1" showInWebsite="1" showInStore="1">
+            <label>Check notification's IP address and HMAC signature</label>
+            <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+            <config_path>payment/adyen_abstract/notifications_ip_hmac_check</config_path>
+            <comment>
+                <![CDATA[
+                If enabled notifications will be accepted only if the IP address matches Adyen's servers and the HMAC
```
cyattilakiss, line 35:
```suggestion
                Enabled notifications will be accepted only if the IP address matches Adyen's servers and the HMAC
```

Helper/IpAddress.php
acampos1916: I added a log message to the checks on cron (cronjob.log) and helper (debug.log):

https://github.com/Adyen/adyen-magento2/commit/f6ed1dbc77052c1363451a5123f5e2d3169e08af#diff-28e83357bad28c013fdb1ead16fee88aR98

https://github.com/Adyen/adyen-magento2/commit/f6ed1dbc77052c1363451a5123f5e2d3169e08af#diff-71e336a11f03788f9a758eab93b6575eR60

Test/Unit/Helper/IpAddressTest.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->ipAddressHelper = new \Adyen\Payment\Helper\IpAddress(
+            $ipAddressUtil,
+            $cache,
+            $serializer
+        );
+    }
+
+    public function testIsIpAddressValid()
+    {
+        $this->assertTrue(is_bool($this->ipAddressHelper->isIpAddressValid([gethostbyname('outgoing1.adyen.com')])));
```
msilvagarcia, line 55: This test is not unitary and will depend on the network layer. I'd just feed the `isIpAddressValid` method with a mocked array of results (like a couple of IP addresses)

Also, we could cover more cases, such as:
* IP addresses that are not setup within Adyen
* invalid input for IP address

acampos1916: Thank you @msilvagarcia. I made some adjustments to the test. I'm mocking the result of some external functions and provide a set of IP addresses to check, including an invalid one. What do you think?

Helper/IpAddress.php
acampos1916: Added a cached IP addresses update on notification receiving @rikterbeek  https://github.com/Adyen/adyen-magento2/pull/713/files#diff-28e83357bad28c013fdb1ead16fee88aR100

PR #716, opened 2020-05-21
[PW-2385] - Use mandatory HMAC validation for notifications

AlexandrosMor:
**Description**
Currently we just check the username password but validating the notification agains the HMAC key would be more secure.

**Tested scenarios**
If configuration field is enabled:
If HMAC key is not correct or not sent, do not accept notification and write error log
If HMAC key is correct accept notification and no error log

### DIFF
--- a/Controller/Process/Json.php
+++ b/Controller/Process/Json.php
@@ -23,6 +23,7 @@
 
 namespace Adyen\Payment\Controller\Process;
 
+use Adyen\Util\HmacSignature;
 use Symfony\Component\Config\Definition\Exception\Exception;
 use Magento\Framework\App\Request\Http as Http;
 
@@ -68,6 +69,11 @@ class Json extends \Magento\Framework\App\Action\Action
     protected $ipAddressHelper;
 
     /**
+     * @var HmacSignature
+     */
+    private $hmacSignature;
+
+    /**
      * Json constructor.
      *
      * @param \Magento\Framework\App\Action\Context $context
@@ -76,6 +82,7 @@ class Json extends \Magento\Framework\App\Action\Action
      * @param \Magento\Framework\Serialize\SerializerInterface $serializer
      * @param \Adyen\Payment\Helper\Config $configHelper
      * @param \Adyen\Payment\Helper\IpAddress $ipAddressHelper
+     * @param HmacSignature $hmacSignature
      */
     public function __construct(
         \Magento\Framework\App\Action\Context $context,
@@ -83,7 +90,8 @@ class Json extends \Magento\Framework\App\Action\Action
         \Adyen\Payment\Logger\AdyenLogger $adyenLogger,
         \Magento\Framework\Serialize\SerializerInterface $serializer,
         \Adyen\Payment\Helper\Config $configHelper,
-        \Adyen\Payment\Helper\IpAddress $ipAddressHelper
+        \Adyen\Payment\Helper\IpAddress $ipAddressHelper,
+        HmacSignature $hmacSignature
     ) {
         parent::__construct($context);
         $this->_objectManager = $context->getObjectManager();
@@ -93,6 +101,7 @@ class Json extends \Magento\Framework\App\Action\Action
         $this->serializer = $serializer;
         $this->configHelper = $configHelper;
         $this->ipAddressHelper = $ipAddressHelper;
+        $this->hmacSignature = $hmacSignature;
 
         // Fix for Magento2.3 adding isAjax to the request params
         if (interface_exists("\Magento\Framework\App\CsrfAwareActionInterface")) {
@@ -202,6 +211,12 @@ class Json extends \Magento\Framework\App\Action\Action
                 $this->_adyenLogger->addAdyenNotification(
                     "Notification has been rejected because the IP address could not be verified"
                 );
+                return false;
+            }
+
+            //Validate the Hmac calculation
+            if (!$this->hmacSignature->isValidNotificationHMAC($this->configHelper->getNotificationsHmacKey(), $response)) {
+                $this->_adyenLogger->addAdyenNotification('HMAC key validation failed');
                 return false;
             }
         }
--- a/Controller/Process/Json.php
+++ b/Controller/Process/Json.php
@@ -214,11 +214,11 @@ class Json extends \Magento\Framework\App\Action\Action
                 return false;
             }
 
-            if ($this->_isHmacSupportedEventCode($response)) {
+            if ($this->hmacSignature->isHmacSupportedEventCode($response)) {
                 //Validate the Hmac calculation
                 if (!$this->hmacSignature->isValidNotificationHMAC($this->configHelper->getNotificationsHmacKey(),
                     $response)) {
-                    $this->_adyenLogger->addAdyenNotification('HMAC key validation failed '. print_r($response, 1));
+                    $this->_adyenLogger->addAdyenNotification('HMAC key validation failed ' . print_r($response, 1));
                     return false;
                 }
             }
@@ -449,27 +449,5 @@ class Json extends \Magento\Framework\App\Action\Action
             ->setHeader('Content-Type', 'text/html')
             ->setBody($message);
         return;
-    }
-
-    /**
-     * Returns true when the event code support HMAC validation
-     *
-     * @param $response
-     */
-    protected function _isHmacSupportedEventCode($response)
-    {
-        $eventCodes = array("ADVICE_OF_DEBIT","AUTHORISATION","AUTHORISATION_PENDING","AUTHORISE_REFERRAL","CANCELLATION",
-            "CANCEL_OR_REFUND", "CAPTURE","CAPTURE_FAILED","CAPTURE_WITH_EXTERNAL_AUTH","CHARGEBACK","CHARGEBACK_REVERSED",
-            "DEACTIVATE_RECURRING","FRAUD_ONLY","FUND_TRANSFER","HANDLED_EXTERNALLY","MANUAL_REVIEW_ACCEPT","NOTIFICATION_OF_CHARGEBACK",
-            "NOTIFICATION_OF_FRAUD", "OFFER_CLOSED","ORDER_OPENED","PAIDOUT_REVERSED","PAYOUT_DECLINE","PAYOUT_EXPIRE","PAYOUT_THIRDPARTY",
-             "PREARBITRATION_LOST","PREARBITRATION_WON","PROCESS_RETRY","RECURRING_CONTRACT","REFUND","REFUNDED_REVERSED","REFUND_FAILED",
-            "REFUND_WITH_DATA", "REQUEST_FOR_INFORMATION","SECOND_CHARGEBACK","SUBMIT_RECURRING","VOID_PENDING_REFUND",
-            "POSTPONED_REFUND","TECHNICAL_CANCEL","AUTHORISATION_ADJUSTMENT","CANCEL_AUTORESCUE","AUTORESCUE");
-        if (array_key_exists('eventCode', $response)) {
-            if (in_array($response['eventCode'], $eventCodes)) {
-                return true;
-            }
-        }
-        return false;
     }
 }
--- a/Controller/Process/Json.php
+++ b/Controller/Process/Json.php
@@ -218,12 +218,9 @@ class Json extends \Magento\Framework\App\Action\Action
                 //Validate the Hmac calculation
                 if (!$this->hmacSignature->isValidNotificationHMAC($this->configHelper->getNotificationsHmacKey(),
                     $response)) {
-                    $this->_adyenLogger->addAdyenNotification('HMAC key validation failed');
+                    $this->_adyenLogger->addAdyenNotification('HMAC key validation failed '. print_r($response, 1));
                     return false;
                 }
-                $this->_adyenLogger->addAdyenNotification('In the list');
-            }else{
-                $this->_adyenLogger->addAdyenNotification('Event code not in my array');
             }
         }
 
--- a/Controller/Process/Json.php
+++ b/Controller/Process/Json.php
@@ -214,10 +214,16 @@ class Json extends \Magento\Framework\App\Action\Action
                 return false;
             }
 
-            //Validate the Hmac calculation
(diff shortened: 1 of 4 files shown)


etc/adminhtml/system/adyen_security.xml
```diff
@@ -38,5 +38,17 @@
                 ]]>
             </comment>
         </field>
+        <field id="notification_hmac_key" translate="label" type="text" sortOrder="20" showInDefault="1"
```
cyattilakiss, line 41: Please add test and live HMAC key configuration, I just checked the prestashop repo where there is only one but it is much harder to manage :(
When retrieving the configuration check the plugin configuration if test or live mode is enabled and retrieve the hmac key based on that.

cyattilakiss: Also please declare this as a sensitive config - you can find an example in this PR: https://github.com/Adyen/adyen-magento2/pull/707 if help needed please ask @acampos1916

AlexandrosMor: done 👍

Controller/Process/Json.php
```diff
@@ -432,4 +450,26 @@ protected function _returnResult($message)
             ->setBody($message);
         return;
     }
+
+    /**
+     * Returns true when the event code support HMAC validation
+     *
+     * @param $response
+     */
+    protected function _isHmacSupportedEventCode($response)
```
rikterbeek, line 459: please remove the underscore we don't use this anymore as described in the ruleset of PSR2

Controller/Process/Json.php
```diff
@@ -432,4 +450,26 @@ protected function _returnResult($message)
...
+    {
+        $eventCodes = array("ADVICE_OF_DEBIT","AUTHORISATION","AUTHORISATION_PENDING","AUTHORISE_REFERRAL","CANCELLATION",
+            "CANCEL_OR_REFUND", "CAPTURE","CAPTURE_FAILED","CAPTURE_WITH_EXTERNAL_AUTH","CHARGEBACK","CHARGEBACK_REVERSED",
+            "DEACTIVATE_RECURRING","FRAUD_ONLY","FUND_TRANSFER","HANDLED_EXTERNALLY","MANUAL_REVIEW_ACCEPT","NOTIFICATION_OF_CHARGEBACK",
+            "NOTIFICATION_OF_FRAUD", "OFFER_CLOSED","ORDER_OPENED","PAIDOUT_REVERSED","PAYOUT_DECLINE","PAYOUT_EXPIRE","PAYOUT_THIRDPARTY",
+             "PREARBITRATION_LOST","PREARBITRATION_WON","PROCESS_RETRY","RECURRING_CONTRACT","REFUND","REFUNDED_REVERSED","REFUND_FAILED",
+            "REFUND_WITH_DATA", "REQUEST_FOR_INFORMATION","SECOND_CHARGEBACK","SUBMIT_RECURRING","VOID_PENDING_REFUND",
+            "POSTPONED_REFUND","TECHNICAL_CANCEL","AUTHORISATION_ADJUSTMENT","CANCEL_AUTORESCUE","AUTORESCUE");
```
rikterbeek, line 467: lets use [] instead of array as m2 supports php7+. as well check your format looks a bit off here.

Controller/Process/Json.php
```diff
@@ -432,4 +450,26 @@ protected function _returnResult($message)
...
+        if (array_key_exists('eventCode', $response)) {
+            if (in_array($response['eventCode'], $eventCodes)) {
```
rikterbeek, line 469: we can merge this into on if statement. in_array is enough we don't need the array_key_exists method.

Helper/Config.php
```diff
@@ -80,6 +99,32 @@ public function getNotificationsIpHmacCheck($storeId = null)
...
+     * Retrieve key for notifications_hmac_key
+     *
+     * @param int $storeId
+     * @return string
+     */
+    public function getNotificationsHmacKey($storeId = null)
+    {
+        $key = "";
+        if ($this->adyenHelper->isDemoMode($storeId)) {
+            $key = (string)$this->getConfigData(
```
rikterbeek, line 112: do we need to parse this to a string ?

Helper/Config.php
```diff
@@ -80,6 +99,32 @@ public function getNotificationsIpHmacCheck($storeId = null)
...
+                self::XML_NOTIFICATIONS_HMAC_KEY_TEST,
+                self::XML_ADYEN_ABSTRACT_PREFIX,
+                $storeId,
+                false
+            );
+        } else {
+            $key = (string)$this->getConfigData(
```
rikterbeek, line 119: do we need to parse this to a string ?

etc/di.xml
```diff
@@ -1075,6 +1075,8 @@
                 <item name="payment/adyen_apple_pay/merchant_identifier_test" xsi:type="string">1</item>
                 <item name="payment/adyen_apple_pay/merchant_identifier_live" xsi:type="string">1</item>
                 <item name="payment/adyen_google_pay/merchant_identifier" xsi:type="string">1</item>
+                <item name="payment/adyen_abstract/notification_hmac_key_test" xsi:type="string">1</item>
```
rikterbeek, line 1078: default should be turned off

etc/di.xml
```diff
@@ -1075,6 +1075,8 @@
...
+                <item name="payment/adyen_abstract/notification_hmac_key_live" xsi:type="string">1</item>
```
rikterbeek, line 1079: default should be off

etc/adminhtml/system/adyen_security.xml
```diff
@@ -38,5 +38,29 @@
...
         </field>
+        <field id="notification_hmac_key_test" translate="label" type="text" sortOrder="20" showInDefault="1"
+               showInWebsite="1" showInStore="1">
+            <label>Hmac key test</label>
+            <backend_model>Magento\Config\Model\Config\Backend\Encrypted</backend_model>
+            <config_path>payment/adyen_abstract/notification_hmac_key_test</config_path>
+            <tooltip>
+                <![CDATA[
+                Set HMAC key at first on Adyen so new notifications are sent with the key. To learn more about these settings refer to
+                <a target="_blank" href="https://docs.adyen.com/classic-integration/hosted-payment-pages/hmac-signature-calculation#page-introduction">Adyen documentation</a>.
```
Aleffio, line 49: this page is for HMAC signature for HPP skins, I think you want to point to the page for HMAC key for webhooks instead:
```suggestion
                <a target="_blank" href="https://docs.adyen.com/development-resources/webhooks/verify-hmac-signatures">Adyen documentation</a>.
```

etc/adminhtml/system/adyen_security.xml
```diff
@@ -38,5 +38,29 @@
...
+        </field>
+        <field id="notification_hmac_key_live" translate="label" type="text" sortOrder="30" showInDefault="1"
+               showInWebsite="1" showInStore="1">
+            <label>Hmac key live</label>
+            <backend_model>Magento\Config\Model\Config\Backend\Encrypted</backend_model>
+            <config_path>payment/adyen_abstract/notification_hmac_key_live</config_path>
+            <tooltip>
+                <![CDATA[
+                Set HMAC key at first on Adyen so new notifications are sent with the key. To learn more about these settings refer to
+                <a target="_blank" href="https://docs.adyen.com/classic-integration/hosted-payment-pages/hmac-signature-calculation#page-introduction">Adyen documentation</a>.
```
Aleffio, line 61:
```suggestion
                <a target="_blank" href="https://docs.adyen.com/development-resources/webhooks/verify-hmac-signatures">Adyen documentation</a>.
```

Controller/Process/Json.php
```diff
@@ -204,6 +213,15 @@ protected function _processNotification($response, $notificationMode)
                 );
                 return false;
             }
+
+            if ($this->_isHmacSupportedEventCode($response)) {
+                //Validate the Hmac calculation
+                if (!$this->hmacSignature->isValidNotificationHMAC($this->configHelper->getNotificationsHmacKey(),
+                    $response)) {
+                    $this->_adyenLogger->addAdyenNotification('HMAC key validation failed '. print_r($response, 1));
```
Aleffio, line 221:
```suggestion
                    $this->adyenLogger->addAdyenNotification('HMAC key validation failed '. print_r($response, 1));
```

etc/di.xml
Aleffio: this value should just say that the hmac key is a "sensitive" type, right?

cyattilakiss: yes, this is just to separate these config fields as sensitive ones

Helper/Config.php
```diff
@@ -80,6 +99,32 @@ public function getNotificationsIpHmacCheck($storeId = null)
...
+    /**
+     * Retrieve key for notifications_hmac_key
+     *
+     * @param int $storeId
+     * @return string
+     */
+    public function getNotificationsHmacKey($storeId = null)
+    {
+        $key = "";
+        if ($this->adyenHelper->isDemoMode($storeId)) {
```
acampos1916, line 111: Since isDemoMode() is a config function and this is the Config helper we can move it to this class now that we need it (flag/bool kind like getNotificationsCanCancel()). We want to ease the dependencies on the Data.php helper as it is getting bigger and will probably be slimmed down in the future

AlexandrosMor: Good idea. I removed the Data.php dependency and add an idDemoMode method

etc/di.xml
cyattilakiss: @AlexandrosMor this one needs to be 1 because it is sensitive config

etc/di.xml
cyattilakiss: @AlexandrosMor this one needs to be 1 because it is sensitive config

Helper/Config.php
```diff
@@ -80,6 +91,43 @@ public function getNotificationsIpHmacCheck($storeId = null)
...
+    }
+    /**
+     * Returns true if the enviroment is set to test, false for live
+     *
+     * @param int $storeId
+     * @return bool
+     */
+    public function isDemoMode($storeId = null)
+    {
+        return $this->getConfigData('demo_mode', 'adyen_abstract', $storeId, true);
```
acampos1916, line 128: 'adyen_abstract' is already in a constant in this class, and 'demo_mode' can be a new constant to follow the class and method syntax. You can copy getNotificationsIpHmacCheck()

etc/di.xml
```diff
@@ -1075,6 +1075,8 @@
                 <item name="payment/adyen_apple_pay/merchant_identifier_test" xsi:type="string">1</item>
                 <item name="payment/adyen_apple_pay/merchant_identifier_live" xsi:type="string">1</item>
                 <item name="payment/adyen_google_pay/merchant_identifier" xsi:type="string">1</item>
+                <item name="payment/adyen_abstract/notification_hmac_key_test" xsi:type="string">0</item>
+                <item name="payment/adyen_abstract/notification_hmac_key_live" xsi:type="string">0</item>
```
acampos1916, line 1079: There was some confusion on previous comments. These should be '1' indeed because both are sensitive.

etc/di.xml
rikterbeek: yes true

Helper/Config.php
AlexandrosMor: Ok I replaced it with the constant

Helper/Config.php
```diff
@@ -80,6 +91,43 @@ public function getNotificationsIpHmacCheck($storeId = null)
...
 
+    /**
+     * Retrieve key for notifications_hmac_key
+     *
+     * @param int $storeId
+     * @return string
+     */
+    public function getNotificationsHmacKey($storeId = null)
+    {
+        $key = "";
```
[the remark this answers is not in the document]
AlexandrosMor: Removed! :)

etc/di.xml
AlexandrosMor: Yes revert to 1

Controller/Process/Json.php
```diff
@@ -204,6 +213,15 @@ protected function _processNotification($response, $notificationMode)
                 );
                 return false;
             }
+
+            if ($this->hmacSignature->isHmacSupportedEventCode($response)) {
+                //Validate the Hmac calculation
+                if (!$this->hmacSignature->isValidNotificationHMAC($this->configHelper->getNotificationsHmacKey(),
```
rikterbeek, line 219: so if hmac key is not set then we should not valdiate the hmac signature.

Helper/Config.php
```diff
@@ -80,6 +92,42 @@ public function getNotificationsIpHmacCheck($storeId = null)
...
+    /**
+     * Returns true if the enviroment is set to test, false for live
+     *
+     * @param int $storeId
+     * @return bool
+     */
+    public function isDemoMode($storeId = null)
+    {
+        return $this->getConfigData(self::XML_DEMO_MODE, self::XML_ADYEN_ABSTRACT_PREFIX, $storeId, true);
+    }
```
rikterbeek, line 129: we should have this check already in the util class let's not duplicate this.

etc/adminhtml/system/adyen_security.xml
```diff
@@ -38,5 +38,29 @@
...
+            <backend_model>Magento\Config\Model\Config\Backend\Encrypted</backend_model>
+            <config_path>payment/adyen_abstract/notification_hmac_key_live</config_path>
+            <tooltip>
+                <![CDATA[
+                Set HMAC key at first on Adyen so new notifications are sent with the key. To learn more about these settings refer to
+                <a target="_blank" href="https://docs.adyen.com/development-resources/webhooks/verify-hmac-signatures">Adyen documentation</a>.
+                ]]>
+            </tooltip>
+        </field>
+    </group>v
```
acampos1916, line 65:
```suggestion
    </group>
```

Controller/Process/Json.php
rikterbeek: @AlexandrosMor still valid concern can you fix this ?

rikterbeek: ok ignore my concern. So if the NotificationsIpHmacCheck is enabled it should valdiate the HMAC at all times. The check if this is set is above so this is solved.

Issue #678, opened 2020-03-27
[PW-2383] Unique constraint violation with duplicate vault token


danslo:
**Bug Description**

The Magento Vault has two unique bits of information:

- The `public_hash`. In Adyen's case, a hash of a concatenated string containing these parts:
  - Customer ID
  - Payment Method Code
  - JSON encoded array of payment details (type / masked cc / expiration date)
- The `gateway_token`. As the name suggests, this is simply the token returned by the gateway.

In the database, this is enforced using UNIQUE KEYs:

```
CREATE TABLE `vault_payment_token` (
...
  UNIQUE KEY `VAULT_PAYMENT_TOKEN_PUBLIC_HASH` (`public_hash`),
  UNIQUE KEY `VAULT_PAYMENT_TOKEN_PAYMENT_METHOD_CODE_CSTR_ID_GATEWAY_TOKEN` (`payment_method_code`,`customer_id`,`gateway_token`)
)
```

However, in the [code](https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Vault/Model/PaymentTokenManagement.php#L215-L233), only the `public_hash` is being used to check for duplicate tokens.

There might arise a situation where Adyen will send the same `gateway_token`, but the `public_hash` is different. An example of this is when the expiration date is updated. Adyen's [documentation](https://docs.adyen.com/checkout/tokenization/managing-tokens) says:

>Your shoppers can request changes to their saved payments details. The details associated with a token are card number, expiry month and year of the card, and the cardholder name. You can update the expiryMonth, expiryYear, and holderName of a card. **The token remains unchanged when you update these details.**

**To Reproduce**

- Enable CC Vault in Adyen configuration.
- Create a new account, to ensure we don't have any existing tokens stored.
- Place an order with:
  - Card: 5555 4444 3333 1111
  - Expiration: 10/20
  - CVC: 737
- Attempt to place another order with:
  - Card: 5555 4444 3333 1111
  - Expiration: 03/30
  - CVC: 737
- You should now be presented with a "Unique Constraint Violation Found" error. 

The public hash is different, but the gateway token remains the same.

**Expected behavior**

No "Unique Constraint Violation Found" error is shown.

**Additional context**

We have currently fixed this using a plugin on `Magento\Vault\Api\PaymentTokenManagementInterface`:

```php
public function aroundSaveTokenWithPaymentLink(
    PaymentTokenManagementInterface $paymentTokenManagement,
    callable $proceed,
[...]

PR #718, opened 2020-05-26
[PW-2383] Fix vault storing for updated expiration date

cyattilakiss:
**Description**

If shopper tries using credit cards with the same number but different expiration dates and save it to vault, they're faced with an error message saying "Unique constraint violation found".

**Tested scenarios**
Saving CC to vault: 5555444433331111 10/20 737
Resaving CC to vault: 5555444433331111 03/30 737

**Fixed issue**:  
#678 Unique constraint violation with duplicate vault token


acampos1916:
@cyattilakiss nice solution. What scenarios did you test?

PR #719, opened 2020-05-26
[PW-2286]: Avoiding order cancellation in Result controller

acampos1916:
**Description**
For payment methods that return to adyen/process/result we want to avoid order cancellation if the payment is unsuccessful. Instead, the OFFER_CLOSED notification should be responsible for cancelling the order.

This also means that [the OFFER_CLOSED notification must be enabled in Adyen's Customer Area.](https://support.adyen.com/hc/en-us/articles/360002149510-Why-do-I-see-an-open-order-in-Magento-but-not-in-my-Adyen-Customer-Area-) 

**Tested scenarios**
Refused or cancelled payment with iDeal does not cancel the order.

**Fixed issue**:  PW-2286

### DIFF
--- a/Controller/Process/Result.php
+++ b/Controller/Process/Result.php
@@ -107,7 +107,10 @@ class Result extends \Magento\Framework\App\Action\Action
                 $session->getQuote()->setIsActive(false)->save();
                 $this->_redirect('checkout/onepage/success', ['_query' => ['utm_nooverride' => '1']]);
             } else {
-                $this->_cancel($response);
+                $this->_adyenLogger->addAdyenResult(
+                    __('Payment for order %1 was unsuccessful, it will be cancelled when the OFFER_CLOSED notification has been processed.', $this->_order->getIncrementId())
+                );
+                $this->restoreCart($response);
                 $failReturnPath = $this->_adyenHelper->getAdyenAbstractConfigData('return_path');
                 $this->_redirect($failReturnPath);
             }
@@ -121,16 +124,12 @@ class Result extends \Magento\Framework\App\Action\Action
     /**
      * @param $response
      */
-    protected function _cancel($response)
+    protected function restoreCart($response)
     {
         $session = $this->_session;
 
         // restore the quote
         $session->restoreQuote();
-
-        $order = $this->_order;
-
-        $this->_adyenHelper->cancelOrder($order);
 
         if (isset($response['authResult']) && $response['authResult'] == \Adyen\Payment\Model\Notification::CANCELLED) {
             $this->messageManager->addError(__('You have cancelled the order. Please try again'));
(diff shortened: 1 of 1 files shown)


Controller/Process/Result.php
```diff
@@ -270,17 +269,17 @@ protected function _validateUpdateOrder($order, $response)
...
                 break;
             case Notification::CANCELLED:
-                $this->_adyenLogger->addAdyenResult('Cancel or Hold the order');
+                $this->_adyenLogger->addAdyenResult('Cancel or Hold the order on OFFER_CLOSED notification');
                 $result = false;
                 break;
             case Notification::REFUSED:
                 // if refused there will be a AUTHORIZATION : FALSE notification send only exception is idea
-                $this->_adyenLogger->addAdyenResult('Cancel or Hold the order');
+                $this->_adyenLogger->addAdyenResult('Cancel or Hold the order on OFFER_CLOSED notification');
```
cyattilakiss, line 277: The AUTHORISATION notification success = false is going to cancel it

acampos1916: Fixing it
```suggestion
                $this->_adyenLogger->addAdyenResult('Cancel or Hold the order on AUTHORISATION success=false notification');
```

cyattilakiss:
Also please see the sonarcloud warnings!

Controller/Process/Result.php
```diff
@@ -107,7 +107,10 @@ public function execute()
                 $session->getQuote()->setIsActive(false)->save();
                 $this->_redirect('checkout/onepage/success', ['_query' => ['utm_nooverride' => '1']]);
             } else {
-                $this->_cancel($response);
+                $this->_adyenLogger->addAdyenResult(
+                    __('Payment for order %1 was unsuccessful, it will be cancelled when the OFFER_CLOSED notification has been processed.', $this->_order->getIncrementId())
```
msilvagarcia, line 111:
```suggestion
                    sprintf(
                        'Payment for order %d was unsuccessful, ' .
                        'it will be cancelled when the OFFER_CLOSED notification has been processed.',
                        $this->_order->getIncrementId()
                    )
```

acampos1916: @msilvagarcia good point. I'll change the integer specifier %d for a string %s since order numbers can have leading zeros.

Controller/Process/Result.php
```diff
@@ -107,7 +107,14 @@ public function execute()
                 $session->getQuote()->setIsActive(false)->save();
                 $this->_redirect('checkout/onepage/success', ['_query' => ['utm_nooverride' => '1']]);
             } else {
-                $this->_cancel($response);
+                $this->_adyenLogger->addAdyenResult(
+                    sprintf(
+                        'Payment for order %d was unsuccessful, ' .
```
acampos1916, line 112:
```suggestion
                        'Payment for order %s was unsuccessful, ' .
```

PR #734, opened 2020-06-10
[PW-2329]: Installments converter for generic component

acampos1916:
**Description**
The generic component uses a special format to display installments. Here we're introducing a converter to transform the old array into this format.

**Tested scenarios**
Installments configuration is converted to the required format.

**Fixed issue**:  PW-2329


etc/adminhtml/system/adyen_cc.xml
```diff
@@ -77,17 +77,20 @@
...
                 </depends>
-                <validate>required-entry</validate>
-                <comment><![CDATA[
-                    Generate the desired installments configuration <a href="">here</a> and paste it on this field.
-                ]]></comment>
-                <config_path>payment/adyen_cc/installments_code</config_path>
+                <tooltip>Configure your installment for each credit card type: Insert the minimum amount required to
+                    make the configured installment available in the amount range column.
+                    Example: if the amount range is configured to 100 and the number of installments to 4x, the shopper
+                    will see the 4x option only if the payment total is higher or equal than 100.
```
Aleffio, line 89: Just wondering if the tooltip text is too long, maybe an idea to split it up between a comment and a tooltip?
(it can also stay like this)
```suggestion
                <comment>Configure your installment for each credit card type: Insert the minimum amount required to 
                make the configured installment available in the amount range column.
                </comment>
                <tooltip>
                    Example: if the amount range is configured to 100 and the number of installments to 4x, the shopper
                    will see the 4x option only if the payment total is higher or equal than 100.
```

acampos1916: Nice!

Helper/Installments.php
```diff
@@ -0,0 +1,77 @@
...
+ * See LICENSE.txt for license details.
+ *
+ * Author: Adyen <magento@adyen.com>
+ */
+
+namespace Adyen\Payment\Helper;
+
+class Installments
+{
+    static $card_types = array(
```
rikterbeek, line 28: Can you explain what the value means ? This code needs some more comments to make it clear what it does.

Helper/Installments.php
```diff
@@ -0,0 +1,77 @@
...
+        'HIPERCARD' => 4,
+        'MC' => 8,
+        'VI' => 16
+    );
+
+    /**
+     *
+     * @param array $installmentsArray
+     * @return string
+     *
```
rikterbeek, line 41: please describe the purpose of this method as this is a very complex one.

etc/adminhtml/system/adyen_cc.xml
```diff
@@ -77,17 +77,22 @@
...
+                <comment>Configure your installment for each credit card type: Insert the minimum amount required to 
+                make the configured installment available in the amount range column.
+                </comment>
+                <tooltip>
+                    Example: if the amount range is configured to 100 and the number of installments to 4x, the shopper
+                    will see the 4x option only if the payment total is higher or equal than 100.
+                </tooltip>
+                <frontend_model>Adyen\Payment\Block\Adminhtml\System\Config\Field\Installments</frontend_model>
+                <backend_model>Adyen\Payment\Model\Config\Backend\Installments</backend_model>
+                <config_path>payment/adyen_cc/installments</config_path>
```
rikterbeek, line 95: we are changing the config_path to installments instead of installments_code. Can't we be backwards compatible ?

Helper/Installments.php
acampos1916: Added more info to the doc block to explain the purpose.

Helper/Installments.php
acampos1916: Added the description to the variable doc block. This is a bit indicator indicator used in the generic component to filter the supported card types.

etc/adminhtml/system/adyen_cc.xml
acampos1916: We had changed it to installments_code in this feature branch to enter the format required by the generic component manually. But then decided to implement a converter instead and restored the original config_path. So now it is backwards compatible as it is unchanged.

PR #740, opened 2020-06-16
Filter out children from invoice item collection

sbothnermzentrale:
**Description**
Different number of line items for authorization and capture request body leads to an error with payment method "Rate Pay". 

2300: Request basket not valid Basket error: ITEM "Product Name" Product Name: not found in history (150)

**Tested scenarios**
n.a.

**Fixed issue**: 
Authorization/CheckoutDataBuilder (https://github.com/Adyen/adyen-magento2/blob/develop/Gateway/Request/CheckoutDataBuilder.php#L217) filters out all children of quote item by default. This means that there are different amounts of data in the lines for authorization and capture requests.


Gateway/Request/CaptureDataBuilder.php
```diff
@@ -102,7 +102,8 @@ protected function getOpenInvoiceData($payment)
         // The latest invoice will contain only the selected items(and quantities) for the (partial) capture
         $latestInvoice = $invoices->getLastItem();
 
-        foreach ($latestInvoice->getItems() as $invoiceItem) {
+        foreach ($latestInvoice->getItems() as $invoiceItem) {            
+            if($invoiceItem->getOrderItem()->getParentItem()) continue;
```
acampos1916, line 106:
```suggestion
            if ($invoiceItem->getOrderItem()->getParentItem()) {
                continue;
            }
```

acampos1916: Adjustment to pass SonarCloud Quality Gate

acampos1916:
@sbothnermzentrale thank you for this contribution.

If I understand correctly you are proposing that the items with a parent are skipped (`if($invoiceItem->getOrderItem()->getParentItem())`) when getting the invoice lines for capture because they are not looped through when doing the authorization request (`getAllVisibleItems()`). I think this makes sense.

cyattilakiss:
Hi @sbothnermzentrale

Thank you for your contribution, we are going to release this fix in our next release.

Best,
Attila
Adyen

PR #743, opened 2020-06-18
Remove Apple Pay extra check if valid active cards are present

acampos1916:
**Description**
This check always returned false even if active cards were present in the wallet. [A similar fix was applied for the Magento 1 plugin](https://github.com/Adyen/adyen-magento/pull/781/commits/ea832c4458edc13af8f07276c0f81b0664b4315e).

**Tested scenarios**
Safari in regular and private mode shows Apple Pay button.


rikterbeek:
@acampos1916,

I have seen this as well being false many times but weird as apple has this into there public docs:
https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession/1778000-canmakepaymentswithactivecard

Regards,
Rik
Adyen

Issue #726, opened 2020-06-04
[PW-2475]: Adyen Klarna "invoice on shipment" does not capture funds


MikeParkin:
**Describe the bug**
When the config option for "payment/adyen_abstract/capture_on_shipment" is enabled, the user would expect that an order is both invoiced in Magento and the payment is captured in Adyen when creating a shipment. 

**To Reproduce**
1. Enable payment/adyen_abstract/capture_on_shipment
2. Create a Klarna purchase
3. Ship the order (without invoicing first)

**Expected behavior**
A shipment and invoice should be generated, with the payment captured in Adyen.

**Actual result**
The shipment and invoice are created, but the funds are not captured.

**Cause**
The invoice object needs to be told to capture (this flag is set when invoicing the order in the admin panel by Magento, but it is not set in your observer). 

Additionally, the error message can never be seen because there is a missing % in the pattern.

```
diff --git a/Observer/Adminhtml/BeforeShipmentObserver.php b/Observer/Adminhtml/BeforeShipmentObserver.php
index aa86468..1957a10 100755
--- a/Observer/Adminhtml/BeforeShipmentObserver.php
+++ b/Observer/Adminhtml/BeforeShipmentObserver.php
@@ -26,6 +26,7 @@ namespace Adyen\Payment\Observer\Adminhtml;
 
 use Magento\Framework\Event\Observer;
 use Magento\Payment\Observer\AbstractDataAssignObserver;
+use Magento\Sales\Model\Order\Invoice;
 
 /**
  * Class DataAssignObserver
@@ -71,10 +72,11 @@ class BeforeShipmentObserver extends AbstractDataAssignObserver
                         // set transaction id so you can do a online refund from credit memo
                         $pspReference = $order->getPayment()->getAdyenPspReference();
                         $invoice->setTransactionId($pspReference);
+                        $invoice->setRequestedCaptureCase(Invoice::CAPTURE_ONLINE);
                         $invoice->register()->pay();
                         $invoice->save();
                     } catch (\Exception $e) {
-                        throw new \Exception(sprintf('Error saving invoice. The error message is:', $e->getMessage()));
+                        throw new \Exception(sprintf('Error saving invoice. The error message is: %s', $e->getMessage()));
                     }
                 }
             }
```

Furthermore, you have only added the observer to the Adminhtml namespace, this means that any cron jobs that run that create a shipment (for example, automatically importing shipments from an ERP) never create the invoice or capture the payment.

This can be added like this: 

```
[...]

PR #755, opened 2020-07-02
Fix issue with capture on shipment

msilvagarcia:
**Description**
When setting the plugin to capture on shipment, the capture is not actually happening. This is because the invoice is not configured on how to manage the capture process (online or offline). This commit provides the invoice object with information about the capture mode.

**Tested scenarios**
1. Set capture to manual and capture on shipment on for OpenInvoice payment methods in the plugin.
1. Set capture to manual in Customer Area.
1. Make a purchase and pay with Klarna pay later.
1. In Magento 2 admin, submit a shipment for this order.


floorjubbega:
Fixes #726

Issue #756, opened 2020-07-02
PaymentMethod::getCurrentPaymentAmount has a wrong behavior


rvitaliy:
https://github.com/Adyen/adyen-magento2/blob/c19023412c8bee916d2e71581dd17d6c292ee5d0/Helper/PaymentMethods.php#L258

**Expected behavior**
this method must return a float or throw exception because getGrandTotal is invalid for execute a payment with adyen

cyattilakiss:
Hi @rvitaliy ,

Thank you for raising this issue and providing a PR to fix that, we really appreciate it.
Could you please provide us a test case when this issue occurs so we can test it against the current implementation? 

Thank you in advance.
Best,
Attila
Adyen

rvitaliy:
@cyattilakiss 
Yes sure, to test this code you must reproduce an cart with grandTotal <= 0.0.
You can try to create a product with price at 0 and add it to your cart.


I think that this code is self explanatory that something is wrong in this:
```
    /**
     * @return bool|int
     */
    protected function getCurrentPaymentAmount()
    {
        if (($grandTotal = $this->getQuote()->getGrandTotal()) > 0) {
            return $grandTotal;
        }
        return 10;
    }
```

why default return is 10?
phpdoc declare return bool|int but getGrandTotal declare return a float (it's not true magento return numeric string) in my PR i manage all possible type returned from getGrandTotal we are ready to receive int | float | string(numeric) and validate if value is > 0.0 in other case we throw exception.

AlexandrosMor:
Hello @rvitaliy,

We merged your PR and it is going to be included in the next magento 2 release. 

kind regards,
Alexandros

PR #757, opened 2020-07-02
refs #756: fix PaymentMethod::getCurrentPaymentAmount

rvitaliy:
**Fixed issue**:  refs #756


rvitaliy:
Travis CI fails, but it's not related to changes.
composer can't download Magento because authentication fails

cyattilakiss:
Hi @rvitaliy ,

Thank you for opening this PR we are going to review this asap :)

Best,
Attila
Adyen

Helper/PaymentMethods.php
```diff
@@ -252,15 +252,32 @@ protected function fetchAlternativeMethods($country)
         return $paymentMethods;
     }
 
-    /**
-     * @return bool|int
-     */
-    protected function getCurrentPaymentAmount()
+    protected function getCurrentPaymentAmount(): float
```
acampos1916, line 255: Supporting PHP<7
```suggestion
    protected function getCurrentPaymentAmount()
```

Helper/PaymentMethods.php
```diff
@@ -252,15 +252,32 @@ protected function fetchAlternativeMethods($country)
...
     {
-        if (($grandTotal = $this->getQuote()->getGrandTotal()) > 0) {
+        $grandTotal = $this->getQuote()->getGrandTotal();
+
+        if (!\is_numeric($grandTotal)) {
```
acampos1916, line 259: I think there's no need to specify the global namespace for these functions?
```suggestion
        if (!is_numeric($grandTotal)) {
```

Helper/PaymentMethods.php
```diff
@@ -252,15 +252,32 @@ protected function fetchAlternativeMethods($country)
...
+            throw new \Exception(
+                \sprintf(
```
acampos1916, line 261: I think there's no need to specify the global namespace for these functions?
```suggestion
                sprintf(
```

Helper/PaymentMethods.php
```diff
@@ -252,15 +252,32 @@ protected function fetchAlternativeMethods($country)
...
+
+        $grandTotal = (float) $grandTotal;
+
+        if ($grandTotal > 0.0) {
             return $grandTotal;
         }
-        return 10;
+
+        throw new \Exception(
+            \sprintf(
```
acampos1916, line 275: I think there's no need to specify the global namespace for these functions?
```suggestion
            sprintf(
```

rvitaliy:
Hi @cyattilakiss!
I seen your last 2 commits.
The result is same, but if we work with **float** type we must compare with **float** and not with int.
PHP allow us to do it, but it's theoretically wrong :) 

It's just a note to keep in mind.
If you prefer we can keep these changes.

cyattilakiss:
> Hi @cyattilakiss!
> I seen your last 2 commits.
> The result is same, but if we work with **float** type we must compare with **float** and not with int.
> PHP allow us to do it, but it's theoretically wrong :)
> 
> It's just a note to keep in mind.
> If you prefer we can keep these changes.

Hey @rvitaliy 
Thank you for the response. This time I prefer to keep it and not cast the zero to float for only the comparison. (PHP will do it anyway in the background)
Thanks and now we are going to merge it finally :)
Best,
Attila

PR #770, opened 2020-07-21
[PW-2087] Add Swish Payment Method and upgrade Component version to 3.4.0 This commit has the following changes:

rikterbeek:
* With the added environment into the component is does not mather if we use the test or live component url. To remove complexity we will use the live endpoint on both environments.
* Upgrade component version from 3.2.0 to 3.4.0. This requires a different way of loading in the component. This will change all payment method integration that is using checkout component.
* Add swish payment method support with support for pending payment results and redering a QR code.
* Environment and originKey is now as well needed for alternative payment methods so moved this into a genericConfigProvider.
* The threeDS2processor is being used for 3DS2 and action payment methods. TODO we need to rename this to a generic name into the code this will mean a breaking change as interface will rename and PWA needs to adopt different API endpoint as this is a quick feature release I kept it out of scope.
* Rename threeDS2PaymentData to generic adyenPaymentData as this will be used for alternative payent methods as well.
* Remove unused ThreeDS2ResponseValidator file

**Description**

**Tested scenarios**

**Fixed issue**:


view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
```diff
@@ -257,30 +259,14 @@ define(
                             }
                             var oneClickCardNode = document.getElementById('cvcContainer-' + self.value);
 
-                            // this should be fixed in new version of checkout card component
-                            var hideCVC = false;
```
Aleffio, line 261: don't we still need this for RECURRING billing agreements?

Helper/Data.php
```diff
@@ -34,6 +34,7 @@ class Data extends AbstractHelper
     const MODULE_NAME = 'adyen-magento2';
     const TEST = 'test';
     const LIVE = 'live';
+    // Only used for backend orders! Checkout in front-end is using different checkout version see web folder
```
cyattilakiss, line 37: thanks for the comment so it won't confuse support or us

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -73,6 +77,8 @@ define(
                 return this;
             }, initialize: function () {
 
+                this.messageContainer
```
cyattilakiss, line 80: Shouldn't this be removed?
```suggestion
```

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -296,7 +301,7 @@ define(
              * @returns {{method: *}}
              */
             getData: function () {
-                const browserInfo = threeDS2Utils.getBrowserInfo();
+                var browserInfo = threeDS2Utils.getBrowserInfo();
```
msilvagarcia, line 304: nice catch!

view/frontend/requirejs-config.js
```diff
@@ -11,5 +11,10 @@ var config = {
                 'Magento_CheckoutAgreements/js/model/place-order-mixin': true
             }
         }
+    },
+    map: {
+        '*': {
+            'adyenCheckout':  'https://checkoutshopper-live.adyen.com/checkoutshopper/sdk/3.4.0/adyen.js'
```
Aleffio, line 17: does it matter that we always use the `-live` endpoint, instead of using the proper one? (`-test` or `-live`)

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -639,9 +646,94 @@ define(
...
+
+                self.popupModal.modal("openModal");
+                self.actionComponent = self.checkoutComponent.createFromAction(action).mount(actionNode);
+            },
+            handleOnAdditionalDetails: function (state, component) {
+
+                var self = this;
+
+                // call endpoint with state.data
+                debugger;
```
cyattilakiss, line 706:
```suggestion
```

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
rikterbeek: yes can be removed.

view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
rikterbeek: Thanks will investigate if the hideCVC still is possible in the new component.

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -639,9 +646,94 @@ define(
...
+                var request = state.data;
+                request.orderId = self.orderId;
+
+                threeds2.processThreeDS2(request).done(function () {
```
cyattilakiss, line 710: let's add a comment for future reference that now the threeDS2Processor handles the actions as well so we won't forget that this is already here when we refactor this

view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
cyattilakiss: the component now supposed to handle it but good to verify it on the e2e instance when uploaded.

rikterbeek: valid point will revert this piece of code.

rikterbeek: yeah indeed wondering how this works now.

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
rikterbeek: sure can do.

Model/AdyenThreeDS2Process.php
```diff
@@ -92,21 +92,23 @@ public function initiate($payload)
...
 
         // Init payments/details request
         $result = [];
 
-        if ($paymentData = $payment->getAdditionalInformation("threeDS2PaymentData")) {
+        if ($paymentData = $payment->getAdditionalInformation("adyenPaymentData")) {
             // Add payment data into the request object
             $request = [
-                "paymentData" => $payment->getAdditionalInformation("threeDS2PaymentData")
+                "paymentData" => $payment->getAdditionalInformation("adyenPaymentData")
```
cyattilakiss, line 107: The data is already retrieved into the $paymentData var
```suggestion
                "paymentData" => $paymentData
```

view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
rikterbeek: @cyattilakiss tests are successful, are we testing this setup though with recurring contracts only ?

rikterbeek: I could revert it to: 
 if (!this.hasVerification()) {
 hideCVC = true;
}
just to cover this scenario. The maestro fix here is not needed anymore that for sure.
@Aleffio ok with this approach ?

Aleffio: 👍

cyattilakiss: @rikterbeek we only test with one click - that should be the default setup for the e2e instance

view/frontend/web/js/view/payment/adyen-methods.js
```diff
@@ -67,23 +67,13 @@ define(
                 var self = this;
```
cyattilakiss, line 67: Not used can be removed as well

```suggestion
```

view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
rikterbeek: done.

PR #771, opened 2020-07-22
fix payment method POS

rikterbeek:
**Description**

**Tested scenarios**

**Fixed issue**:


Helper/PaymentMethods.php
```diff
@@ -413,5 +413,6 @@ public function getConnectedTerminals()
         }
 
         return $responseData;
+        return [];
```
msilvagarcia, line 416:
```suggestion
```

This line will never be reached

rikterbeek: True thanks.

PR #775, opened 2020-07-24
[PW-2390] Add action component on success page if presentToShopper an…

rikterbeek:
…d Action result is provided.

* Added a adyenCheckout3101 requireJS variable to use latest component version. This we can use for all new payment method integrations or upgrades.
* Success block loads all values needed for the component
* Success page renders the component
* Layout renders the styles.css file for showing the component with the right styling

**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Block/Checkout/Success.php
+++ b/Block/Checkout/Success.php
@@ -32,22 +32,28 @@ class Success extends \Magento\Framework\View\Element\Template
     /**
      * @var \Magento\Sales\Model\Order $order
      */
-    protected $_order;
+    protected $order;
 
     /**
      * @var \Magento\Checkout\Model\Session
      */
-    protected $_checkoutSession;
+    protected $checkoutSession;
 
     /**
      * @var \Magento\Checkout\Model\OrderFactory
      */
-    protected $_orderFactory;
+    protected $orderFactory;
+
 
     /**
-     * @var \Magento\Framework\Pricing\Helper\Data
+     * @var \Adyen\Payment\Helper\Data
      */
-    public $priceHelper;
+    protected $adyenHelper;
+
+    /**
+     * @var \Magento\Store\Model\StoreManagerInterface
+     */
+    protected $storeManager;
 
     /**
      * Success constructor.
@@ -63,11 +69,15 @@ class Success extends \Magento\Framework\View\Element\Template
         \Magento\Checkout\Model\Session $checkoutSession,
         \Magento\Sales\Model\OrderFactory $orderFactory,
         \Magento\Framework\Pricing\Helper\Data $priceHelper,
+        \Adyen\Payment\Helper\Data $adyenHelper,
+        \Magento\Store\Model\StoreManagerInterface $storeManager,
         array $data = []
     ) {
-        $this->_checkoutSession = $checkoutSession;
-        $this->_orderFactory = $orderFactory;
+        $this->checkoutSession = $checkoutSession;
+        $this->orderFactory = $orderFactory;
         $this->priceHelper = $priceHelper;
+        $this->adyenHelper = $adyenHelper;
+        $this->storeManager = $storeManager;
         parent::__construct($context, $data);
     }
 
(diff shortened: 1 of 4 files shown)


view/frontend/templates/checkout/success.phtml
```diff
@@ -25,7 +25,32 @@
...
+        });
+    </script>
+
+
+    <div id="ActionContainer"></div>
+
+
+<?php else: ?>
+
+    <?php if ($block->isBoletoPayment()): ?>
```
acampos1916, line 53: For phpcs
```suggestion
    </script>
    <div id="ActionContainer"></div>
<?php else: ?>
    <?php if ($block->isBoletoPayment()): ?>
```

rikterbeek: yes thanks will fix it

view/frontend/templates/checkout/success.phtml
```diff
@@ -25,7 +25,32 @@
  * @var \Adyen\Payment\Block\Checkout\Success $block
  */
 ?>
-<?php if ($block->isBoletoPayment()): ?>
+
+
+<?php if ($block->renderAction()): ?>
+    <script type="text/javascript">
```
cyattilakiss, line 31: is there a way to outsource it into a separate JS file?

view/frontend/templates/checkout/success.phtml
```diff
@@ -25,7 +25,32 @@
...
+        require([
+            'jquery',
+            'adyenCheckout3101'  // the alias for "mage/accordion"
+        ], function ($, AdyenCheckout) {
+            var action = JSON.parse('<?php echo $block->getAction(); ?>');
```
cyattilakiss, line 36: Magento automation check requires this format:

```suggestion
            var action = JSON.parse('<?= $block->escapeHtml($block->getAction()); ?>');
```

view/frontend/templates/checkout/success.phtml
```diff
@@ -25,7 +25,32 @@
...
+            var checkoutComponent = new AdyenCheckout({
+                locale: '<?php echo $block->getLocale(); ?>',
+                environment: '<?php echo $block->getEnvironment(); ?>',
+                originKey: '<?php echo $block->getOriginKey(); ?>'
```
cyattilakiss, line 40:
```suggestion
                locale: '<?= $block->escapeHtml($block->getLocale()); ?>',
                environment: '<?= $block->escapeHtml($block->getEnvironment()); ?>',
                originKey: '<?= $block->escapeHtml($block->getOriginKey()); ?>'
```

view/frontend/templates/checkout/success.phtml
```diff
@@ -25,7 +25,32 @@
...
+            });
+            var actionNode = document.getElementById('ActionContainer');
+            checkoutComponent.createFromAction(action).mount(actionNode);
```
cyattilakiss, line 43: Remove the extra actionNode, checkout component can handle it
Put the createFromAction into a try catch

```suggestion
            try {
                checkoutComponent.createFromAction(action).mount('#ActionContainer');
            } catch(err) {
                // Action component cannot be created
            }
```

view/frontend/templates/checkout/success.phtml
```diff
@@ -135,4 +162,5 @@
         <?php endif; ?>
         </tbody>
     </table>
-<?php endif; ?>
+    <?php endif; ?>
+<?php endif; ?>
```
cyattilakiss, line 166:
```suggestion
<?php endif; ?>

```

view/frontend/templates/checkout/success.phtml
rikterbeek: No as it needs the php tags for the vraiables.

view/frontend/templates/checkout/success.phtml
rikterbeek: @acampos1916 this is not correct it need to be in the if statement

view/frontend/templates/checkout/success.phtml
```diff
@@ -25,7 +25,28 @@
  * @var \Adyen\Payment\Block\Checkout\Success $block
  */
 ?>
-<?php if ($block->isBoletoPayment()): ?>
+<?php if ($block->renderAction()): ?>
+    <script type="text/javascript">
+        require([
+            'jquery',
+            'adyenCheckout3101'  // the alias for "mage/accordion"
```
acampos1916, line 32:
```suggestion
            'adyenCheckout3101'
```

view/frontend/templates/checkout/success.phtml
```diff
@@ -25,7 +25,28 @@
...
  */
 ?>
-<?php if ($block->isBoletoPayment()): ?>
+<?php if ($block->renderAction()): ?>
+    <script type="text/javascript">
+        require([
+            'jquery',
+            'adyenCheckout3101'
+        ], function ($, AdyenCheckout) {
+            var action = JSON.parse('<?= $block->getAction(); ?>');
```
cyattilakiss, line 34: I don't know if we can escape it or not so to be sure that we don't change the structure of the action I would just add this:

```suggestion
            var action = JSON.parse('<?= /* @noEscape */ $block->getAction(); ?>');
```

Block/Checkout/Success.php
```diff
@@ -151,14 +161,55 @@ public function getMultibancoData()
...
     }
 
+    /**
+     * If PresentToShopper resultCode and action has provided render this with the checkout component on the success page
+     * @return bool
+     */
+    public function renderAction()
+    {
+        if (
+            !empty($this->getOrder()->getPayment()->getAdditionalInformation('resultCode')) &&
```
msilvagarcia, line 171: Is this `!empty` check even needed if you're comparing to `PresentToShopper` later?

view/frontend/templates/checkout/success.phtml
msilvagarcia: Would be nice to add some checks on the backend as well, then.

rikterbeek: We can add this tag. @msilvagarcia not sure what you mean by add checks in the backend. As this comes from the API we just need to pass it to the component. The component will use it there will be no server call afterwards.

Block/Checkout/Success.php
rikterbeek: there is a potential that the resultCode does not exists, so that is why I added this check to prevent warnings.

Issue #384, opened 2019-01-23
setState function can cause order state to be set to NULL


yanielliot:
Plugin version: >=2.3.1

setState function added in 2.3.1 can result in order state being set to NULL. 

```
    protected function _setState($status)
    {
        $statusObject = $this->_orderStatusCollection->create()
            ->addFieldToFilter('main_table.status', $status)
            ->addFieldToFilter('state_table.is_default', true)
            ->joinStates()
            ->getFirstItem();

        $this->_order->setState($statusObject->getState());
        $this->_adyenLogger->addAdyenNotificationCronjob('State is changed to  ' . $statusObject->getState());
    }
```

Line:

```
->addFieldToFilter('state_table.is_default', true)
```

makes the assumption that the status set in Adyen module config is the default status for a least 1 state. If this is not true no state is returend and the function sets order to state NULL.

Additionally line:

```
->getFirstItem();
```

may result in order state being set differently than user intended as status can be default for multiple states so may not return desired/expected state.

Would recommend that either config status fields are replaced with all possible state/status combinations so user can explicitly set what state and status the order is set to at each step. Alternatively config status fields could be limited to statuses of specific order states that make sense for that step from viewpoint of standard Magento order flow (eg to me Manual Review Accept status would be one belonging to state "processing").

Either way feel order state should be known and passed into setState function along with order status, rather than the function trying to work out which state to use, as order state and status are regularly used as key indicators during order processing and 3rd party systems integrations.

ebaschiera:
I second this issue.
For my use case, it's enough to remove the following line:
`->addFieldToFilter('state_table.is_default', true)`

Thanks!

ebaschiera:
Hi Devs,
this bug is as small as easy to solve. Do you really need a PR for that? :)

The bug report is detailed and summarize deeply the issue.

Just try setting up a couple of statuses per state and then configure Adyen module to set one of the non-default statuses. It will set order status to NULL.

Thanks!

AlexandrosMor:
Hello  @ebaschiera,

It would be great if you can do a pull request so we can review it and merge the changed code. 

regards,
Alexandros

PR #778, opened 2020-07-30
avoid setting order state to NULL

ebaschiera:
**Description**
Avoid setting order state to NULL. This may happen when setting a status that is not default for a state.

**Tested scenarios**

**Fixed issue**:  Adyen/adyen-magento2#384

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -1945,7 +1945,6 @@ class Cron
     {
         $statusObject = $this->_orderStatusCollection->create()
             ->addFieldToFilter('main_table.status', $status)
-            ->addFieldToFilter('state_table.is_default', true)
             ->joinStates()
             ->getFirstItem();
 


rikterbeek:
Hi @ebaschiera,

Thanks for the PR. We do check the default because we want to use the default rather then use the first status of that state. I think the better approach is if this results into a null to have a fallback to the first status, are you able to modify your PR to match this behaviour ?

Regards,
Rik
Adyen

ebaschiera:
Hi @rikterbeek ,
I think you didn't get all the details of the issue... The query is built the other way round.
Let's start from the beginning. The code in question is the following:

```
    /**
     * Set State from Status
     */

    protected function _setState($status)
    {
        $statusObject = $this->_orderStatusCollection->create()
            ->addFieldToFilter('main_table.status', $status)
            ->addFieldToFilter('state_table.is_default', true)
            ->joinStates()
            ->getFirstItem();

        $this->_order->setState($statusObject->getState());
        $this->_adyenLogger->addAdyenNotificationCronjob('State is changed to  ' . $statusObject->getState());
    }
```

Let's say that I have a custom order status "confirmed" and its parent state is "processing".
The query on `$this->_orderStatusCollection` would be the following: 

```
SELECT `main_table`.*, `state_table`.`state`, `state_table`.`is_default`, `state_table`.`visible_on_front` FROM `sales_order_status` AS `main_table`
 LEFT JOIN `sales_order_status_state` AS `state_table` ON main_table.status=state_table.status WHERE (`main_table`.`status` = 'confirmed') AND (`state_table`.`is_default` = '1')
```

So if my status "confirmed" is not the default for its parent state, the query will return 0 results.
And the following line will set order state to NULL:

`$this->_order->setState($statusObject->getState());`

That's not correct, because my status "confirmed" has a parent state ("processing") that is perfectly suitable for this use case.

Thanks!

rikterbeek:
Hi @ebaschiera,

You are right thought it was about setting the status not state. Thanks for the detailed explanation. Looks good to me we will test and merge when it is reviewed by 2 or more devs.

Regards,
Rik
Adyen

Issue #780, opened 2020-07-31
IP address and HMAC signature check => Notification has been rejected because the IP address could not be verified


LittleBigDev:
**Describe the bug**
"Notification has been rejected because the IP address could not be verified". After investigation, this is because of IP validation. More details in Additional Context section

**To Reproduce**
Prerequisites:
In received notification request headers, `HTTP_X_FORWARDED_FOR` must look like `XX.XX.XX.XX, YY.YY.YY.YY`

1. In admin, enable "Check notification's IP address and HMAC signature"
2. In front, prepare an order to be placed
3. Monitor `var/log/adyen/notification.log` file with a `tail -f` for example
4. Enter credit card info and submit payment form

**Expected behavior**
Regular notification logs in log file

**Actual behavior**
```
[2020-07-31 09:09:03] AdyenLoggerTest.ADYEN_NOTIFICATION: Notification has been rejected because the IP address could not be verified {"is_exception":false} []
```

**Desktop (please complete the following information):**
 - Irrelevant

**Additional context**
Our test server is hosted on AWS EC2. I don't knwow why exactly we have HTTP_X_FORWARDED_FOR with concatenated IPs, but here we are : there are concatenated IPs in this header.

`\Adyen\Payment\Helper\IpAddress::isIpAddressValid()` method is unable to match any IP because passed IP values were not exploded in `\Adyen\Payment\Controller\Process\Json::isIpValid()`.
In `isIpValid()` method, IP addresses retrieved from `$_SERVER['HTTP_X_FORWARDED_FOR']` should be exploded on comma character (`,`), then trimmed, and only then pushed into `$ipAddress` array.

Thank you by advance !

AlexandrosMor:
Hello @LittleBigDev,

Thank for raising this issue. We are going to look into it and come back to you as soon as possible.

regards,
Alexandros

acampos1916:
Hey @LittleBigDev,

This has been addressed in #784, thanks for the feedback.

Ángel

LittleBigDev:
Thank you @acampos1916 

I left a comment there. This is a double bummer for us : your fix doesn't validate IPs from `HTTP_X_FORWARDED_FOR` anymore. And because of that (and our load balancer), we won't be able to enable either IP validation or HMAC key validation :(

Actually, any server architecture hiding real remote address to magento won't be compatible with your IP and HMAC key validation feature :/

PR #784, opened 2020-08-05
Remove X-forwarded-for header and IP whitelisting/ HMAC is in different configuration

AlexandrosMor:
**Description**
PW-2750 Currently the ip whitelist is only enabled when the HMAC check is turned on, it would be nice to enable it by default.
The IP whitelist is only enabled by default without depending on HMAC configuration.

[DOCS] A new config field has been introduced to the Security section.

PW-2749 X-forwarded-for header can be controlled by an attacker and used to bypass the whitelist protection
Remove X-forwarded-for header

[DOCS] For the IP verification to check the user's IP the webserver/proxy should be configured to plug the client IP to the REMOTE_ADDR header.

**Tested scenarios**

**Fixed issue**:  #780

### DIFF
--- a/Controller/Process/Json.php
+++ b/Controller/Process/Json.php
@@ -351,12 +351,10 @@ class Json extends \Magento\Framework\App\Action\Action
     protected function isIpValid()
     {
         $ipAddress = [];
-
         //Getting remote and possibly forwarded IP addresses
         if (!empty($_SERVER['REMOTE_ADDR'])) {
-            array_push($ipAddress, $_SERVER['REMOTE_ADDR']);
+            $ipAddress = explode(',', $_SERVER['REMOTE_ADDR']);
         }
-
         return $this->ipAddressHelper->isIpAddressValid($ipAddress);
     }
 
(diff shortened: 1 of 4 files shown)


LittleBigDev:
Hello, 

I reviewed this fix and I can see you removed `HTTP_X_FORWARDED_FOR` from IP validation for security reasons. I'm ok with that.

With our load balancer, the real incoming IP is stored in `HTTP_X_FORWARDED_FOR`. So we can't enable IP verif (no Adyen notif will be validated because `$_SERVER['REMOTE_ADDR']` will be 127.0.0.1, which is not a valid IP).
So, we have to disable IP validation.

But I'm sad to figure out that hmac key validation is tightly coupled to IP verif. If IP verif is disabled, no hmac key verif will happen.
So, because of our load balancer, we are not able to check hmac key either.

This is a double bummer :(

acampos1916:
Hey @LittleBigDev thanks for this feedback.

I can check from a security POV if we can decouple both checks.

But `HTTP_X_FORWARDED_FOR` has been removed because the client could spoof any desired value in that header, in your case I understand that you trust the information passed by your proxy but this is not common for every implementation.

`REMOTE_ADDR` should, as its name implies, contain the remote address. So maybe it's better to wire the correct IP in the proxy to that header.

LittleBigDev:
Thank you, we will deal with fixing our REMOTE_ADDR.

And if you can decouple both checks, I will be happy :)

acampos1916:
Hey @LittleBigDev, the checks are being decoupled in this PR https://github.com/Adyen/adyen-magento2/pull/809

Issue #785, opened 2020-08-05
ChallengeRequest in PWA that use M2 as headless eCommerce


Fifciu:
**Describe the bug**
I am using Vue Storefront 1 with Magento2 as backend. I am working on Adyen module for VSF that works with M2. Everything goes pretty fine expect ChallengeRequest. After click on the button in challenge I am getting blank page inside iframe. I checked `3dnotif.shtml` inside my Network Tab and here we have code like:
```js
var transStatus = 'Y';
var threeDSServerTransID = '17413d77-b3dc-45d2-8682-6e156b0b7e44';
var type = 'challengeResult';
var parentOrigin = 'https://my-m2-url.com';
var data = {
      result: {
           transStatus: transStatus
      },
      threeDSServerTransID: threeDSServerTransID,
      type: type
}
var result = JSON.stringify(data);
window.parent.postMessage(result, parentOrigin);
```

The problem is - PWA URL is different than M2 URL. To solve this problem I had to change:
```php
$request['origin'] = $this->adyenHelper->getOrigin();
```
to:
```php
$request['origin'] = $payload[PaymentInterface::KEY_ADDITIONAL_DATA]['origin'];
```
Inside `Helper/Requests.php` -> `buildThreeDS2Data`.

Then I am just sending a proper new origin from Secure Proxy API which is bridge between my PWA & M2.
This is pretty tricky solution. I would love to see much more user friendly one - e.g. setting Frontend origin via Magento's dashboard. 

**Expected behavior**
Possible to set Frontend origin via M2 Dashboard.
It could look like:
```
$request['origin'] = $originSetInMagentoAdminPanel || $this->adyenHelper->getOrigin();
```

PR #786, opened 2020-08-05
add setting for payment origin (for pwa integrations)

pablogiralt:
**Description**
To integrate Magento Adyen extension with PWA app (In our case Vue Storefront) we need to be able to set origin as the pwa's url instead of magento's when doing the ChallengeRequest for 3DS2 verification.

**Current Behaviour**
After click on the button in challenge I am getting blank page inside iframe. I checked 3dnotif.shtml inside my Network Tab and here we have code like:

```
var transStatus = 'Y';
var threeDSServerTransID = '17413d77-b3dc-45d2-8682-6e156b0b7e44';
var type = 'challengeResult';
var parentOrigin = 'https://my-m2-url.com';
var data = {
      result: {
           transStatus: transStatus
      },
      threeDSServerTransID: threeDSServerTransID,
      type: type
}
var result = JSON.stringify(data);
window.parent.postMessage(result, parentOrigin);
```

**Proposed solution**
I created a setting in magento adyen's setting to add pwa url. 
If this url is set, then it is used as origin in ChallengeRequest

**Tested scenarios**

**Fixed issue**:  #785

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1536,6 +1536,7 @@ class Data extends AbstractHelper
     }
 
     /**
+     * @param null|int|string $storeId
      * @return string
      */
     public function getOrigin($storeId)
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1538,7 +1538,7 @@ class Data extends AbstractHelper
     /**
      * @return string
      */
-    public function getOrigin()
+    public function getOrigin($storeId)
     {
         if ( $paymentOriginUrl = $this->getAdyenAbstractConfigData("payment_origin_url", $storeId) ) {
             return $paymentOriginUrl;
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1539,6 +1539,9 @@ class Data extends AbstractHelper
      */
     public function getOrigin()
     {
+        if ( $paymentOriginUrl = $this->getAdyenAbstractConfigData("payment_origin_url", $storeId) ) {
+            return $paymentOriginUrl;
+        }
         $objectManager = \Magento\Framework\App\ObjectManager::getInstance();
         $state = $objectManager->get(\Magento\Framework\App\State::class);
         $baseUrl = $this->storeManager->getStore()->getBaseUrl(\Magento\Framework\UrlInterface::URL_TYPE_WEB);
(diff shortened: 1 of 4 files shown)


Helper/Data.php
```diff
@@ -1539,6 +1539,9 @@ private function createAdyenClient()
      */
     public function getOrigin()
```
cyattilakiss, line 1540: The parameter is missing - phpdoc should also be updated

```suggestion
    public function getOrigin($storeId)
```

cyattilakiss:
Hi @pablogiralt,

Thank you for the PR. It's indeed a good point to let the users of the plugin specify another origin than Magento's itself.
We are going to review the PR and merge it asap.

Best,
Attila
Adyen

etc/adminhtml/system/adyen_pwa.xml
```diff
@@ -0,0 +1,44 @@
...
+            <![CDATA[
+                <p>
+                This settings are specific for PWA integrations
+                </p>
+            ]]>
+        </comment>
+
+        <field id="payments_origin_url" translate="label" type="text" sortOrder="10" showInDefault="1" showInWebsite="1" showInStore="1">
+            <label>Payment Origin URL</label>
+            <tooltip>Only relevant if you process payment from an external url different to that of magento</tooltip>
```
acampos1916, line 39:
```suggestion
            <tooltip>Only relevant if you process payments from an external URL different to that of Magento</tooltip>
```

pablogiralt:
@acampos1916 @cyattilakiss added your suggested changes.  Can you review? 

Thanks

etc/adminhtml/system/adyen_pwa.xml
```diff
@@ -0,0 +1,44 @@
...
+ */
+-->
+<include xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Config:etc/system_include.xsd">
+    <group id="adyen_pwa" translate="label" type="text" sortOrder="95" showInDefault="1" showInWebsite="1" showInStore="1">
+        <label><![CDATA[Advanced: PWA]]></label>
+        <frontend_model>Magento\Config\Block\System\Config\Form\Fieldset</frontend_model>
+        <comment>
+            <![CDATA[
+                <p>
+                This settings are specific for PWA integrations
```
Aleffio, line 32:
```suggestion
                These settings are specific for PWA integrations
```

Aleffio (approved):

cyattilakiss (approved):

PR #788, opened 2020-08-10
[PW-2567] Guest checkout tokenisation

AlexandrosMor:
**Description**
This PR is about to tokenise all the payments when available even if the user is guest because some merchants work with a long delivery time and if they deliver after x weeks they are not able to capture the money without a token available.

**Tested scenarios**

Store token in Adyen when this config is enabled and the user is guest.
Don't store token when this config is not enabled and the user is guest.
Don't change the tokenisation when the user is not a guest.

**Fixed issue**:

### DIFF
--- a/Helper/Requests.php
+++ b/Helper/Requests.php
@@ -24,6 +24,7 @@
 namespace Adyen\Payment\Helper;
 
 use Adyen\Payment\Observer\AdyenOneclickDataAssignObserver;
+use Adyen\Util\Uuid;
 use Magento\Framework\App\Helper\AbstractHelper;
 use Magento\Vault\Model\Ui\VaultConfigProvider;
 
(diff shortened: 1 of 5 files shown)


Helper/Requests.php
```diff
@@ -37,12 +38,13 @@ class Requests extends AbstractHelper
      * @var \Adyen\Payment\Helper\Data
      */
     private $adyenHelper;
-
     /**
      * Requests constructor.
      *
      * @param Data $adyenHelper
+     * @param \Adyen\Payment\Logger\AdyenLogger $adyenLogger
```
cyattilakiss, line 45:
```suggestion
```

etc/config.xml
```diff
@@ -103,6 +103,7 @@
                 <can_void>1</can_void>
                 <can_cancel>1</can_cancel>
                 <group>adyen</group>
+                <guest_checkout_tokenisation>0</guest_checkout_tokenisation>
```
cyattilakiss, line 106: What is the purpose of having this here?

AlexandrosMor: I removed it

PR #792, opened 2020-08-14
Fetch country id from order address instead of country

JeroenVanLeusden:
**Description**
`\Magento\Sales\Api\Data\OrderAddressInterface` does not contain `getCountry()` method.

**Tested scenarios**
- Create order using pay by mail payment method.
- Open url and check if `countryCode` is added to url.


acampos1916:
@JeroenVanLeusden I tested it and indeed the country code is fetched with `getCountryId()`.
`getCountry()` is not present there.

acampos1916 (approved):

cyattilakiss (approved):

PR #793, opened 2020-08-17
Fix swish payment methods when running it with the swish app

rikterbeek:
For some payment methods like swish with resultCode pending we store this into the property adyenPaymentData instead of paymentData. Check if adyenPaymentData exists if so add this as paymentData in request. This will solve issues for Swish payment method. We need to refactor a lot of code to make everywhere to use consistency in the naming this will be a separate story to be picked up.

**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Controller/Process/Result.php
+++ b/Controller/Process/Result.php
@@ -392,7 +392,6 @@ class Result extends \Magento\Framework\App\Action\Action
         if (!empty($this->_session->getLastRealOrder()) &&
             !empty($this->_session->getLastRealOrder()->getPayment())
         ) {
-
             if (!empty($this->_session->getLastRealOrder()->getPayment()->getAdditionalInformation('paymentData'))) {
                 $request['paymentData'] = $this->_session->getLastRealOrder()->getPayment()->
                 getAdditionalInformation('paymentData');
--- a/Controller/Process/Result.php
+++ b/Controller/Process/Result.php
@@ -402,7 +402,7 @@ class Result extends \Magento\Framework\App\Action\Action
             }
 
             // for pending payment that redirect we store this under adyenPaymentData
-            // we need to refactor the code in the plugin that all paymentData is stored in paymentData and not in adyenPaymentData
+            // TODO: refactor the code in the plugin that all paymentData is stored in paymentData and not in adyenPaymentData
             if (!empty($this->_session->getLastRealOrder()->getPayment()->getAdditionalInformation('adyenPaymentData'))) {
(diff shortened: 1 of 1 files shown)


Controller/Process/Result.php
```diff
@@ -390,11 +390,31 @@ protected function validatePayloadAndReturnResponse($response)
...
-            !empty($this->_session->getLastRealOrder()->getPayment()->getAdditionalInformation("paymentData"))
+            !empty($this->_session->getLastRealOrder()->getPayment())
         ) {
-            $request['paymentData'] = $this->_session->getLastRealOrder()->getPayment()->
-            getAdditionalInformation("paymentData");
+
+            if (!empty($this->_session->getLastRealOrder()->getPayment()->getAdditionalInformation('paymentData'))) {
+                $request['paymentData'] = $this->_session->getLastRealOrder()->getPayment()->
+                getAdditionalInformation('paymentData');
+
```
acampos1916, line 399:
```suggestion
```

Controller/Process/Result.php
```diff
@@ -390,11 +390,30 @@ protected function validatePayloadAndReturnResponse($response)
...
+                // remove paymentData from db
+                $this->_session->getLastRealOrder()->getPayment()->unsAdditionalInformation('paymentData');
+                $this->_session->getLastRealOrder()->getPayment()->save();
+            }
+
+            // for pending payment that redirect we store this under adyenPaymentData
+            // we need to refactor the code in the plugin that all paymentData is stored in paymentData and not in adyenPaymentData
```
Aleffio, line 406: maybe marking this as a todo or with a story number so we remember to remove this comment when this is done?
```suggestion
            // TODO: refactor the code in the plugin that all paymentData is stored in paymentData and not in adyenPaymentData
```

Controller/Process/Result.php
```diff
@@ -390,11 +390,30 @@ protected function validatePayloadAndReturnResponse($response)
         $request = [];
 
         if (!empty($this->_session->getLastRealOrder()) &&
-            !empty($this->_session->getLastRealOrder()->getPayment()) &&
-            !empty($this->_session->getLastRealOrder()->getPayment()->getAdditionalInformation("paymentData"))
+            !empty($this->_session->getLastRealOrder()->getPayment())
         ) {
-            $request['paymentData'] = $this->_session->getLastRealOrder()->getPayment()->
-            getAdditionalInformation("paymentData");
+
```
acampos1916, line 395:
```suggestion
```

PR #808, opened 2020-09-02
[PW-2096] Validate agreements before placing order for Google Pay

msilvagarcia:
**Description**

Currently, the Google Pay component does not respect additional validation in Checkout page. This pull request attempts to fix this issue by making the Google Pay button disabled by default and enabling it only when the validation is successful.

**Tested scenarios**

Checked button status in Checkout page while changing the agreement to the store's terms.

**Fixed issue**:

### DIFF
--- a/view/frontend/web/template/payment/google-pay-form.html
+++ b/view/frontend/web/template/payment/google-pay-form.html
@@ -38,7 +38,7 @@
         </label>
     </div>
 
-    <div class="payment-method-content">
+    <div class="payment-method-content" data-bind="event: { change: onPaymentMethodContentClick }">
         <div class="payment-method-billing-address">
             <!-- ko foreach: $parent.getRegion(getBillingAddressFormName()) -->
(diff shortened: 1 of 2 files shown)


view/frontend/web/template/payment/google-pay-form.html
```diff
@@ -38,7 +38,7 @@
         </label>
     </div>
 
-    <div class="payment-method-content">
+    <div class="payment-method-content" data-bind="event: { change: onPaymentMethodContentClick }">
```
cyattilakiss, line 41: what if the google pay is the only enabled payment method, is the button by default visible and enabled or is it disabled already?

msilvagarcia: It is disabled by default. Which might cause problems when there are no agreements in place, now that I think about it.

acampos1916 (approved):

cyattilakiss (approved):

PR #816, opened 2020-09-15
[PW-2984]Add the minify exclude element in config

AlexandrosMor:
**Description**
Added the adyen.js and css files in the minification exclude config in magento. 

**Tested scenarios**
How to test it:
Set magento to developer mode
In the admin find the Minify JavaScript Files flag and set it to yes. (Store->Configuration->Advanced->Developer)
Set magento mode to production
Test the checkout page that loads properly

**Fixed issue**:

### DIFF
--- a/etc/config.xml
+++ b/etc/config.xml
@@ -236,6 +236,11 @@
                     <adyen_payment>/3.4.0/adyen</adyen_payment>
                 </minify_exclude>
             </js>
+            <css>
+                <minify_exclude>
+                    <adyen_payment>/3.4.0/adyen</adyen_payment>
+                </minify_exclude>
+            </css>
         </dev>
     </default>
 </config>
--- a/etc/config.xml
+++ b/etc/config.xml
@@ -230,5 +230,12 @@
                 <group>adyen</group>
             </adyen_google_pay>
         </payment>
+        <dev>
+            <js>
+                <minify_exclude>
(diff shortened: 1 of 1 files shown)


etc/config.xml
```diff
@@ -230,5 +230,17 @@
                 <group>adyen</group>
             </adyen_google_pay>
         </payment>
+        <dev>
+            <js>
+                <minify_exclude>
+                    <adyen_payment>/adyen</adyen_payment>
```
acampos1916, line 236: I think this will also exclude any local assets that matches to `*/adyen*`, is there a better pattern to match external URLs? I've recommended `adyen.com/checkoutshopper` before but I'm not sure if it's a good match

etc/config.xml
```diff
@@ -230,5 +230,17 @@
...
+                </minify_exclude>
+            </js>
+            <css>
+                <minify_exclude>
+                    <adyen_payment>/adyen</adyen_payment>
```
acampos1916, line 241: Same for CSS as above

acampos1916 (approved):

cyattilakiss (approved):

kpitn:
@AlexandrosMor the fix is not available on 6.6.1 => https://github.com/Adyen/adyen-magento2/blob/6.6.1/etc/config.xml

PR #824, opened 2020-09-22
PWA redirect support

### LINKED ISSUE #787: [PW-2871] 3DS1 in PWA
**Describe the bug**
I am developing Adyen x VueStorefront integration with Magento2. With small changes in plugin I was able to succeed with 3ds2 (https://github.com/Adyen/adyen-magento2/issues/785). However, it looks like I cannot support 3DS1. When I test card with 3ds1 I am getting:
```js
// GET /adyen/orders/<order_id>/payment-status
{
  threeDS2: false,
  type: 'RedirectShopper'
}
```
How should I consume that response? In Adyen Dashboard, I have just "Open"-ed transaction. Payment is not done due to luck of 3ds1 auth.

**Expected behavior**
Support for 3DS1 Auth in PWAs

real34:
**Description**
This PR brings several new features that enable a wider range of payment flows in a PWA context.
It was successfully tested (Q/A is still in progress for more in-depth testing) with the following methods: directEbanking, ideal, ebanking_FI, bcmc, dotpay, facilipay (Oney), klarna, giropay, trustly, 3DS and 3DS2.

**Tested scenarios**
- customer can be redirected to the payment provider: Magento now return correct `action` details for `RedirectShopper` values in the `/payment-status` endpoint
- customers can be redirected back to the PWA from the payment provider in redirect actions
- 3DS1 flows can now be fully achieved thanks to the new `/adyen/threeDSProcess` API endpoint (it would fix #787, however #821 introduced a "regression" preventing the user to be redirected back properly… to be discussed)
- PWA can get instant details about a payment so they can present an accurate information to the shopper ([according to the `resultCode`](https://docs.adyen.com/checkout/payment-result-codes)) no matter the platform it is redirected back from. A new `/adyen/orders/:orderId/payment-details` API endpoint was created to support this, because notifications are not processed fast enough (Magento receives them ~10-20 seconds after the customer is redirect back, and then it needs to be processed with the next CRON run)

**Fixed issue**:  #787 

## ToDo

- [ ] 3DS1 redirection seems to now be failing since #821 reverted the feature added in #806 and #786. I think that we need to find the correct canonical way to use correct urls (and this PR will have to be updated to use it) 
- [ ] improve the `/adyen/orders/:orderId/payment-details` new API endpoint:
  - [ ] I'd appreciate your feedback on the signature and ways to receive params, so it feels natural and consistent with other endpoints
  - [ ] it seems to me that it might need further security to prevent leaking sensitive information. The most important I can think of is ensuring that the customers are allowed to get details about the order. Your feedback is welcome, and if possible I'd appreciate to be shown other locations in the code where it is implemented (I couldn't fine what security strategies were in place for the `/payment-status` API)
- [ ] implement a way to restore quote in a PWA context so customers are not frustrated if they cancel their payment (it could be another PR/issue by itself). The Adyen module calls `$session->restoreQuote()` at controller level because it is session based. The code could be duplicated to do it from an order id. However I'd appreciate your opinion about how / when it should be done… in the `payment-details` endpoint? A new separate endpoint?
See https://github.com/magento/magento2/blob/390505a93555198c93331ed9159664fbccb70cd9/app/code/Magento/Checkout/Model/Session.php#L558-L567

Please let me know if you need further information.

### DIFF
--- a/Model/AdyenOrderPaymentDetails.php
+++ b/Model/AdyenOrderPaymentDetails.php
@@ -1,0 +1,105 @@
+<?php
+/**
+ *                       ######
+ *                       ######
+ * ############    ####( ######  #####. ######  ############   ############
+ * #############  #####( ######  #####. ######  #############  #############
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen Payment module (https://www.adyen.com/)
+ *
+ * Copyright (c) 2019 Adyen BV (https://www.adyen.com/)
+ * See LICENSE.txt for license details.
+ *
+ * Author: Adyen <magento@adyen.com>
+ */
+
+namespace Adyen\Payment\Model;
+
+use Adyen\Payment\Api\AdyenOrderPaymentDetailsInterface;
+
+class AdyenOrderPaymentDetails implements AdyenOrderPaymentDetailsInterface
+{
+
+    /**
+     * @var \Adyen\Payment\Helper\Data
+     */
+    private $adyenHelper;
+
+    /**
+     * @var \Magento\Sales\Model\OrderFactory
+     */
+    private $orderFactory;
+
+    /**
+     * @var \Adyen\Payment\Logger\AdyenLogger
+     */
+    private $adyenLogger;
+
+    /**
+     * AdyenThreeDS2Process constructor.
+     *
+     * @param \Adyen\Payment\Helper\Data $adyenHelper
+     * @param \Magento\Sales\Model\OrderFactory $orderFactory
+     * @param \Adyen\Payment\Logger\AdyenLogger $adyenLogger
+     */
+    public function __construct(
+        \Adyen\Payment\Helper\Data $adyenHelper,
+        \Magento\Sales\Model\OrderFactory $orderFactory,
+        \Adyen\Payment\Logger\AdyenLogger $adyenLogger
+    ) {
+        $this->adyenHelper = $adyenHelper;
+        $this->orderFactory = $orderFactory;
+        $this->adyenLogger = $adyenLogger;
+    }
+
+    public function getDetails($orderId, $payload)
+    {
+        // Decode payload from frontend
+        $payload = json_decode($payload, true);
+
+        // Validate JSON that has just been parsed if it was in a valid format
+        if (json_last_error() !== JSON_ERROR_NONE) {
+            throw new \Magento\Framework\Exception\LocalizedException(
+                __('Payment details failed because the request was not a valid JSON')
+            );
+        }
+
+        // Create order by order id
+        $order = $this->orderFactory->create()->load($orderId);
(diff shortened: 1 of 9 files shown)


Model/AdyenOrderPaymentDetails.php
```diff
@@ -0,0 +1,105 @@
...
+
+        // Validate JSON that has just been parsed if it was in a valid format
+        if (json_last_error() !== JSON_ERROR_NONE) {
+            throw new \Magento\Framework\Exception\LocalizedException(
+                __('Payment details failed because the request was not a valid JSON')
+            );
+        }
+
+        // Create order by order id
+        $order = $this->orderFactory->create()->load($orderId);
```
acampos1916, line 76: Should we use a ResourceModel instead? `load()` I believe is deprecated. It is being used in other parts of the plugin still but we can prevent introducing new usages.

Model/AdyenOrderPaymentDetails.php
```diff
@@ -0,0 +1,105 @@
...
+            throw new \Magento\Framework\Exception\LocalizedException(
+                __('Payment details are not available for this order')
+            );
+        }
+
+        $additionalInformation = $payment->getAdditionalInformation();
+        if (!empty($additionalInformation['paymentData']) && empty($payload['paymentData'])) {
+            $payload['paymentData'] = $additionalInformation['paymentData'];
+        }
+        // ToDo check that Order belongs to the customer or can be queried
```
acampos1916, line 88: @cyattilakiss any ideas on how to check if the order belongs to the customer here?

brightonmike:
Is there any update on this? We are attempting to integrate Klarna into a headless checkout and we're unable to do so as the button rendered does not do anything  - I believe this is because I am not getting a redirect URL for the button to use when clicked.

acampos1916:
Hello @real34,

Again thank you for this PR, we have some reports of users adapting it to their needs in order to implement PWA projects. For the moment we've decided not to merge it as we're working on a major refactoring of the plugin that would make this change unnecessary. The release candidate branch can be seen here https://github.com/Adyen/adyen-magento2/tree/7.0.0-rc.1.

@brightonmike please feel free to use this PR as a patch if it resolves your Klarna issue or wait for the v7 release.

Thanks again!

real34:
@acampos1916 thanks for your feedback.
Do you have any ETA of the 7.0.0 release?

We have customers awaiting for us to ensure everything is working fine with an official Adyen version instead of our fork. I'd like to come back to them to provide a seamless migration path and see in our PWA if we'll be able to be backwards compatible.

acampos1916:
@real34 we estimate that the v7 would be released in 3-4 weeks. Everyone’s feedback on this new major version is important, so let us know if you encounter any issues.

haelbichalex:
@acampos1916 Is there a new schedule for the v7 release already? Unfortunately we had to disable creditcard payments in our shop because of this issue.

Etto91:
hi, i have tried the 7.0.0-rc.2 but i got this error on the magento checkout.
What am i doing wrong?

Screenshot 2021-03-26 at 13 22 47

acampos1916:
> hi, i have tried the 7.0.0-rc.2 but i got this error on the magento checkout.
> What am i doing wrong?
@Etto91 `retrievePaymentMethods` is provided by `adyenPaymentService` (`Adyen_Payment/js/model/adyen-payment-service`), I would check if that file is correctly deployed in your instance https://github.com/Adyen/adyen-magento2/blob/7.0.0-rc.2/view/frontend/web/js/model/adyen-payment-service.js

PR #830, opened 2020-09-30
[PW-3095] Set reservedOrderId to null before placing an order

AlexandrosMor:
**Description**
In case a customer tries to pay with a non Adyen payment method which reserves the order id then our payment methods will use that reserved order id until we have a successful payment which resets it. It causes issues since we can get multiple refused or offer closed notifications which closes the order before we finally make an authoirze one.

To solve this we added an interceptor and every payment-information request we would set the quote reservedOrderId to null. Therefore we wouldn’t bump unnecessary the orderIds and if it’s null it would not do anything. Also with this approach we can make sure every request of ours has a new order id as merchant reference even if any other service sets the reserved order id before it gets to us.

**Tested scenarios**

**Fixed issue**:


Plugin/PaymentInformationResetOrderId.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
```
acampos1916, line 2: Add Copyright comment 👍

Plugin/PaymentInformationResetOrderId.php
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Adyen\Payment\Plugin;
+
```
acampos1916, line 5:
```suggestion
```

Plugin/PaymentInformationResetOrderId.php
AlexandrosMor: @acampos1916  done 👍

etc/di.xml
```diff
@@ -1028,6 +1028,12 @@
     <type name="Magento\Vault\Api\PaymentTokenRepositoryInterface">
         <plugin name="AdyenPaymentVaultDeleteToken" type="Adyen\Payment\Plugin\PaymentVaultDeleteToken" sortOrder="10"/>
     </type>
+    <type name="Magento\Checkout\Model\PaymentInformationManagement">
+        <plugin name="AdyenPaymentInformationResetOrderId" type="Adyen\Payment\Plugin\PaymentInformationResetOrderId" sortOrder="20"/>
```
acampos1916, line 1032: These don't have to be sequential 
```suggestion
        <plugin name="AdyenPaymentInformationResetOrderId" type="Adyen\Payment\Plugin\PaymentInformationResetOrderId" sortOrder="10"/>
```

etc/di.xml
```diff
@@ -1028,6 +1028,12 @@
...
+    </type>
+    <type name="Magento\Checkout\Model\GuestPaymentInformationManagement">
+        <plugin name="AdyenPaymentInformationResetOrderId" type="Adyen\Payment\Plugin\PaymentInformationResetOrderId" sortOrder="30"/>
```
acampos1916, line 1035: These don't have to be sequential
```suggestion
        <plugin name="AdyenPaymentInformationResetOrderId" type="Adyen\Payment\Plugin\PaymentInformationResetOrderId" sortOrder="10"/>
```

etc/di.xml
```diff
@@ -1028,6 +1028,12 @@
     <type name="Magento\Vault\Api\PaymentTokenRepositoryInterface">
         <plugin name="AdyenPaymentVaultDeleteToken" type="Adyen\Payment\Plugin\PaymentVaultDeleteToken" sortOrder="10"/>
     </type>
+    <type name="Magento\Checkout\Model\PaymentInformationManagement">
```
cyattilakiss, line 1031: does it work if you change it to the interface?

etc/di.xml
```diff
@@ -1028,6 +1028,12 @@
...
+        <plugin name="AdyenPaymentInformationResetOrderId" type="Adyen\Payment\Plugin\PaymentInformationResetOrderId" sortOrder="10"/>
+    </type>
+    <type name="Magento\Checkout\Model\GuestPaymentInformationManagement">
```
cyattilakiss, line 1034: here as well?

cyattilakiss (approved):

acampos1916 (approved):

msilvagarcia (approved):

faelkeller:
Hi guys! 
What is the reason to reset an used reservedOrderId? 
I need it to control some parts of system and this PR is reseting it.
If do I remove this code, can I have any issue on integration with adyen payment?
Regards,

PR #840, opened 2020-10-05
[PW-3174] Use the generic component for card payments

cyattilakiss:
**Description**
Adjust the card payments to use the generic component and it's new features

**Tested scenarios**

**Fixed issue**:


Gateway/Validator/CheckoutResponseValidator.php
```diff
@@ -69,25 +69,27 @@ public function validate(array $validationSubject)
         // validate result
         if (!empty($response['resultCode'])) {
             $payment->setAdditionalInformation('resultCode', $response['resultCode']);
+
```
AlexandrosMor, line 72: Could we outsourse those lines so the PaymentResponseHandler can benefit? These fields validation is common for both classes

cyattilakiss: wanted to do it in a separate PR because this PR is already quite big.

AlexandrosMor (approved):

acampos1916 (approved):

PR #817, opened 2020-09-16
Release 6.6.0

acampos1916:
**Features**
#795 #822 SEPA Direct Debit tokenization 
#809 Separate checks for IP and HMAC for notifications received
#812 3DS2/3DS1 fallback
#816 Add the minify exclude element in config 

**Fixes**
#808 Validate checkout agreements before placing order with Google Pay
#813 Remove dependency for Magento setup exception
#814 Fix for Vault with 3DS2 Credit Cards 
#821 3DS storeview URL fix for redirect

PR #850, opened 2020-10-09
[PW-3220] Revert the config to exclude adyen js from minification

AlexandrosMor:
**Description**
We had to remove the fix from the config.xml to exclude the adyen checkout component js file from minification. The reason is that the ```<minify_exclude>``` is supported by magento version 2.3 and above and not from the older versions. We are planning to make a generic fix for this which will covers all the versions.
For now merchants who are using 2.3 and above can use the adyen-magento2 [version release 6.6.0](https://github.com/Adyen/adyen-magento2/pull/817)

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/etc/config.xml
+++ b/etc/config.xml
@@ -230,17 +230,5 @@
                 <group>adyen</group>
             </adyen_google_pay>
         </payment>
-        <dev>
-            <js>
-                <minify_exclude>
-                    <adyen_payment>adyen.com/checkoutshopper</adyen_payment>
-                </minify_exclude>
-            </js>
-            <css>
-                <minify_exclude>
-                    <adyen_payment>adyen.com/checkoutshopper</adyen_payment>
-                </minify_exclude>
-            </css>
-        </dev>
     </default>
 </config>


msilvagarcia (approved):

cyattilakiss (approved):

Mecalux-Developer:
Is the generic fix done yet?

Dnd-Grym:
Hi ! @AlexandrosMor 
Do you have a date estimation for this generic fix ?
regards

AlexandrosMor:
Hello @Dnd-Grym,

We are currently testing the new version and if everything goes well we are planning to release the plugin in the coming two weeks. 

kind regards,
Alexandros

acampos1916:
Hello @cosmicvibes, in version 7.x.x of the plugin the requirement for the external JS file has been removed and it is now bundled in the plugin, which makes it possible to enable the minification without issues. Cheers!

cosmicvibes:
Thanks @acampos1916, sorry I totally didn't reply - thanks for the update!

PR #868, opened 2020-11-02
Allow guest checkout for PMs requiring shopperReference such as Oney

ataberkylmz:
**Description**

Enabled 'shopperReference' key to be included in all of the checkout requests instead of only two scenarios where shopper either needs to be logged in or guest tokenization setting should be enabled. This way, payment methods such as Oney (Facilypay) can be used even if the shopper is not logged in. 
**Tested scenarios**

N/A
**Fixed issue**:  
PW-2493


acampos1916:
Thanks @ataberkylmz,

Hey @cyattilakiss would it it be an issue to always generate and include the shopperReference?

msilvagarcia:
Yeah, I believe so. This check was introduced in #788 , so I don't think it would be good to just remove it.

cyattilakiss:
> Thanks @ataberkylmz,
> 
> Hey @cyattilakiss would it it be an issue to always generate and include the shopperReference?

Hey @acampos1916 
I am not aware if it would cause any issue. In prestashop for example we always have a shopperReference even for guest users.

We send the storePaymentMethod = true in the request only if the recurring is set or the oneclick + customer checkbox is clicked. Other than that since it's a unique reference it will just appear in the adyen overview and that's all.

Before the #788 we only sent shopperReference when the customer was logged in, the reason was because we believed the shopperReference is only required for payment methods which for some reason requires it but it turned out that it's not the case and it's okay to send shopperReference for guest users as well as long as they are unique.

msilvagarcia:
@cyattilakiss are you saying that merchants could capture these payment methods even without a token after a long time has passed?

cyattilakiss:
> @cyattilakiss are you saying that merchants could capture these payment methods even without a token after a long time has passed?

No, this change would only allow payment methods to work which requires the shopperReference but the token would not be created because that is only created when the storePaymentMethod = true field is also in the request.

AlexandrosMor (approved):

cyattilakiss (approved):

AlexandrosMor:
Test scenarios

with logged in shopper:
end2end tests
Oney payment for logged 
Sepa
CC and save payment method
oneclick 
afterpay
klarna stored

With guest shopper:
end2end tests
Oney payment for guest  
Ideal
Afterpay

Oney pay worked with Aggreement type one click and reccuring

PR #874, opened 2020-11-04
[PW-3387] Fix One click with 3DS1 returns error 

AlexandrosMor:
**Description**
After introducing the same site cookie fix the oneclick with 3DS1 was using the RedirectDataBuilder and the shopper was not redirecting to transparent\redirect. Removing it from di.xml the redirect for oneclick 3DS1 work properly
**Tested scenarios**
logged in user
oneclick 
oneclick 3ds1 

**Fixed issue**:  #


Helper/Requests.php
```diff
@@ -364,10 +364,7 @@ public function buildThreeDS2Data($additionalData, $storeId, $request = [])
      */
     public function buildRedirectData($storeId, $request = [])
```
cyattilakiss, line 365: is this function used anywhere else than the redirectDataBuilder.php? Let's check it and if not then it can be removed.

msilvagarcia (approved):

acampos1916 (approved):

cyattilakiss (approved):

PR #880, opened 2020-11-10
Update adyen-payment-service.js

KadoBOT:
Clean JS file.


AlexandrosMor (approved):
Ternary expression reduce readability in my opinio

PR #882, opened 2020-11-10
Update adyen-hpp-method.js

KadoBOT:
Improve code

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
@@ -99,7 +99,15 @@ define(
 
                 adyenPaymentService.retrieveAvailablePaymentMethods(function () {
                     var paymentMethods = adyenPaymentService.getAvailablePaymentMethods();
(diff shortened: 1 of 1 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -99,7 +99,15 @@ define(
 
                 adyenPaymentService.retrieveAvailablePaymentMethods(function () {
                     var paymentMethods = adyenPaymentService.getAvailablePaymentMethods();
-                    if (JSON.stringify(paymentMethods).indexOf("ratepay") > -1) {
+                    var paymentMethodsStr;
+                    
+                    try {
+                        paymentMethodsStr = JSON.stringify(paymentMethods);
+                    } catch (e) {
+                        // Failed to stringify paymentMethods
```
AlexandrosMor, line 107: In case of an exception why there is no action ?

KadoBOT: Nothing will happen if there is an error. Before that, when there was no `try..catch`, it would throw an error and the rest of the javascript wouldn't be loaded.

But feel free to add some proper error handling there.

PR #890, opened 2020-11-24
Use Oney as Open Invoice payment method in Cron

msilvagarcia:
**Description**

Oney payment methods are not being processed as an open invoice payment method, not allowing for manual capture. Since Oney is an open invoice, I suggest we add it to the list in `Adyen\Payment\Helper\Data`.

**Tested scenarios**

None.

**Fixed issue**:

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1066,7 +1066,8 @@ class Data extends AbstractHelper
     {
         if (strpos($paymentMethod, 'afterpay') !== false ||
             strpos($paymentMethod, 'klarna') !== false ||
-            strpos($paymentMethod, 'ratepay') !== false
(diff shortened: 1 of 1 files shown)


cyattilakiss (approved):

Aleffio (approved):
something to keep in mind is that Oney does not allow Partial Captures(unlike other invoice methods) https://docs.adyen.com/payment-methods/oney#shopper-experience

cyattilakiss:
> something to keep in mind is that Oney does not allow Partial Captures(unlike other invoice methods) https://docs.adyen.com/payment-methods/oney#shopper-experience

@msilvagarcia could you please create a follow up ticket to check this and add a restriction if needed?

PR #899, opened 2020-12-03
[PW-3462] Remove mapping the bcmc to maestro

AlexandrosMor:
**Description**

The mapping is not needed anymore. The bancontact is available in magento and displayed in orders
**Tested scenarios**

Test the bancontact in magento2 using Adyen test card and checking if it is authorised as bcmc in the backoffice and in the orders if it is set correct

**Fixed issue**:  #899

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
@@ -190,13 +190,7 @@ define(
                                     self.installments(0);
                                 }
                             }
-
-                            // for BCMC as this is not a core payment method inside magento use maestro as brand detection
-                            if (creditCardType == "BCMC") {
-                                self.creditCardType("MI");
-                            } else {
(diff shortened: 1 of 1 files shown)


acampos1916 (approved):

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -190,13 +190,7 @@ define(
...
                                 }
                             }
-
-                            // for BCMC as this is not a core payment method inside magento use maestro as brand detection
-                            if (creditCardType == "BCMC") {
-                                self.creditCardType("MI");
-                            } else {
-                                self.creditCardType(creditCardType);
-                            }
+                            self.creditCardType(creditCardType);
```
cyattilakiss, line 193: looks good, maybe a follow up ticket would be nice to update it the card type based on the payments or the payments/details response (IF available) :)

cyattilakiss (approved):

PR #910, opened 2020-12-14
[PW-3157] [iDeal] First notification is unsuccessful but a second notification 2 minutes later send a success response => order canceled but payment is in success

cyattilakiss:
**Description**
Delay processing of AUTHORISATION success=false notifications by 10 min
In some edge cases (for example 3DS1 authentication) an authorisation
success=false notification can be sent earlier than the success=true one
The reason can be a closed and reopened mobile application where the
customer performs the 3DS1 authentication
To avoid closing false unsuccessful orders, processing authorisation
success=false and offer_closed notifications are delayed

**Tested scenarios**
OFFER_CLOSED and AUTHORISATION success=false notification processing

**Fixed issue**:

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -382,26 +382,68 @@ class Cron
      */
     private function shouldSkipProcessingNotification($notification)
     {
-        // OFFER_CLOSED notifications needs to be at least 10 minutes old to be processed
-        $offerClosedMinDate = new \DateTime();
-        $offerClosedMinDate->modify('-10 minutes');
+        switch ($notification['event_code']) {
+            // Remove OFFER_CLOSED and AUTHORISATION success=false notifications for some time from the processing list
+            // to ensure they won't close any order which has an AUTHORISED notification arrived a bit later than the
+            // OFFER_CLOSED or the AUTHORISATION success=false one.
+            case Notification::OFFER_CLOSED:
+                // OFFER_CLOSED notifications needs to be at least 10 minutes old to be processed
+                $offerClosedMinDate = new \DateTime();
+                $offerClosedMinDate->modify('-10 minutes');
+                $createdAt = \DateTime::createFromFormat('Y-m-d H:i:s', $notification['created_at']);
+                // To get the difference between $offerClosedMinDate and $createdAt, $offerClosedMinDate time in seconds
+                // is deducted from $createdAt time in seconds, divided by 60 and rounded down to integer
+                $minutesUntilProcessing = floor(
+                    ($createdAt->getTimestamp() - $offerClosedMinDate->getTimestamp()) / 60
+                );
 
-        // Remove OFFER_CLOSED notifications arrived in the last 10 minutes from the list to process to ensure it
-        // won't close any order which has an AUTHORISED notification arrived a bit later than the OFFER_CLOSED one.
-        $createdAt = \DateTime::createFromFormat('Y-m-d H:i:s', $notification['created_at']);
-        // To get the difference between $offerClosedMinDate and $createdAt, $offerClosedMinDate time in seconds is
-        // deducted from $createdAt time in seconds, divided by 60 and rounded down to integer
-        $minutesUntilProcessing = floor(($createdAt->getTimestamp() - $offerClosedMinDate->getTimestamp()) / 60);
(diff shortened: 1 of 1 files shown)


Model/Cron.php
```diff
@@ -382,26 +382,68 @@ public function processNotification()
...
+                if (
+                    strcmp($notification['success'], 'true') == 0 ||
+                    strcmp($notification['success'], '1') == 0
+                ) {
+                    // do not skip this notification but process it now
+                    return false;
+                }
+
+                // AUTHORISATION success=false notifications needs to be at least 10 minutes old to be processed
+                $authorisationSuccessFalseMinDate = new \DateTime();
```
acampos1916, line 425: Maybe we should combine both switch cases or move the timestamp generation/comparison outside the switch as the logic is pretty similar? I think the only difference is the log message

cyattilakiss: was thinking to move the `$createdAt = \DateTime::createFromFormat('Y-m-d H:i:s', $notification['created_at']);` line outside of the switch but the -10 min calculation I wanted to keep notification specific so if needed we can change the minutes per notification type.

acampos1916: I see, sounds good. Maybe when the cron is refactored we have notification classes and with a delayTime property

acampos1916 (approved):

Model/Cron.php
cyattilakiss: yes definitely the refactored version of this fix is already defined here: PW-3638

Model/Cron.php
```diff
@@ -382,26 +382,68 @@ public function processNotification()
...
-                )
-            );
+        switch ($notification['event_code']) {
+            // Remove OFFER_CLOSED and AUTHORISATION success=false notifications for some time from the processing list
+            // to ensure they won't close any order which has an AUTHORISED notification arrived a bit later than the
+            // OFFER_CLOSED or the AUTHORISATION success=false one.
+            case Notification::OFFER_CLOSED:
+                // OFFER_CLOSED notifications needs to be at least 10 minutes old to be processed
+                $offerClosedMinDate = new \DateTime();
+                $offerClosedMinDate->modify('-10 minutes');
```
msilvagarcia, line 392:
```suggestion
                $offerClosedMinDate = new \DateTime('-10 minutes');
```

Model/Cron.php
```diff
@@ -382,26 +382,68 @@ public function processNotification()
...
+                $createdAt = \DateTime::createFromFormat('Y-m-d H:i:s', $notification['created_at']);
+                // To get the difference between $offerClosedMinDate and $createdAt, $offerClosedMinDate time in seconds
+                // is deducted from $createdAt time in seconds, divided by 60 and rounded down to integer
+                $minutesUntilProcessing = floor(
+                    ($createdAt->getTimestamp() - $offerClosedMinDate->getTimestamp()) / 60
+                );
```
msilvagarcia, line 398:
```suggestion
                $minutesUntilProcessing = $createdAt->diff($offerClosedMinDate)->i;
```

Model/Cron.php
```diff
@@ -382,26 +382,68 @@ public function processNotification()
...
+                }
+
+                // AUTHORISATION success=false notifications needs to be at least 10 minutes old to be processed
+                $authorisationSuccessFalseMinDate = new \DateTime();
+                $authorisationSuccessFalseMinDate->modify('-10 minutes');
+                $createdAt = \DateTime::createFromFormat('Y-m-d H:i:s', $notification['created_at']);
+
+                $minutesUntilProcessing = floor(
+                    ($createdAt->getTimestamp() - $authorisationSuccessFalseMinDate->getTimestamp()) / 60
+                );
```
msilvagarcia, line 431: If accepted, apply suggestions from above here as well :)

msilvagarcia (changes requested):
Some suggestions added to use more features from the PHP APIs.

Maybe we can extract a method to determine when a notification can be processed based on the time alone? I remember there are more points here where we can reuse this.

msilvagarcia (commented):
Check comments in another review, sorry for the noise.

acampos1916 (approved):

peterojo (approved):

PR #919, opened 2020-12-21
[PW-3130] Use the generic component for stored payment methods

cyattilakiss:
**Description**
Use the generic way of implementation and the new components in the one click payment methods as well (the same way as card payment methods and alternative payment methods)

**Tested scenarios**
One click payment method using billing agreements, non 3D, 3DS, 3DS2
Recurring payment method using billing agreements

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1713,7 +1713,7 @@ class Data extends AbstractHelper
                         $billingAgreement->getAgreementId(),
                         $order->getId()
                     )) {
-                        // save into sales_billing_agreement_order
+                        // save into billing_agreement_order
                         $billingAgreement->addOrderRelation($order);
                     }
                     // add to order to save agreement
@@ -1733,6 +1733,7 @@ class Data extends AbstractHelper
             $comment = $order->addStatusHistoryComment($message);
 
             $order->addRelatedObject($comment);
+            $order->save();
         }
     }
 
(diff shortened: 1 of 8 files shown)


Helper/Data.php
```diff
@@ -1733,6 +1733,7 @@ public function createAdyenBillingAgreement($order, $additionalData)
             $comment = $order->addStatusHistoryComment($message);
 
             $order->addRelatedObject($comment);
+            $order->save();
```
cyattilakiss, line 1736: create ticket to save the order not with the order object but with a resource model

view/frontend/web/template/payment/oneclick-form.html
```diff
@@ -71,6 +71,12 @@
             <fieldset
                     data-bind="attr: {class: 'fieldset payment items ccard ' + getCode(), id: 'payment_form_' + $parent.getCode() + '_' + value}">
 
+                <div id="oneclick_actionModalWrapper">
```
cyattilakiss, line 74: remove the other modal "threeDS2ModalOneClick"

view/frontend/web/js/view/payment/method-renderer/adyen-oneclick-method.js
```diff
@@ -21,548 +21,490 @@
...
+                            },
+                        };
+                        layout([messagesComponent]);
+
+                        messageComponents[name] = messageContainer;
+                    });
+                this.messageComponents = messageComponents;
+
+                let paymentMethodsObserver = adyenPaymentService.getPaymentMethods();
+                let paymentMethodsResponse = paymentMethodsObserver();
```
cyattilakiss, line 119: paymentMethodsObserver() double check it what does it return

acampos1916 (approved):

AlexandrosMor (approved):

PR #922, opened 2020-12-21
7.0.0-rc.1 changes

cyattilakiss:
**Important note for this PR:**
We are going to keep this PR in draft until we gather enough feedback for the 7.0.0-rc.1 release
Please note all the necessary release notes, config changes, doc changes etc here so it will be added to the release note of the 7.0.0 release

**Description**
This release introduces the following major changes and features:
### Generic implementation of Adyen's Web Components
Most of the frontend logic to render the payment methods and process callbacks has been replaced with a generic flow that handles all supported payment methods. The dependency on the external resource `adyen.js` has been replaced for a local `bundle.js` file. The component's result `data` will include most required parameters to execute API calls and the manual configuration of these requests has been removed.

### Payment Details API endpoint
`/V1/adyen/threeDS2Process`->`/V1/adyen/paymentDetails`: AdyenThreeDS2Process.php has been replaced with AdyenPaymentDetails.php. This new endpoint is used from the frontend to process a payload with the web component's result and the `orderId` in the following format:
```
{
    "details": {
        "threeds2.fingerprint": "..."
    },
    "paymentData": "...",
    "orderId": "..."
}
```
This will trigger a `/paymentDetails` call and the handling of the response. Depending on the result code the order will be placed, cancelled, or the [additional action](https://docs.adyen.com/checkout/components-web#step-4-additional-front-end) will be returned to the frontend to execute `createFromAction`.

PWA Implementations can use this endpoint to process payment details calls.

### Configuration to charge orders based on storeview or global currency
The plugin was designed to charge the order amount using the Display Currency. We're introducing a new configuration to also select the Global/Website Currency to process payments. Display Currency remains as the default option.

A new column has been introduced for the sales_order table to store the order's currency based on this new configuration at the time of placement, with a new event observer to populate it. This has been done so refunds can be processed with the correct currency.

This setting is initially hidden to ensure processing of payments in the expected currency. To modify this configuration use the Magento CLI.

``
bin/magento config:set payment/adyen_abstract/charged_currency (base|display)
``

* Display currency – The storeview or display currency is used (Default).
* Global/Website currency – The base currency is used. This will be either the Global or Website currency depending on how Magento is configured.


Setup/UpgradeSchema.php
```diff
@@ -402,4 +406,33 @@ public function updateSchemaVersion540(SchemaSetupInterface $setup)
...
+
+    /**
+     * Upgrade to 6.7.0
+     *
+     * New sales_order column for the currency charged based on the Adyen config option (base or display)
+     *
+     * @param SchemaSetupInterface $setup
+     * @return void
+     */
+    public function updateSchemaVersion670(SchemaSetupInterface $setup)
```
acampos1916, line 418: Use version 7.0.0 instead

PR #924, opened 2020-12-22
[PW-3111] Use getOrigin helper function for returnUrl parameter in redirect controller

cyattilakiss:
**Description**
Always add merchantReference in the returnUrl so order can be retrieved even if session is unavailable
Delete process/redirect controller -> not used anymore, process/result controller handles all the customer redirects back from payments
Use PWA origin when creating the process/result url in the transparent/redirect controller.


Controller/Process/Result.php
```diff
@@ -382,42 +384,57 @@ protected function _getOrder($incrementId = null)
...
      * @throws \Adyen\AdyenException
      */
-    protected function validatePayloadAndReturnResponse($response)
+    protected function validatePayloadAndReturnResponse($result)
     {
         $client = $this->_adyenHelper->initializeAdyenClient($this->storeManager->getStore()->getId());
         $service = $this->_adyenHelper->createAdyenCheckoutService($client);
 
-        $request = [];
+        if (!empty($result['merchantReference'])) {
```
acampos1916, line 392:
```suggestion
        if (!empty($result['merchantReference'])) {
                //Replace with order repository search for best practice
```

Controller/Process/Result.php
```diff
@@ -382,42 +384,57 @@ protected function _getOrder($incrementId = null)
...
+            $order = $this->_orderFactory->create()->loadByIncrementId($result['merchantReference']);
```
acampos1916, line 393:
```suggestion
        if (!empty($result['merchantReference'])) {
            //Replace with order repository search for best practice
            $order = $this->_orderFactory->create()->loadByIncrementId($result['merchantReference']);
```

acampos1916 (commented):
Only added a TODO

PR #930, opened 2020-12-24
Handle 3DS1 with Google pay

cyattilakiss:
**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Gateway/Request/GooglePayAuthorizationDataBuilder.php
+++ b/Gateway/Request/GooglePayAuthorizationDataBuilder.php
@@ -40,16 +40,23 @@ class GooglePayAuthorizationDataBuilder implements BuilderInterface
     private $adyenLogger;
 
     /**
+     * @var \Magento\Store\Model\StoreManagerInterface
+     */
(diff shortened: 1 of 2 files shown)


Gateway/Request/GooglePayAuthorizationDataBuilder.php
```diff
@@ -67,6 +76,10 @@ public function build(array $buildSubject)
             $this->adyenLogger->addAdyenDebug("PaymentToken is empty");
         }
 
+        $requestBody['returnUrl'] = $this->storeManager->getStore()->getBaseUrl(
```
acampos1916, line 79: Should we use the PWA origin conditionally as we did here? https://github.com/Adyen/adyen-magento2/blob/7.0.0-rc.1/Block/Transparent/Redirect.php#L74-L80

cyattilakiss: let's validate it with the test merchants because I am not sure if it's needed here as well

PR #938, opened 2020-12-29
[PW-3704] Move to Client key from Origin key

cyattilakiss:
**Description**
Use Client key instead of Origin key in the checkout flow
Still keep origin key for backend orders

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1611,6 +1611,7 @@ class Data extends AbstractHelper
      *
      * @return string
      * @throws \Adyen\AdyenException
+     * @deprecared please use getClientKey instead
      */
     public function getOriginKeyForBaseUrl()
     {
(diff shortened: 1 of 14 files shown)


Helper/Data.php
```diff
@@ -788,6 +788,33 @@ public function getAPIKey($storeId = null)
...
     }
 
+    /**
+     * Retrieve the Client key
+     *
+     * @param null|int|string $storeId
+     * @return string
+     */
+    public function getClientKey($storeId = null)
+    {
```
peterojo, line 798: Not a big deal but I thing this function can be simplified.

```suggestion
    return trim(
          $this->getAdyenAbstractConfigData(
              $this->isDemoMode($storeId) ? 'client_key_test' : 'client_key_live',
              $storeId
          )
      );
```

peterojo (approved):

acampos1916:
Pushed a commit to add the new config paths to the sensitive values array 19e0534

AlexandrosMor (approved):

AlexandrosMor (approved):

PR #941, opened 2021-01-07
[PW-3705] Implement importOrderPayment method in Adyen billing agreement

AlexandrosMor:
**Description**

Implement importOrderPayment method in Adyen billing agreement because so far we were using the one from Paypal. 
The issue is that the magento 2.4 is using php 7.4 as in this version if an array is not null but the index does not exist then throws exception. 
Adding the new implementation of `importOrderPayment` the `$baData = $payment->getBillingAgreementData()` is checked and if the billing_agreement_id is not exist then the recurringDetailReference is assigned to `baData['billing_agreement_id']`. 
The importOrderPayment method is used and replaced in Data.php and Cron.php

**Tested scenarios**

**Fixed issue**:  
Magento 2.3 
 Create a BA 
Cron create BA 
Magento 2.4 
Create a BA 
 Cron create BA

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1689,7 +1689,11 @@ class Data extends AbstractHelper
                     // create new BA
                     $billingAgreement = $this->billingAgreementFactory->create();
                     $billingAgreement->setStoreId($order->getStoreId());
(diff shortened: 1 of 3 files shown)


cyattilakiss (approved):
could you please also create a PW ticket tagged as refactoring so we can remove this piece of code in a later stage when we clean up this part?

acampos1916 (approved):

peterojo (approved):

Issue #739, opened 2020-06-15
[PW-2611] Unserialise error in 3DS 1 flow when storing vault token in adyen_cc


LRotherfield:
**Describe the bug**
When saving a vault token in the 3DS 1 flow using the adyen_cc method, the code errors here: https://github.com/Adyen/adyen-magento2/blob/c3a7e85cca44b76e2a101438ad521fde81c56f0c/Controller/Process/Redirect.php#L214 with the message `Unable to unserialize value. Error: Syntax error`.  This does not prevent the token from saving since the order is saved which saves the payment here: https://github.com/Adyen/adyen-magento2/blob/c3a7e85cca44b76e2a101438ad521fde81c56f0c/Controller/Process/Redirect.php#L223 
However it does prevent line 215 from being saved and it does create error lines in the logs:
https://github.com/Adyen/adyen-magento2/blob/c3a7e85cca44b76e2a101438ad521fde81c56f0c/Controller/Process/Redirect.php#L215

This happens because additional_data always seems to be null on payments, so line 214 tries to json_decode `NULL` which results in an `InvalidArgumentException` being thrown here magento/framework/Serialize/Serializer/Json.php:39

**To Reproduce**
Checkout using a 3DS 1 enabled card (not 3DS 2) and make sure to check the save for later checkbox to store the recurring token.

**Expected behavior**
No error should be thrown in the log and the payment should have `{"force_save":true}` in the additional_data column

**Additional context**
Magento version: "magento/product-community-edition": "2.3.4-p2"
Adyen module version: "adyen/module-payment": "6.0.0"
Module settings:
payment/adyen_abstract/demo_mode: Tested with both 1 and 0 (test and production)
payment/adyen_cc_vault/active: 1
payment/adyen_cc/active: 1
payment/adyen_cc/threeds2_enabled: 1
payment/adyen_oneclick/active: 1
payment/adyen_oneclick/recurring_payment_type: ONECLICK

acampos1916:
Hello @LRotherfield,

Thank you for the detailed information about this issue. We've created an internal ticket and will follow up to let you know of any progress made.

Cheers,
Ángel

cyattilakiss:
Hi @LRotherfield 

I will close this issue for now since the fix has been merged and released quite some time ago just eh issue hasn't been closed.
Please feel free to reopen it in case you still experience the problem after upgrading.

Best,
Attila
Adyen

PR #945, opened 2021-01-13
[PW-2611] Checking if order payment additional data is present before adding 'force_save' flag

acampos1916:
**Description**
If the order payment's additional data is NULL the redirect ends in error because there's an attempt to unserialize the value. We're adding an early check to prevent this.

**Tested scenarios**
3DS1 payments that return `recurring.recurringDetailReference` on `/payments/details` call won't end in unserialize error.

**Fixed issue**:  #739


cyattilakiss (dismissed):

Controller/Process/Redirect.php
```diff
@@ -205,10 +205,12 @@ public function execute()
                             $extensionAttributes = $this->getExtensionAttributes($order->getPayment());
                             $extensionAttributes->setVaultPaymentToken($paymentToken);
                             $orderPayment = $order->getPayment()->setExtensionAttributes($extensionAttributes);
-                            $add = $this->serializer->unserialize($orderPayment->getAdditionalData());
-                            $add['force_save'] = true;
-                            $orderPayment->setAdditionalData($this->serializer->serialize($add));
-                            $this->orderPaymentResource->save($orderPayment);
+                            if (!empty($orderPayment->getAdditionalData())) {
+                                $add = $this->serializer->unserialize($orderPayment->getAdditionalData());
```
peterojo, line 209:
```suggestion
                                $additionalData = $this->serializer->unserialize($orderPayment->getAdditionalData());
```

Controller/Process/Redirect.php
```diff
@@ -205,10 +205,12 @@ public function execute()
...
+                                $add['force_save'] = true;
```
peterojo, line 210:
```suggestion
                                $additionalData['force_save'] = true;
```

Controller/Process/Redirect.php
```diff
@@ -205,10 +205,12 @@ public function execute()
...
+                                $orderPayment->setAdditionalData($this->serializer->serialize($add));
```
peterojo, line 211:
```suggestion
                                $orderPayment->setAdditionalData($this->serializer->serialize($additionalData));
```

peterojo (changes requested):
I think you could be a bit more explicit with these variable names :)

peterojo (approved):

cyattilakiss (approved):

PR #950, opened 2021-01-19
[PW-3715] Add wallet payment methods Apple pay, Google pay and Paypal

cyattilakiss:
**Description**
Implemented new Apple pay integration, removed the old one as a separate payment method
Implemented new Google pay integration, removed the old one as a separate payment method
Implemented new PayPal integration

Fixed payment method logos in checkout

**Tested scenarios**


AlexandrosMor (approved):

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -164,234 +176,224 @@ define(
...
-                                    };
-                                }
+                                /*Use the storedPaymentMethod object and the custom onChange function as the configuration object together*/
+                                var configuration = Object.assign(paymentMethod,
+                                    {
+                                        showPayButton: showPayButton,
+                                        countryCode: country,
+                                        currency: quote.totals().quote_currency_code,
+                                        amount: { // Use this after above removed
+                                            value: 600, // todo get the real amount
```
acampos1916, line 288: This is not to be replaced in this PR?

peterojo (approved):

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
cyattilakiss: good one, indeed this needs to be done in this - adding it now

etc/frontend/di.xml
```diff
@@ -33,16 +33,13 @@
                 <item name="adyen_hpp_config_provider" xsi:type="object">Adyen\Payment\Model\Ui\AdyenHppConfigProvider</item>
                 <item name="adyen_boleto_config_provider" xsi:type="object">Adyen\Payment\Model\Ui\AdyenBoletoConfigProvider</item>
                 <item name="adyen_pos_cloud_config_provider" xsi:type="object">Adyen\Payment\Model\Ui\AdyenPosCloudConfigProvider</item>
-                <item name="adyen_apple_pay_config_provider" xsi:type="object">Adyen\Payment\Model\Ui\AdyenApplePayConfigProvider</item>
-                <item name="adyen_google_pay_config_provider" xsi:type="object">Adyen\Payment\Model\Ui\AdyenGooglePayConfigProvider</item>
-            </argument>
+                </argument>
```
msilvagarcia, line 36:
```suggestion
            </argument>
```
oh no.

msilvagarcia (approved):

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
cyattilakiss: Added the changes

acampos1916 (approved):

brightonmike:
@acampos1916 Noooooooo.

We built an integration and we were told to resolve Klarna issues by using the rc2 branch. We were not warned about this change being a part of that.

You have now added this code which breaks our Apple Pay and Google Pay headless integrations two weeks before launch of the site.

Please urgently response with suggested course of action.

acampos1916:
Hello @brightonmike indeed this release candidate removes the separate Apple and Google Pay, these are now supported by the checkout component like the rest of the payment methods. The headless integration would look about the same, just not using the Apple and Google separated payment methods.

It should be easier now to implement this, we will shortly share here documentation to make it clearer.

brightonmike:
@acampos1916 Hugely frustrating for us as we'd already built our headless Apple/Google pay integrations and were not aware they'd be switched for HPP. It also means we lose the ability to display them as express payment methods.

In the process of migrating I've found a bug whereby not passing `totalPriceLabel` as part of the Apple Pay configuration is causing Apple Pay not to display. It's a strange one as it displayed the first few times but then returned an error saying total label was not defined. Unsure if this should be coming through from paymentMethodsExtraDetails.

cyattilakiss:
Hi @brightonmike 

Have you implemented the Apple/Google pay as express payment methods or do you plan to do so?
It's a pretty valid feedback and that's the reason we kept the changes in beta so we can still change it if needed for PWA - thank you for this.

Can you use the rc.1 to resolve the issues you were facing or do you need the rc.2 for that?

Please let us discuss it internally if we can rollback the removal of these 2 payment methods and to assess what are the benefits, and if yes then what are the necessary steps to keep use of the centralised logic in the hpp.

Best,
Attila
Adyen

brightonmike:
@cyattilakiss 

Since this we've managed to replicate Google/Apple Pay as regular methods and Apple Pay as express.

To make it work as express, I'm having to submit a dummy address to obtain the Adyen payment methods data needed. I'm then using the Adyen Apple Pay events to trigger updates for shipping address/methods. I'm unsure if this is going to cause issues down the line with this approach but I can't see a way around it.

cyattilakiss:
Hi @brightonmike 

Thanks for the reply and sorry for the delayed message regarding this topic.
I am curious if you have managed to "replicate Google/Apple Pay as regular methods and Apple Pay as express" by using the adyen_hpp "payment method" wihtout the need of separate apple pay and google pay payment methods in the adyen plugin.

We see that indeed in case you want to use the latest checkout components then you need to do a retrieve-payment-methods call anyway in order to retrieve the configurations for the checkout component and to initiate these wallet payment methods correctly.

After investigating this approach we think it is a good approach to keep these wallet mathods in the adyen_hpp payment methods because with this approach
 - you don't need to do multiple retrieve-payment-methods call to the adyen plugin endpoint (only 1 which then provides the configuration for all the wallet methods that you want to use as express methods)
 - you don't need to handle the wallet payment methods as separate payment methods while the rest of the payment methods are handled in the adyen_hpp method
 - it is easier to handle the default non PWA checkout as well

I am curious about your opinion and also about your current experience with the express checkout using the adyen_hpp instead of the adyen_applepay method.

Thank you in advance,
Attila
Adyen

Issue #951, opened 2021-01-19
Store my details checkbox still available while Vault is disabled


Skullsneeze:
**Describe the bug**
When the vault (and stored payment methods are enabled) I expect that the "Store my details" checkbox for the creditcard payment method is no longer available. However this is currently not the case.

**To Reproduce**
1. Go to Adyen configuration in Magento admin
2. Click on the configure button for "CreditCard API integration"
3. Set the "Vault enabled" option to "No"
4. Save the configuration and go the the checkout payment page
5. Select the Adyen CC option, and you will still see the option to save your details

**Expected behavior**
With the Vault option disabled I expect there to be no option to store my payment details for the creditcard payment method.

**Magento version**
2.3.6

**Plugin version**
6.6.9

**Desktop (please complete the following information):**
 - OS: macOs Mojave 10.14.6
 - Browser: Chrome
 - Version: 87.0.4280.141

**Additional context**
Perhaps I am misunderstanding the configuration, in which case it might be worth considering updating the naming of certain options, or adding a more clear description.

Skullsneeze:
I've found that this is not exactly a bug, just that a different configuration on my end was needed. Just in case anyone find themselves facing this "issue", under "Advanced: Billing Agreements" you need to set the "Enable OneClick" option to "No".

Also I think I initially misunderstood the Vault functionality, which is what caused the confusion for me

PR #952, opened 2021-01-19
Updated check to disable storing of CC details when vault is disabled

Skullsneeze:
**Description**
When checking if it should be possible to store CC details we currently check if the vault is _**disabled**_. I would however expect that storing of payment details should only be possible if the vault option is _**enabled**_.

**Tested scenarios**
- Checkout flow

**Fixed issue**:  #951


Skullsneeze:
Misunderstanding of the config caused this issue. More details can be found in #951

PR #959, opened 2021-02-04
[PW-3708] Cloning quote with quote helper on failed payment

acampos1916:
**Description**
When a payment for an order is refused a new order is ready to be placed using the same quote. In this PR the previous quote is being cloned (item lines, qtys, product options, coupon code, addresses) and replaced in the session so the following order is attached to it.

**Tested scenarios**
* CC
    * Non-3DS: failed payment does not place order -> successful payment uses the same quote and places the order
    * 3DS1: failed payment places `payment_review` order using quote ID and clones quote -> successful payment uses the cloned quote and places the order
    * 3DS2: failed payment places `canceled` order using quote ID and clones quote -> successful payment uses the cloned quote and places the order
* Redirect PMs
    * e.g. IDeal: failed payment places `new` order using quote ID and clones quote -> successful payment uses the cloned quote and places the order

**Fixed issue**:  PW-3708


msilvagarcia (dismissed):

cyattilakiss (approved):

AlexandrosMor (approved):

msilvagarcia (approved):

acampos1916:
This is to be merged along with a feature toggle.

Controller/Process/Result.php
```diff
@@ -69,9 +68,9 @@ class Result extends \Magento\Framework\App\Action\Action
     protected $quoteFactory;
```
peterojo, line 68: I think you can remove `$quoteFactory` as well

Helper/Quote.php
```diff
@@ -0,0 +1,104 @@
...
+use Magento\Quote\Api\CartRepositoryInterface;
+use Magento\Quote\Model\Quote as QuoteModel;
+use Magento\Quote\Model\Quote\AddressFactory as QuoteAddressFactory;
+use Magento\Quote\Model\ResourceModel\Quote\Address as QuoteAddressResource;
+use Magento\Sales\Model\Order;
+
+class Quote
+{
+    /**
+     * @var CartRepositoryInterface
```
peterojo, line 40:
```suggestion
     * @var QuoteRepository
```

Controller/Process/Result.php
acampos1916: You're right!

peterojo (approved):

cyattilakiss (approved):

PR #960, opened 2021-02-05
[PW-4027] Shopper reference is too short for api v66

AlexandrosMor:
**Description**
Shopper reference was too short, less than 3 digits and therefore the tokens cannot be retrieved on the /paymentMethods call. It is updated so it is always 3 digits at least.

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Helper/Requests.php
+++ b/Helper/Requests.php
@@ -101,7 +101,7 @@ class Requests extends AbstractHelper
         $request = []
     ) {
         if ($customerId > 0) {
-            $request['shopperReference'] = $customerId;
+            $request['shopperReference'] = str_pad($customerId, 3, '0', STR_PAD_LEFT);
         }
         else {
             $uuid = Uuid::generateV4();
--- a/Helper/PaymentMethods.php
+++ b/Helper/PaymentMethods.php
@@ -178,7 +178,7 @@ class PaymentMethods extends AbstractHelper
                     $currencyCode
                 ),
             ],
-            "shopperReference" => $this->getCurrentShopperReference(),
+            "shopperReference" => str_pad($this->getCurrentShopperReference(), 3, '0', STR_PAD_LEFT),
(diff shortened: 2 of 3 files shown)


peterojo (approved):

Helper/Requests.php
```diff
@@ -101,7 +101,7 @@ public function buildCustomerData(
         $request = []
     ) {
         if ($customerId > 0) {
-            $request['shopperReference'] = $customerId;
+            $request['shopperReference'] = str_pad($customerId, 3, '0', STR_PAD_LEFT);
```
Aleffio, line 104: do we need something similar when we do the /paymentMethods call
https://github.com/Adyen/adyen-magento2/blob/4ec964b0a33b6ab1796392f6eec4c3d70b1aa153/Helper/PaymentMethods.php#L181

and when we save the cards for terminal api?
https://github.com/Adyen/adyen-magento2/blob/4ec964b0a33b6ab1796392f6eec4c3d70b1aa153/Model/AdyenInitiateTerminalApi.php#L263

AlexandrosMor: The payments method call does not throw an error from the api with single digit shopper reference but I am not sure if there is an issue having different shopper reference for payent methods call and payment. Any opinion @cyattilakiss ?.
The terminal api is not part of the Adyen api version 66 so I am sure it is ok.

Aleffio: /payments won't throw an error, but it will save the card with **shopperReference** "001" for example, while /paymentMethods will try to fetch cards with **shopperReference** "1", and won't find any.
Terminal API will save cards that will be used later for ecom, so I think the same should apply there.

AlexandrosMor: Ok you got a point 👍🏻

Aleffio (approved):

peterojo (approved):

PR #977, opened 2021-03-04
[PW-2089] Config to choose street line to be used as house number

acampos1916:
**Description**
The plugin gets the house number from the full street string based on a regex pattern if the country belongs to a list that requires these values to be separate. But this brought confusion about the format and street line fields to use as house number in the checkout, which is also affected by the Magento config `customer/address/street_lines`.

In this PR a new config is being introduced to select if the house number should be drawn from the full street string as described above or if there's a single street line field to be used as house number.

**Tested scenarios**
     * New config House Street Line = Draw from last street line: House number = regex pattern match on street string and the rest will be used as street name.
     * New config House Street Line = 1-4: House number = street line selected if there are enough `customer/address/street_lines` enabled, else regex pattern match on street string and the rest will be used as street name.

**Fixed issue**:  PW-2089


Helper/Address.php
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Checks if the house number needs to be sent to the Adyen API separately or as it is in the street field
+     *
+     * @param $country
+     * @return bool
+     */
+    private function isSeparateHouseNumberRequired($country)
+    {
+        $countryList = ["dk", "fi", "uk", "us", "se", "at", "no"];
```
cyattilakiss, line 117: separate house numbers are not necessarily required by country but also payment method. I think this list is only applicable for Klarna but for example afterpay require the houseNumberOrName filled with a correct data and not just N/A

I think the best approach would be to stop relying to this function and only use the new configuration that you created.

In case the config says use the street line 1-2-3-4 then simply use those as houseNumberOrName otherwise fall back to the regex and try to parse it.

Model/Config/Source/HouseNumberStreetLine.php
```diff
@@ -0,0 +1,43 @@
...
+
+class HouseNumberStreetLine implements OptionSourceInterface
+{
+    /**
+     * @return array
+     */
+    public function toOptionArray(): array
+    {
+        return [
+            ['value' => 0, 'label' => __('Draw from last street line')],
```
cyattilakiss, line 36: I think it would make sense to also add the use default address parser which would just fall back to the current logic. (also to keep the backwards compatibility so we can introduce this still in v6)

Test/Unit/Helper/AddressTest.php
```diff
@@ -0,0 +1,228 @@
...
+                        'house_number' => 'N/A'
+                    ]
+                ];
+
+                $i++;
+            }
+        }
+
+        return $config;
+    }
```
cyattilakiss, line 227: Huge 👍

Helper/Address.php
acampos1916: Sounds good! I'll remove this check then, and only rely on the config

Model/Config/Source/HouseNumberStreetLine.php
acampos1916: value=0 'Draw from last street line' is [using the default parser](https://github.com/Adyen/adyen-magento2/pull/977/files#diff-b74fba883fda9c3eff5c703a99610a2c6920b89e1f40dc6bb6f7e7c1d0e6af82R58) that was being used. And in config.xml this value is the default so that instances continue seeing the same results. Or do you mean something else?

Helper/Address.php
acampos1916: Done https://github.com/Adyen/adyen-magento2/pull/977/commits/b66a11fee9c716436419126a98aa994bf410e0a8

msilvagarcia (dismissed):
Would be nice to improve the deprecation messages with something more useful. What will replace those methods?

The files with methods being deprecated:
- Block/Redirect/Redirect.php
- Helper/Data.php

acampos1916:
> Would be nice to improve the deprecation messages with something more useful. What will replace those methods?
> 
> The files with methods being deprecated:
> 
> * Block/Redirect/Redirect.php
> * Helper/Data.php

@msilvagarcia good idea. I pushed better deprecation messages just now.

peterojo (approved):
lgtm

Model/Config/Source/HouseNumberStreetLine.php
cyattilakiss: sorry I thought this setting will just use the last enabled street line as house number or name but not actually use the parse. In case this  is the one falling back to the parser then it is indeed already working correctly :)

cyattilakiss (dismissed):
conflict

cyattilakiss (approved):

AlexandrosMor (approved):

Model/Config/Source/HouseNumberStreetLine.php
acampos1916: Indeed, the 'Draw from last street line' option uses the parser

PR #980, opened 2021-03-08
Add extra logging for invoice creation

msilvagarcia:
**Description**

To better understand why invoices are not being created we need to know the order state and order status.

**Tested scenarios**

None

**Fixed issue**:

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -1986,7 +1986,11 @@ class Cron
         } else {
             $this->_adyenLogger->addAdyenNotificationCronjob
             (
-                'It is not possible to create invoice for this order'
+                'It is not possible to create invoice for this order',
+                [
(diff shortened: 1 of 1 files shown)


Model/Cron.php
```diff
@@ -1986,7 +1986,11 @@ protected function _createInvoice()
         } else {
             $this->_adyenLogger->addAdyenNotificationCronjob
             (
-                'It is not possible to create invoice for this order'
+                'It is not possible to create invoice for this order',
+                [
```
cyattilakiss, line 1990: could you pls also log the orderId or orderNumber to make it easier to find the order?

acampos1916 (dismissed):

cyattilakiss (approved):

acampos1916 (approved):

Issue #912, opened 2020-12-15
Load error adyen.min.js (if has minify js enabled)


way2ecommerce:
**Description**
If Magento has minify enabled add "adyen.**min**.js" to this js external file "https://checkoutshopper-live.adyen.com/checkoutshopper/sdk/3.4.0/adyen.js"

**To Reproduce**
1. Go to 'checkout' with minify enabled on production mode
4. See error on firebug or developer console

walkwizus:
There are an update for this issue ?

AlexandrosMor:
Hello @way2ecommerce,

Thank you for raising this issue. Could you send us the adyen-magento2 plugin version that you are using ? This issue is fixed in the plugin version [6.6.0](https://github.com/Adyen/adyen-magento2/releases/tag/6.6.0)  

regards,
Alexandros

walkwizus:
I use the 6.6.8 and it's not fix.

walkwizus:
In production mode Magento call this URL: https://checkoutshopper-live.adyen.com/checkoutshopper/sdk/3.4.0/adyen.min.js As you can see there are a 404.

AlexandrosMor:
hello @walkwizus,

In order to exclude the adyen.js for minificaition we need to change the etc/config.xml but it is not compatible with magento 2.2. We introduce a fix compatible with 2.2 and 2.3 in the version 7 which is not released yet. You can fix it by following this [pull request](https://github.com/Adyen/adyen-magento2/pull/816/files) until the adyen plugin version 7 is released

Kind regards,
Alexandros

Issue #826, opened 2020-09-25
[PW-3111] Using Magento's base URL as 3DS1 returnUrl breaks redirect for PWA storefronts


acampos1916:
**Describe the bug**
#786 and #806 introduced an URL setting for PWA integrations to be used when the storefront's URL and Magento's base URL differ.

#821 uses `$this->urlBuilder->getUrl('adyen/process/redirect');` which does not consider the PWA URL setting, thus 3DS1 redirects fail.

**To Reproduce**
1. Use the plugin in Magento + PWA environment
2. Set a different frontend URL in PWA settings
3. Process a 3DS1 payment

**Expected behavior**
Customer is redirected to the frontend `adyen/process/redirect` endpoint so the GET response can be processed.

**Actual behavior**
Customer is redirected to the Magento `adyen/process/redirect` endpoint.

**Additional context**
Raised in #787

fahu:
What's the progress on this one? We're facing the same issue :(

acampos1916:
Hey @fahu,

Working on this issue is still in our short term roadmap but hasn’t been picked up yet unfortunately. Will notify here when this is done.

TonyRutherford1:
Hi @acampos1916 Any update to this issue? Also facing the same issue

edroberts:
@any updates on this? Over 6 months now since reported - it's not like 3DS1 has just 'disappaeared' from the internet and PWA adoption is only increasing - so unclear why there's no serious priority on this - other gateways can handle it

acampos1916:
Hello @edroberts this has already been fixed in #988 and is being released with #1001. I'm tagging the PR so that the release closes this issue.

Issue #456, opened 2019-06-11
[PW-2029] Charged currency is incorrect


gigadesign1:
**Magento version**: 2.3.1
**Plugin version**: 2.4.0
**Description**
When I configure Magento to display a local currency, this currency is sent to Adyen as the currency the customer has to complete the payment with.

This behaviour is different than other (built in) payment providers in Magento.
Magento itself also displays a notice that the currency that the customer will be charged for will be different:
![Schermafbeelding 2019-06-11 om 13 27 51](https://user-images.githubusercontent.com/1337445/59268587-50f33f80-8c4d-11e9-895f-4abf1b599f8e.png)
![Schermafbeelding 2019-06-11 om 13 28 39](https://user-images.githubusercontent.com/1337445/59268594-551f5d00-8c4d-11e9-8130-ca874a0f4a46.png)

If I complete this payment with a payment method provided by the Adyen plugin, the customer will be charged for £23.98 instead of €25.94

cyattilakiss:
Hi @gigadesign1 ,

Can you please show me (screenshot or just a path) where do you set the currency, or maybe even better if you can just show me how the currency is retrieved to be the one that you want it to be sent to Adyen. Currently we use the quote to retrieve the currency. `$this->quote->getCurrency();` in the CheckoutDataBuilder.php

Thank you in advance,
Attila
Adyen

gigadesign1:
Hi @cyattilakiss,

I think it should be `$this->quote->getBaseCurrency();` and `$this->quote->getBaseGrandTotal();`.

We have set `currency/options/base` to EUR and `currency/options/allow` to AUD,CAD,CHF,DKK,EUR,GBP,SEK,USD

According to the notice that Magento adds to the checkout, I would say that the transaction for this order should be in the base currency and not in the display currency.

cyattilakiss:
Hi @gigadesign1 ,

Let me discuss it internally, I will get back to you soon.

Best,
Attila
Adyen

cyattilakiss:
Hi @gigadesign1 ,

Currently the plugin uses only the display currency and our current merchants are using the plugin as it is. Although we understand that there are exceptions and merchants who wants to charge their customers in the base currency because of the accounting or other financial reasons.

The solution could be a setting in the magento admin where merchants can indicate which currency they want to use during the api calls to Adyen. 
[...]

PR #1001, opened 2021-04-06
7.0.0 release

cyattilakiss:
### Important notes
This is a major release and may introduce backwards incompatibility changes therefore when updating to this version please check carefully the changes that might affect your existing configurations and own customisations.

Documentation is being updated on our [docs page](https://docs.adyen.com/plugins/magento-2)
Extensive [PWA documentation](https://docs.adyen.com/plugins/magento-2/magento-pwa-storefront) is going to be updated soon.

### Description
This release introduces the following major changes and features :
 - Update Checkout API to version 67
   - The default behaviour for this new version of the API is to switch the 3DS1 redirect to the Magento returnUrl from the issuer page from POST to GET request, therefore the transparent redirect controller is not needed anymore and has been removed. In case you have already contacted our support to enable a specific account data property to achieve this behaviour on the plugin version 6 please let them know so they can remove this setting from your merchant account
 - Update Checkout component to version 3.23.0
   - New client key configuration has been introduced in the admin configuration panel as a new required setting. See [docs](https://docs.adyen.com/development-resources/client-side-authentication#get-your-client-key) to see how to retrieve the client key.
   - Deprecated Origin key since it has been replaced with the client key. The old endpoint to retrieve the origin key is still in place but going to be removed.
 - Added PWA support for cards (non 3D, 3DS1 and 3DS2) and local payment methods as well. In order to be able to support PWA integrations we introduced (or adjusted) some endpoints exposed by the plugin like the `/retrieve-adyen-payment-methods`, `/payment-status`, `/adyen/paymentDetails` endpoints (see more about them later in the release notes or in the PWA documentation as you can see at the beginning of this release)
 - Google pay and Apple pay has been removed as standalone Magento payment method and got merged into the alternative payment methods group. (In case you would like to use these payment methods as express payment methods please use the adyen_hpp payment method and the correct `brand`)
 - Paypal popup flow is implemented instead of redirect
 - Charged currency feature toggle (see more below)
 - Generic implementation of Adyen's Web Components (see more below)
 - New recurring settings (see more below)

Since there are migration scripts added to the release as well **please make sure that you run your upgrade command** when deploying the changes.

### New recurring settings
A new configuration is introduced for the recurring settings: 

The `Enable Recurring` and the `Enable OneClick` values in the plugin configuration are replaced with `Vault Enabled` and `Enable OneClick`.

Those values can be found under the `Stored Payment Methods - Requires Adyen Credit Card` in the plugin config and should be used as follows: 

Set the following settings if you would like to use
- **Vault**
  - for OneClick payments (customer can use it in the checkout) set the `Vault Enabled` to Yes and `Enable OneClick` to Yes and in the Customer Area set the `Enable Recurring` to true in the Account > Checkout settings.
  - for Recurring payments (subscriptions / backend orders) set the `Vault Enabled` to Yes and `Enable OneClick` to No and in the Customer Area set the `Enable Recurring` to true in the Account > Checkout settings.
- **Billing agreements**
  - for OneClick payments (customer can use it in the checkout) set the `Vault Enabled` to No and `Enable OneClick` to Yes and in the Customer Area set the `Enable Recurring` to false in the Account > Checkout settings.
  - for Recurring payments (subscriptions / backend orders) set the `Vault Enabled` to No and `Enable OneClick` to No and in the Customer Area set the `Enable Recurring` to true in the Account > Checkout settings.

Please check the [Adyen's docs](https://docs.adyen.com/plugins/magento-2/stored-payment-methods) for more information.

### Charged currency feature toggle
A new configuration is introduced to charge orders based on storeview or global currency
 - The plugin was designed to charge the order amount using the storeview or display currency. We're introducing a new configuration to also select the Global/Website or base currency to process payments. display remains as the default option.
 - A new column has been introduced for the sales_order table to store the order's currency based on this new configuration at the time of order placement, with a new event observer to populate it. This has been done so refunds can be processed with the correct currency.
 - To ensure that payments are processed in the expected currency, this config value has been hidden and can only be changed by extending the plugin's system.xml or by using the Magento CLI: `bin/magento config:set payment/adyen_abstract/charged_currency (base|display)`

**The possible values are:**

 - Display currency – The store view or display currency is used. This is the default setting.
 - Global/Website currency – The base currency is used. Depending on how Magento is configured, this will be either the Global or the Website currency.

You can find more about this in the [docs](https://docs.adyen.com/plugins/magento-2/set-up-the-plugin-in-magento#selecting-charged-currency)

### Generic implementation of Adyen's Web Components
Most of the frontend logic to render the payment methods and process callbacks has been replaced with a generic flow that handles all supported payment methods. The dependency on the external resource `adyen.js` has been replaced for a local `bundle.js` file. The component's result `data` will include most required parameters to execute API calls and the manual configuration of these requests has been removed.

### Payment Details API endpoint
`/V1/adyen/threeDS2Process`->`/V1/adyen/paymentDetails`: AdyenThreeDS2Process.php has been replaced with AdyenPaymentDetails.php. This new endpoint is used from the frontend to process a payload with the web component's result and the `orderId` in the following format:
```
{
    "details": {
        "threeds2.fingerprint": "..."
    },
    "paymentData": "...",
    "orderId": "..."
}
```
This will trigger a `/paymentDetails` call and the handling of the response. 
The response value looks like this:
```
{
    "action": {
        ...
    },
    "additionalData": {
        ...
    },
    "isFinal": boolean,
    "resultCode": "..."
}
```
(`action` and `additionalData` are optional)

Depending on the result code the order will be placed, cancelled, or the [additional action](https://docs.adyen.com/checkout/components-web#step-4-additional-front-end) will be returned to the frontend to execute `createFromAction` or `handleAction`

PWA Implementations can use this endpoint to process payment details calls.

**Tested scenarios**
 - E2E tests

Unique cases:
 - Charged currency
 - Instalments for one click and card payments
 - Google pay 3DS1 fallback

**Confirmed payment methods**
Starting with this release we are going to list all the payment methods that we have confirmed during the release process. Since Adyen supports more than 250 payment methods we are not going to confirm each and every one but only the most used ones.

**Card payment methods**
 - Card payment method (non 3D, 3DS1 and 3DS2)
 - One click payment methods (using billing agreement or Vault)
 - Recurring payment methods (using billing agreement or Vault)

**Local payment methods**
 - Afterpay
 - Apple pay
 - Bancontact
 - Bantransfer iban
 - Doku
 - Google pay
 - iDeal
 - Klarna
 - Oney
 - Zip
 - Multibanco
 - Konbini
 - Swish
 - MBway
 - Twint
 - Paypal
 - Sepa direct debit
 - PIX
 - Ratepay
 - Oxxo
 - Econtexts

### Known issues: 
_We are going to extend the list as soon as any issue got reported regarding this release._

**Fixed issue**:  
#456 [PW-2029] Charged currency is incorrect
#787 [PW-2871] 3DS1 in PWA 
#826 [PW-3111] Using Magento's base URL as 3DS1 returnUrl breaks redirect for PWA storefronts
#912 Load error adyen.min.js (if has minify js enabled)


Controller/Process/Result.php
```diff
@@ -450,10 +475,26 @@ protected function validatePayloadAndReturnResponse($response)
...
+                    $this->_order = $order;
+                } else {
+                    // TODO error handling
+                    $this->_adyenLogger->addError("Wrong merchantReference was set in the query or in the session");
+                    // TODO error page
+                }
+            } else {
+                // TODO error handling
+                $this->_adyenLogger->addError("No merchantReference in the response");
+                // TODO error page
```
cyattilakiss, line 496:
```suggestion
                $this->_adyenLogger->addError("No merchantReference in the response");
                $response['error'] = 'merchantReference is missing from the response';
```

Controller/Process/Result.php
```diff
@@ -450,10 +475,26 @@ protected function validatePayloadAndReturnResponse($response)
...
+            $responseMerchantReference = !empty($response['merchantReference']) ? $response['merchantReference'] : null;
+            $resultMerchantReference = !empty($result['merchantReference']) ? $result['merchantReference'] : null;
+            $merchantReference = $responseMerchantReference ?: $resultMerchantReference;
+            if ($merchantReference) {
+                if ($order->getIncrementId() === $merchantReference) {
+                    $this->_order = $order;
+                } else {
+                    // TODO error handling
+                    $this->_adyenLogger->addError("Wrong merchantReference was set in the query or in the session");
+                    // TODO error page
```
cyattilakiss, line 491:
```suggestion
                    $this->_adyenLogger->addError("Wrong merchantReference was set in the query or in the session");
                    $response['error'] = 'merchantReference mismatch';
```

Controller/Process/Result.php
```diff
@@ -405,42 +409,63 @@ protected function _getOrder($incrementId = null)
...
             }
         } else {
-            $this->_adyenLogger->addError("Can't load the order id from the session");
+            $this->_adyenLogger->addError("Payment object cannot be loaded from order");
         }
 
-        $request["details"] = $response;
-
-        if (!empty($this->_session->getLastRealOrder()) &&
+        // TODO check if this is ever reached
```
cyattilakiss, line 467: let's remove it

Gateway/Validator/CheckoutResponseValidator.php
```diff
@@ -67,28 +68,36 @@ public function validate(array $validationSubject)
...
-                    $payment->setAdditionalInformation('threeDSType', $resultCode);
-                    $payment->setAdditionalInformation(
-                        'threeDS2Token',
-                        $response['authentication']['threeds2.challengeToken']
-                    );
-                    $payment->setAdditionalInformation('adyenPaymentData', $response['paymentData']);
-                    break;
                 case "Authorised":
                 case "Received":
+                    // TODO refactor since the full additionalData is stored in additionalInformation already
```
cyattilakiss, line 100: @acampos1916 I think since we now use the component instead of using these additional data it can be removed from here as well

acampos1916: Looks like it could be removed indeed. But maybe the setCcType should be kept

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -189,148 +160,93 @@ define(
...
-                        'number': this.creditCardNumber(),
-                        'cvc': this.securityCode(),
-                        'expiryMonth': this.expiryMonth(),
-                        'expiryYear': this.expiryYear(),
-                        'holderName': this.creditCardOwner(),
-                        'store_cc': this.storeCc,
+                        'combo_card_type': this.comboCardOption(),
                         //This is required by magento to store the token
                         'is_active_payment_token_enabler' : this.storeCc,
+                        //TODO check installments
```
cyattilakiss, line 236:
```suggestion
```

Gateway/Validator/CheckoutResponseValidator.php
cyattilakiss: Added it to the ticket PW-4496

Controller/Process/Result.php
AlexandrosMor: Done

Gateway/Validator/CheckoutResponseValidator.php
```diff
@@ -67,28 +68,36 @@ public function validate(array $validationSubject)
         $isValid = true;
         $errorMessages = [];
         $resultCode = array_key_exists("resultCode", $response) ? $response['resultCode']:'';
```
cyattilakiss, line 70: Unused var

```suggestion
```

view/frontend/web/js/view/payment/adyen-methods.js
```diff
@@ -48,33 +56,34 @@ define(
...
+                    shippingAddressCountry = quote.shippingAddress().countryId;
+                    fullScreenLoader.startLoader();
+                    // Retrieve adyen payment methods
+                    adyenPaymentService.retrievePaymentMethods().done(function(paymentMethods) {
+                        paymentMethods = JSON.parse(paymentMethods);
+                        adyenPaymentService.setPaymentMethods(paymentMethods);
+                        fullScreenLoader.stopLoader();
+                    }).fail(function() {
+
+                    })
```
cyattilakiss, line 85:
```suggestion
                    }).fail(function() {
                        console.log('Fetching the payment methods failed!');
                    })
```

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -25,92 +25,75 @@ define(
...
-                        'expiryMonth',
-                        'expiryYear',
-                        'installment',
-                        'installments',
-                        'creditCardDetailsValid',
-                        'placeOrderAllowed'
-                    ]);
+            initObservable: function() {
+                this._super().observe([
+                    'creditCardType', // TODO check if still needed
```
cyattilakiss, line 92:
```suggestion
                    'creditCardType',
```

AlexandrosMor (approved):

acampos1916 (approved):

PR #1005, opened 2021-04-08
extract isAutoCapture logic to separate class

danslo:
**Description**

In one of our recent Adyen implementations, there is a need to change auto-capturing dynamically.

Some orders will be partially financed with the deposit being paid through Adyen, and capture/cancellation will be determined later based on their eligibility for financing.

To facilitate this I have introduced a new class `Adyen\Payment\Model\Cron\IsAutoCapture`. This contains the same logic that was previously in `Adyen\Payment\Model\Cron::_isAutoCapture` method. Some cleanup was done to the body of this method.

`Adyen\Payment\Model\Cron` is mostly unchanged, except for calling into this new class and ordering some class imports.

I also introduced a `ManualCaptureMethods` config source, to make things even cleaner.

What this now allows us to do is create plugins on `IsAutoCapture`, and implement any logic required.

I personally think that `Cron` class has become a bit of a monstrosity (2000+ LOC and ~30 dependencies) and could benefit from more composition, but I wanted to keep scope for this PR small to gauge your thoughts. Opening as a draft PR to discuss.

### DIFF
--- a/Model/Config/Source/ManualCaptureMethods.php
+++ b/Model/Config/Source/ManualCaptureMethods.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ *                       ######
+ *                       ######
+ * ############    ####( ######  #####. ######  ############   ############
+ * #############  #####( ######  #####. ######  #############  #############
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen Payment module (https://www.adyen.com/)
+ *
+ * Copyright (c) 2015 Adyen BV (https://www.adyen.com/)
+ * See LICENSE.txt for license details.
+ *
+ * Author: Adyen <magento@adyen.com>
+ */
+
+declare(strict_types=1);
+
+namespace Adyen\Payment\Model\Config\Source;
+
+use Magento\Framework\Data\OptionSourceInterface;
+
+class ManualCaptureMethods implements OptionSourceInterface
+{
+    public function toOptionArray(): array
+    {
+        return [
+            ['value' => 'cup',               'label' => 'China Union Pay'],
+            ['value' => 'cartebancaire',     'label' => 'Carte Bancaire'],
+            ['value' => 'visa',              'label' => 'VISA'],
+            ['value' => 'visadankort',       'label' => 'VISA Dankort'],
+            ['value' => 'mc',                'label' => 'Mastercard'],
+            ['value' => 'uatp',              'label' => 'Universal Air Travel Plan'],
+            ['value' => 'amex',              'label' => 'American Express'],
+            ['value' => 'maestro',           'label' => 'Maestro'],
+            ['value' => 'maestrouk',         'label' => 'Maestro UK'],
+            ['value' => 'diners',            'label' => 'Diners Club'],
+            ['value' => 'discover',          'label' => 'Discover'],
+            ['value' => 'jcb',               'label' => 'JCB'],
+            ['value' => 'laser',             'label' => 'Laser'],
+            ['value' => 'paypal',            'label' => 'PayPal'],
+            ['value' => 'sepadirectdebit',   'label' => 'SEPA Direct Debit'],
+            ['value' => 'dankort',           'label' => 'Dankort'],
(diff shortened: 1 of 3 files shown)


danslo:
Apparently I cannot add reviewers to draft PRs, so I'll tag a couple of people. @acampos1916 @msilvagarcia @AlexandrosMor @rikterbeek

acampos1916:
@danslo thank you for this contribution, we'll review and test it and will probably have some stuff to discuss here. I'm adding the reviewers.

Any reason why it is still in draft? Are you still thinking of committing code?

danslo:
Honestly regarding the draft, I didn't want to risk it getting merged because:
- We haven't finished implementation and proper testing on this feature yet.
- This might benefit from some automated tests.
- But maybe most importantly; to get some quick feedback if this is the way to go, or if you guys have a better idea (not very familiar with the codebase).

Also the CI Builds failed though I don't think that's related to this PR.

Model/Config/Source/ManualCaptureMethods.php
```diff
@@ -0,0 +1,61 @@
...
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen Payment module (https://www.adyen.com/)
+ *
+ * Copyright (c) 2015 Adyen BV (https://www.adyen.com/)
```
acampos1916, line 18:
```suggestion
 * Copyright (c) 2021 Adyen BV (https://www.adyen.com/)
```

Model/Cron/IsAutoCapture.php
```diff
@@ -0,0 +1,198 @@
...
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen Payment module (https://www.adyen.com/)
+ *
+ * Copyright (c) 2015 Adyen BV (https://www.adyen.com/)
```
acampos1916, line 18:
```suggestion
 * Copyright (c) 2021 Adyen BV (https://www.adyen.com/)
```

Model/Config/Source/ManualCaptureMethods.php
```diff
@@ -0,0 +1,61 @@
...
+            ['value' => 'sepadirectdebit',   'label' => 'SEPA Direct Debit'],
+            ['value' => 'dankort',           'label' => 'Dankort'],
+            ['value' => 'elo',               'label' => 'Elo'],
+            ['value' => 'hipercard',         'label' => 'Hipercard'],
+            ['value' => 'mc_applepay',       'label' => 'Mastercard Apple Pay'],
+            ['value' => 'visa_applepay',     'label' => 'VISA Apple Pay'],
+            ['value' => 'amex_applepay',     'label' => 'American Express Apple Pay'],
+            ['value' => 'discover_applepay', 'label' => 'Discover Apple Pay'],
+            ['value' => 'maestro_applepay',  'label' => 'Maestro Apple Pay'],
+            ['value' => 'paywithgoogle',     'label' => 'Google Pay'],
```
acampos1916, line 58: The labels are not really necessary, right? As part of some more refactoring we're working on this could be useful so we might move it elsewhere at some point. But there's no problem if we keep it here for now

acampos1916:
> * But maybe most importantly; to get some quick feedback if this is the way to go, or if you guys have a better idea (not very familiar with the codebase).

This looks good to me @danslo. We're currently planning the cron class refactoring so moving this logic out is a good idea. Have you had time to implement and test this?

Model/Config/Source/ManualCaptureMethods.php
danslo: Weren't necessary for this PR, but moving these settings to a magento "config source" made sense, and consumers of that expect arrays of [value, label] object. Also means that it can easily be made configurable via system.xml in the future as a bonus

danslo:
@acampos1916 
> This looks good to me @danslo. We're currently planning the cron class refactoring so moving this logic out is a good idea. Have you had time to implement and test this?

We're currently still preparing for the feature to go live, but we have been doing quite a bit of testing around setting manual/auto mode via plugins, and it is working properly. It must be said that we currently only use creditcard payments so please don't take that as an absolute guarantee that I didn't break anything else in the process 😬

cyattilakiss (approved):

msilvagarcia (commented):
I think this is stale. Should we close it?

tnaber:
Let's close this as it has become stale. @danslo you can reopen this if you want te continue work on this :)

Issue #783, opened 2020-08-04
PayByMail : skin code is required otherwise HMAC error due to fallback to inexistant settings


rbayet:
**Describe the bug**
It's mandatory to add a skin code parameter to the PayByMail configuration in an admin, otherwise the code will try to pull the skin_code and hmac_test/hmac_live settings from the HPP section.
Problem is, those settings do no exist / no longer exist in the HPP section.

**To Reproduce**
1. Make sure that you have a notification_hmac_key_test/notification_hmac_key_live settings in Stores > Configuration > ... > Adyen ... > Advanced Security (because, why not ?)
2. In Stores > Configuration > ... > Adyen ..., open the Pay by Mail integration configuration screen
3. Copy/Paste your global hmac_test/hmac_live settings into the local hmac_test/hmac_live settings (because why not ? the fields help hint is only "Should match with the HMAC test/live key defined in the Adyen Customer Area")
4. Enable the method
5. Clear the cache
6. Try to create an order from admin and select Adyen Pay By Mail

**Expected behavior**
The order is placed and the customer receives an email with a Payment link.

**Actuel behavior**
The order is not placed and the error message `Order saving error: You did not provide a HMAC key` is displayed.

**Screenshots**
![image](https://user-images.githubusercontent.com/2353374/89290399-191d3780-d659-11ea-8613-ff2be6b634a4.png)

![image](https://user-images.githubusercontent.com/2353374/89290456-2d613480-d659-11ea-872a-ae5e700912f2.png)

**Additional context**
Adyen magento module version : 6.3.0
Magento 2 version : Magento Commerce 2.3.4

**Possible problem origin**
The \Adyen\Payment\Gateway\Command\PayByMailCommand::getFormFields conditionnally pull inexistant settings from the HPP system.xml section 
```
    /**
     * @param \Magento\Sales\Model\Order\Payment $payment
     * @param float|bool $paymentAmount
     * @return array
     */
    protected function getFormFields($payment, $paymentAmount = false)
    {
        $order = $payment->getOrder();
[...]
        // check if paybymail has it's own skin
        $skinCode = trim($this->_adyenHelper->getAdyenPayByMailConfigData('skin_code'));
        if ($skinCode == "") {
            // use HPP skin and HMAC
            $skinCode = $this->_adyenHelper->getAdyenHppConfigData('skin_code');
            $hmacKey = $this->_adyenHelper->getHmac();
[...]

PR #1013, opened 2021-04-16
[PW-4502] Pay by Mail does not use store ID's and uses the default configuration

ataberkylmz:
**Description**
If you have multiple store views in Magento and want to use Pay By Mail in a certain store view, it was not possible as the function getHmacPayByMail() were not accepting store ids as an argument so it always tries to get the default config regardless of the store view selected.

**Tested scenarios**
- [x]  Payment with differently configured Store View
- [x]  Payment without a skin code (legacy code)

**Fixed issue**:
PW-4502 - Pay by Mail does not use store ID's and uses the default configuration


Gateway/Command/PayByMailCommand.php
```diff
@@ -136,18 +136,8 @@ protected function getFormFields($payment, $paymentAmount = false)
         $realOrderId = $order->getRealOrderId();
         $storeId = $order->getStore()->getId();
 
-        // check if paybymail has it's own skin
-        $skinCode = trim($this->_adyenHelper->getAdyenPayByMailConfigData('skin_code'));
-        if ($skinCode == "") {
-            // use HPP skin and HMAC
-            $skinCode = $this->_adyenHelper->getAdyenHppConfigData('skin_code');
```
acampos1916, line 143: @ataberkylmz I think the adyen_hpp skin_code was removed from the config and that's why you're removing this, right?

Gateway/Command/PayByMailCommand.php
```diff
@@ -163,20 +153,20 @@ protected function getFormFields($payment, $paymentAmount = false)
...
-            } else {
-                $countryCode = "";
-            }
+        if (is_object($order->getBillingAddress()) && $order->getBillingAddress()->getCountryId() != "") {
+            $countryCode = $order->getBillingAddress()->getCountryId();
+        } else {
+            $countryCode = "";
         }
 
-        $deliveryDays = $this->_adyenHelper->getAdyenHppConfigData('delivery_days', $storeId);
```
acampos1916, line 178: Same here? The config was removed?

etc/adminhtml/system/adyen_pay_by_mail.xml
```diff
@@ -47,18 +47,23 @@
             <tooltip>How many days do you want this payment page to be active</tooltip>
             <config_path>payment/adyen_pay_by_mail/session_validity</config_path>
         </field>
-        <field id="skin_code" translate="label" type="text" sortOrder="50" showInDefault="1" showInWebsite="1" showInStore="1">
+        <field id="ship_before" translate="label" type="text" sortOrder="50" showInDefault="1" showInWebsite="1" showInStore="1">
+            <label>Ship Before Date</label>
+            <tooltip>The date within which the ordered goods or services need to be shipped or provided to the shopper</tooltip>
```
acampos1916, line 52: I see this config is type "text" but the tooltip says it is a date, the expected value is the number of days right? In that case I would change the label to "Ship Before Date (Days)" to keep the same name as in the docs. The Tooltip can also say something like "Days after the the Pay by Mail is generated within which the ordered goods or services..."

Also, let's add a number validation like in line 42 of this file

etc/config.xml
```diff
@@ -149,6 +149,7 @@
                 <allowspecific>0</allowspecific>
                 <sort_order>7</sort_order>
                 <session_validity>3</session_validity>
+                <ship_before>5</ship_before>
```
acampos1916, line 152: Why is the default 5? Do we need to mention it in the admin form?

Gateway/Command/PayByMailCommand.php
ataberkylmz: @acampos1916 Indeed! adyen_hpp/skin_code did not exist anymore so having the check there does not make sense. Same goes for the adyen_hpp/delivery_days.

Gateway/Command/PayByMailCommand.php
ataberkylmz: @acampos1916 Indeed the same as above, it did not exist anymore since the 4.x.x (not sure about the exact version maybe 4.2.0).

Gateway/Command/PayByMailCommand.php
acampos1916: Great. So if this is not being used anywhere else maybe you can check if this issue is solved or if something extra is needed https://github.com/Adyen/adyen-magento2/issues/783

etc/config.xml
ataberkylmz: @acampos1916 It was also defaulting to 5 days before the field as the previous HPP config was removed. However, this field is used to calculate the expiry date for the offline payment methods such as 7-Eleven or Boleto. Other than that, it is not a mandatory field.

Gateway/Command/PayByMailCommand.php
ataberkylmz: It is resolved to some extend by removing the non-existent configs, however if you leave it empty, you'll still get that error. We can use `validate-no-empty` validator but this would force merchants to enter a value there so I'd really recommend it.

etc/adminhtml/system/adyen_pay_by_mail.xml
ataberkylmz: Added the validator to ship_before and also session_validity, too.

msilvagarcia (approved):

acampos1916 (approved):

PR #1016, opened 2021-04-20
[PW-4538] Using type-brand structure for giftcards

### LINKED ISSUE #1011: [PW-4537] Giftcards unusable in v6.6.9 - v7.0.0
**Describe the bug**
We have recently updated the Adyen extension (Magento 2) for one of our clients. Our client was using an older version: 4.2.1.

Using giftcards to pay for an order no longer works since the update. We have tried the following versions of this repository: 6.6.9, 6.7.0 and 7.0.0.

In 6.6.9 and 6.7.0: The user receives the following error: "Error with payment method please select different payment method.". Adyen API returns "Invalid card number".

In 7.0.0: There are some input fields for your giftcard code but none of them is functional. Opening one of the giftcard payment methods opens _all_ giftcard payment methods. None of the payment methods (also non-giftcard methods!) work anymore as soon as the user selected one of the giftcard payment method.

**To Reproduce**

_Versions 6.6.9, 6.7.0 and 6.7.0_
1. Go to the backoffice of Adyen and enable all of the following giftcard payment methods: "vvvcadeaubon", "genericgiftcard", "boekenbon".
2. Go to your Magento 2 checkout. Click on one of the giftcard payment methods you have just enabled and click "Place order".
3. An error is visible on the frontend: "Error with payment method please select different payment method."
4. The response from the Adyen API is "Invalid card number".

_Additional steps for version 7.0.0 only_
- Clicking on one of the giftcard payment methods (step 2) opens all of the giftcard payment methods instead of just the one that you have clicked. See screenshot attached.
- The visible input fields are not usable and you can not use the 'Place Order' button because jQuery could not validate the form.
- None of the payment methods (also non-gifcard methods!) work anymore as soon as you have opened one (which opens them all) of the giftcard payment methods.

**Expected behavior**

_For versions 6.6.9 and 6.7.0_
Being forwarded to a secure payment page.

_For version 7.0.0_
Functional input fields to pay with one of the giftcards. 

**Magento version**
Tested on both 2.3.5-p2 and 2.4.2.

**Plugin version**
6.6.9, 6.7.0 and 7.0.0.

**Screenshots**

_Version 6.6.9 and 6.7.0_
On Chrome Version 89.0.4389.82, Ubuntu 20.04.
image

_Version 7.0.0_
On Chrome Version 89.0.4389.82, Ubuntu 20.04.
screenshot-mag development-2021 04 14-16_21_05

acampos1916:
**Description**
In Checkout API v65 giftcards now include the `brand` parameter as an identifier for the type of payment method and this requires some adjustments to the following:

- The adyen_hpp JS component renderer now checks for `brandMethods` so that the `type` and `brand` are assigned to `methodGroup` and `methodIdentifier`. Regular methods assign the `type` to both.
- The adyen_hpp HTML template now uses the `methodIdentifier` to render the methods forms.
- The /payments/details call in the Result controller validates the params from the Open API specs of v67, previously a `type` param was being sent that rejected the calls for giftcards.
- The type-brand methods are now checked to get the specific logo instead of the group (e.g. vvvcadeaubon instead of giftcard).

**Tested scenarios**
E2E tests
Giftcard payments with multiple brands enabled
HPP payments

**Fixed issue**:  PW-4538 fixes #1011

### DIFF
--- a/Helper/PaymentMethods.php
+++ b/Helper/PaymentMethods.php
@@ -30,6 +30,11 @@ use Magento\Framework\App\Helper\AbstractHelper;
  */
 class PaymentMethods extends AbstractHelper
 {
+
+    const METHODS_WITH_BRAND_LOGO = [
+        "giftcard"
+    ];
+
     /**
      * @var \Magento\Quote\Api\CartRepositoryInterface
      */
@@ -183,7 +188,10 @@ class PaymentMethods extends AbstractHelper
         // Add extra details per payment method
         $paymentMethodsExtraDetails = [];
         $paymentMethodsExtraDetails = $this->showLogosPaymentMethods($paymentMethods, $paymentMethodsExtraDetails);
-        $paymentMethodsExtraDetails = $this->addExtraConfigurationToPaymentMethods($paymentMethods, $paymentMethodsExtraDetails);
+        $paymentMethodsExtraDetails = $this->addExtraConfigurationToPaymentMethods(
+            $paymentMethods,
+            $paymentMethodsExtraDetails
+        );
         $response['paymentMethodsExtraDetails'] = $paymentMethodsExtraDetails;
 
         //TODO this should be the implemented with an interface
@@ -389,7 +397,9 @@ class PaymentMethods extends AbstractHelper
             );
 
             foreach ($paymentMethods as $paymentMethod) {
-                $paymentMethodCode = $paymentMethod['type'];
+                $paymentMethodCode = array_search($paymentMethod['type'], self::METHODS_WITH_BRAND_LOGO) !== false
+                    ? $paymentMethod['brand']
+                    : $paymentMethod['type'];
 
                 $asset = $this->assetRepo->createAsset(
                     'Adyen_Payment::images/logos/' .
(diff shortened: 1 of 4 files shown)


Helper/PaymentMethods.php
```diff
@@ -389,7 +397,9 @@ protected function showLogosPaymentMethods($paymentMethods, array $paymentMethod
             );
 
             foreach ($paymentMethods as $paymentMethod) {
-                $paymentMethodCode = $paymentMethod['type'];
+                $paymentMethodCode = array_search($paymentMethod['type'], self::METHODS_WITH_BRAND_LOGO) !== false
```
peterojo, line 400: Not super important but I think `In_array` is a cleaner way to do this
```suggestion
                $paymentMethodCode = in_array($paymentMethod['type'], self::METHODS_WITH_BRAND_LOGO)
```

AlexandrosMor (dismissed):

AlexandrosMor:
@acampos1916 Could you check the code smell ?

peterojo (approved):

AlexandrosMor (approved):

Issue #1014, opened 2021-04-16
[PW-4572] Issue with notifications skipped in shouldSkipProcessingNotification()


ebaschiera:
**Describe the bug**
Hi,
at line https://github.com/Adyen/adyen-magento2/blob/7.0.0/Model/Cron.php#L419 you should not use the "`->i`" to calculate how many minutes difference there is between the two DateTime objects. Because the "i" attribute doesn't take into account the hours or even days in DateInterval.

Let's see that issue with an example. Let's say that for some reason the `adyen_payment_process_notification` cron job  doesn't run for one day. We will have an unprocessed notification from a day before.

That code will run like this:
```
$offerClosedMinDate = new \DateTime('-10 minutes'); // 2021-04-16 16:40:00 (Timezone: CEST)
$createdAt = \DateTime::createFromFormat('Y-m-d H:i:s', $notification['created_at']); // 2021-04-15 16:45:00
$minutesUntilProcessing = $createdAt->diff($offerClosedMinDate)->i; //5
```

So the notification will be skipped even if it's from a day before.

I suggest to convert those DateTime objects into seconds and make an arithmetic substraction, like this:
`$minutesUntilProcessing = ($createdAt->format('U') - $offerClosedMinDate->format('U')) / 60; //-1445`

We had several notifications stuck because the cron job was configured to run every 5 minutes (for system load mitigation purpose) and unfortunately that age check was failing consistently.

Thanks!

acampos1916:
Hello @ebaschiera thanks for opening this issue. In production mode the [recommended cron frequency is 1min](https://docs.adyen.com/plugins/magento-2/set-up-the-plugin-in-magento#step-2-run-cron) but makes sense that this shouldn't prevent the old notifications from being processed. Opening an internal ticket to test your solution.

zakisaad:
Dropping in as we also experience issues with this on a cron-heavy site, and have contemplated adding in code similar to the suggested by OP.

Will deal with it for now, what are the chances a fix for this will drop in the 7.0.1 release? We are holding off moving from 6.x.x until the first 7.x.x point release, would be ideal if it also included a fix for this!

TIA

acampos1916:
Hi @zakisaad, working to include this in 7.0.1 ASAP 👍

PR #1022, opened 2021-05-03
Support for Blik

acampos1916:
**Description**
The payment method Blik doesn't return details in the paymentMethods object, so instead of checking for this object we're attempting to create and mount all HPP payment methods instead. If there isn't a Web Component the error is handled by the try/catch.

Unhappy flows for Blik send a AUTHORISATION success=false notification that was being ignored in some scenarios during cron run, same as #1014 

**Tested scenarios**
Happy and unhappy flows for Blik
ShowPayButton methods like Paypal
Details methods like iDeal

**Fixed issue**:  PW-4528

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -445,7 +445,7 @@ class Cron
                 $authorisationSuccessFalseMinDate = new \DateTime('-10 minutes');
                 $createdAt = \DateTime::createFromFormat('Y-m-d H:i:s', $notification['created_at']);
 
-                $minutesUntilProcessing = $createdAt->diff($authorisationSuccessFalseMinDate)->i;
+                $minutesUntilProcessing = ($createdAt->format('U') - $authorisationSuccessFalseMinDate->format('U')) / 60;
 
                 if ($minutesUntilProcessing > 0) {
                     $this->_adyenLogger->addAdyenNotificationCronjob(
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
@@ -244,13 +244,6 @@ define(
                                     showPayButton = true;
                                 }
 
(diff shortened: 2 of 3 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -359,8 +352,6 @@ define(
                                             component.mount(containerId);
                                         }).catch(e => {
                                             result.isAvailable(false);
-                                            console.log(paymentMethod.methodIdentifier +
-                                                ' is not available, the method will be hidden from the payment list');
```
acampos1916, line 363: Now that all PMs are attempted regardless of the details object the console gets crowded with a similar message. The error logged below.

Model/Cron.php
```diff
@@ -445,7 +445,7 @@ private function shouldSkipProcessingNotification($notification)
                 $authorisationSuccessFalseMinDate = new \DateTime('-10 minutes');
                 $createdAt = \DateTime::createFromFormat('Y-m-d H:i:s', $notification['created_at']);
 
-                $minutesUntilProcessing = $createdAt->diff($authorisationSuccessFalseMinDate)->i;
+                $minutesUntilProcessing = ($createdAt->format('U') - $authorisationSuccessFalseMinDate->format('U')) / 60;
```
msilvagarcia, line 448: Why adding this change here?

msilvagarcia (commented):
I guess there's a change that wasn't meant for this PR?

Model/Cron.php
acampos1916: Yep, found this issue during testing for Blik, is in the desc:
>Unhappy flows for Blik send a AUTHORISATION success=false notification that was being ignored in some scenarios during cron run, same as #1014

msilvagarcia (approved):

peterojo (approved):

PR #1009, opened 2021-04-12
[PW-4510] Add paybright in open invoice list

AlexandrosMor:
**Description**

**Tested scenarios**

**Fixed issue**:

PR #1004, opened 2021-04-08
Remove unknown function call allowRecurring

freestream:
This method was removed in version 6.7 (https://github.com/Adyen/adyen-magento2/commit/0d9fa3cc2a806b584a14a28aeb6774b3b8af8c8f). Currently crashing order placement though admin if grand total is grater than 0.

https://github.com/Adyen/adyen-magento2/blob/6.7.0/Block/Form/Cc.php

PR #1024, opened 2021-05-04
Release 7.0.1

acampos1916:
**Fixes**
Remove unknown function call allowRecurring (#1004)
Add Paybright to open invoice list (#1009)
Adding Cartes Bancaires to supported payment methods (#1015)
[PW-4538] Using type-brand structure for giftcards (#1016)
Update the time calculation between offer closed and created (#1018)
Removing unused AdyenRequestMerchantSessionInterface (#1021)
Removing trailing commas from JS files (#1019)
Setting Pay by Mail configuration to use store ID (#1013)
Attempt to create and mount the component for all HPP methods (Support for Blik) (#1022)
Allowing zero sub totals in Helper/PaymentMethods.php (#1028)
Remove storing additionalData for banktransfer_iban and boleto (#1032)


msilvagarcia (dismissed):

msilvagarcia (approved):

AlexandrosMor (dismissed):

peterojo (approved):

cyattilakiss:
@AlexandrosMor version bump is missing

Helper/PaymentMethods.php
```diff
@@ -209,13 +217,13 @@ protected function getCurrentPaymentAmount()
 
         $total = (float)$total;
 
-        if ($total > 0) {
+        if ($total >= 0) {
```
cyattilakiss, line 220: what is the reason behind thins change? I think we don't support 0 cart amount.

Model/Cron.php
```diff
@@ -415,8 +415,7 @@ private function shouldSkipProcessingNotification($notification)
                 // OFFER_CLOSED notifications needs to be at least 10 minutes old to be processed
                 $offerClosedMinDate = new \DateTime('-10 minutes');
                 $createdAt = \DateTime::createFromFormat('Y-m-d H:i:s', $notification['created_at']);
-
-                $minutesUntilProcessing = $createdAt->diff($offerClosedMinDate)->i;
+                $minutesUntilProcessing = ($createdAt->format('U') - $offerClosedMinDate->format('U')) / 60;
```
cyattilakiss, line 418: what was the issue with the previous?

AlexandrosMor: The usage of ->i is not accurate to calculate the minutes difference between the two DateTime objects created and offer closed . The "i" attribute doesn't take into account the hours and the days in DateInterval.
Using $minutesUntilProcessing = ($createdAt->format('U') - $offerClosedMinDate->format('U')) / 60; Is taking the account the days and the hours

cyattilakiss: ah got it, thanks for the explanation

AlexandrosMor (approved):

peterojo (approved):

PR #1030, opened 2021-05-05
[PW-4496] Rename redirectUrl to successpage in config providers

AlexandrosMor:
**Description**

Rename redirectUrl in config providers since all of them are successpage now

**Tested scenarios**

Test successpage
CC
3DS1
Ideal
Ideal Refused

**Fixed issue**:

### DIFF
--- a/Model/Ui/AdyenBoletoConfigProvider.php
+++ b/Model/Ui/AdyenBoletoConfigProvider.php
@@ -81,7 +81,7 @@ class AdyenBoletoConfigProvider implements ConfigProviderInterface
             'payment' => [
                 self::CODE => [
                     'isActive' => true,
-                    'redirectUrl' => $this->_urlBuilder->getUrl(
+                    'successpage' => $this->_urlBuilder->getUrl(
                         'checkout/onepage/success/',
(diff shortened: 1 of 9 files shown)


acampos1916 (approved):

Model/Ui/AdyenBoletoConfigProvider.php
```diff
@@ -81,7 +81,7 @@ public function getConfig()
             'payment' => [
                 self::CODE => [
                     'isActive' => true,
-                    'redirectUrl' => $this->_urlBuilder->getUrl(
+                    'successpage' => $this->_urlBuilder->getUrl(
```
cyattilakiss, line 84: is there a reason why you went for the `successpage` instead of `successPage`?

AlexandrosMor: not a good reason :). I will change it to keep the naming convention

AlexandrosMor: Could @acampos1916 @cyattilakiss check again ?

acampos1916 (approved):

PR #1034, opened 2021-05-05
Using charged currency to enable the grand total display

acampos1916:
**Description**
Now that the plugin can be configured to process payments in the base or the display currency we're using the configuration value to modify the return from `Magento_Tax/js/view/checkout/summary/grand-total::isBaseGrandTotalDisplayNeeded` through a mixin. For reference, this is the function that is being overriden.

```
isBaseGrandTotalDisplayNeeded: function () {

            var total = this.totals();

            if (!total) {
                return false;
            }

            return total['base_currency_code'] != total['quote_currency_code']; //eslint-disable-line eqeqeq
        }
```

**Tested scenarios**

- Configured charged currency base, base currency differs from storeview's: The "You will be charged for XX" message is shown.
- Configured charged currency base, base currency equal to the storeview's: No message is shown.
- Configured charged currency display, base currency differs from storeview's: No message is shown.
- Configured charged currency display, base currency equal to the storeview's: No message is shown.

**Fixed issue**:  PW-4480

### DIFF
--- a/Model/Ui/AdyenGenericConfigProvider.php
+++ b/Model/Ui/AdyenGenericConfigProvider.php
@@ -23,32 +23,42 @@
 
 namespace Adyen\Payment\Model\Ui;
 
+use Adyen\Payment\Helper\Config;
+use Adyen\Payment\Helper\Data;
+use Adyen\Payment\Model\Config\Source\RenderMode;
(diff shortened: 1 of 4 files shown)


AlexandrosMor (approved):

cyattilakiss:
@acampos1916 can this be merged?

PR #1035, opened 2021-05-06
[PW-4636] Remove non-digits from postcode for Brazilian addresses

AlexandrosMor:
**Description**

The post codes for Brazilian address includes a `-` or other symbols that we need to remove before the payments request. 

**Tested scenarios**

Tested with boleto
**Fixed issue**:

### DIFF
--- a/Helper/Requests.php
+++ b/Helper/Requests.php
@@ -277,7 +277,7 @@ class Requests extends AbstractHelper
                     $requestDelivery["postalCode"] = preg_replace(
                         '/[^\d]/',
                         '',
-                        $shippingAddress->getPostcode()
+                        $requestDelivery["postalCode"]
                     );
                 }
             }
--- a/Helper/Requests.php
+++ b/Helper/Requests.php
@@ -204,6 +204,13 @@ class Requests extends AbstractHelper
 
             if (!empty($billingAddress->getPostcode())) {
                 $requestBilling["postalCode"] = $billingAddress->getPostcode();
+                if ($billingAddress->getCountryId() == "BR") {
+                    $requestBilling["postalCode"] = preg_replace(
+                        '/[^\d]/',
+                        '',
+                        $billingAddress->getPostcode()
+                    );
+                }
             }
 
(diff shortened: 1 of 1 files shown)


cyattilakiss (dismissed):

Helper/Requests.php
```diff
@@ -218,6 +214,17 @@ public function buildAddressData($billingAddress, $shippingAddress, $storeId, $r
                 $requestBilling["country"] = $billingAddress->getCountryId();
             }
 
+            if (!empty($billingAddress->getPostcode())) {
+                $requestBilling["postalCode"] = $billingAddress->getPostcode();
+                if ($billingAddress->getCountryId() == "BR") {
+                    $requestBilling["postalCode"] = preg_replace(
+                        '/[^\d]/',
+                        '',
+                        $billingAddress->getPostcode()
```
acampos1916, line 223: The postcode is already in the var right?
```suggestion
                        $requestBilling["postalCode"]
```

acampos1916 (dismissed):

Helper/Requests.php
AlexandrosMor: yes it is, we could go for your suggestion

Helper/Requests.php
```diff
@@ -268,6 +271,17 @@ public function buildAddressData($billingAddress, $shippingAddress, $storeId, $r
                 $requestDelivery["country"] = $shippingAddress->getCountryId();
             }
 
+            if (!empty($shippingAddress->getPostcode())) {
+                $requestDelivery["postalCode"] = $shippingAddress->getPostcode();
+                if ($shippingAddress->getCountryId() == "BR") {
+                    $requestDelivery["postalCode"] = preg_replace(
+                        '/[^\d]/',
+                        '',
+                        $shippingAddress->getPostcode()
```
acampos1916, line 280: Here as well?
```suggestion
                        $requestDelivery["postalCode"]
```

msilvagarcia (approved):

cyattilakiss (approved):

PR #1036, opened 2021-05-06
Getting item taxAmount from row's taxAmount

acampos1916:
**Description**
Payment, Capture and Refund requests were sending the item taxAmount incorrectly as this value in Magento's objects is not per item but rather per row. So rows with qty>1 were using the wrong taxAmount.

The taxAmount in the payments call is not required for open invoice items so it is being removed to avoid further similar issues in that step at least.

**Tested scenarios**
Payment, capture and refund of orders with multiple line items with qty>1, shipping costs and discounts.

**Fixed issue**:  PW-4638

### DIFF
--- a/Gateway/Request/CheckoutDataBuilder.php
+++ b/Gateway/Request/CheckoutDataBuilder.php
@@ -286,7 +286,6 @@ class CheckoutDataBuilder implements BuilderInterface
             $formFields['lineItems'][] = [
                 'id' => $item->getId(),
                 'amountExcludingTax' => $formattedPriceExcludingTax,
-                'taxAmount' => $formattedTaxAmount,
                 'description' => $item->getName(),
                 'quantity' => $numberOfItems,
                 'taxCategory' => $item->getProduct()->getAttributeText('tax_class_id'),
@@ -305,7 +304,6 @@ class CheckoutDataBuilder implements BuilderInterface
             $formFields['lineItems'][] = [
                 'id' => 'Discount',
                 'amountExcludingTax' => $itemAmount,
-                'taxAmount' => $itemVatAmount,
                 'description' => $description,
                 'quantity' => $numberOfItems,
                 'taxCategory' => 'None',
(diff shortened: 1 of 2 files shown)


Gateway/Request/CheckoutDataBuilder.php
```diff
@@ -337,7 +335,6 @@ protected function getOpenInvoiceData($order): array
             $formFields['lineItems'][] = [
                 'id' => 'shippingCost',
                 'amountExcludingTax' => $formattedPriceExcludingTax,
-                'taxAmount' => $formattedTaxAmount,
```
acampos1916, line 340: I thought of removing these items from the payments request because it is not required. The Capture and Refund calls do seem to need it so those are left there.

cyattilakiss: @maassenbas is it similar to what you did in SFCC?

maassenbas: In SFCC, we set the taxPercentage=0 instead of removing the taxAmount. 
https://github.com/Adyen/adyen-salesforce-commerce-cloud/pull/456/files

acampos1916: Gotcha @maassenbas, I guess you do that for lines that don't have tax? In our case the lines might have tax, we were previously sending the amount and the percentage but I see here that only the percentage is required (at least for capture and refunds) https://docs.adyen.com/payment-methods/klarna/invoice-lines#partial-capture-or-refund

cyattilakiss: @acampos1916 I'm not even sure if it's going to solve the problem or not.
In case you remove the tax amount but keep the line item tax percentage then the tax amount will still be the "same"  because it is calculated by item. Were you able to test it with both scenarios when the tax is calculated on the whole cart and also when it's calculated per item?

cyattilakiss (approved):

Issue #1045, opened 2021-05-14
Cannot checkout using Adyen payment method in Magento admin


jlczapski:
**Describe the bug**
When trying to create an order for a customer in Magento admin panel, Adyen payment method is not rendering due to error: 

```Please configure an API Key and a live endpoint prefix(if in Production Mode) in your Adyen Required Settings```

When looking in codebase, we can see the following statement, which checks for checkout origin keys:

```
if (!"<?= $block->getCheckoutOriginKeys(); ?>") {
     document.getElementById('noApiKey').style = "visibility: visible; display:inline-block";
     return;
}
```

But when searching for this method ```$block->getCheckoutOriginKeys()```, it appears it no longer exists? So the payment method is always being prevented from rendering, even if API keys are set in configuration.

**To Reproduce**
Steps to reproduce the behaviour:
1. Log into Magento admin panel
2. Sales > Orders
3. Click 'Create New Order'
4. Click 'Add to Order'
5. Search for a SKU using the form and tick at least one product
6. Click 'Add Selected to Bag'
7. Scroll down to 'Payment and Shipping Information'
8. Error message displays beneath 'Debit/Credit Card'

**Expected behaviour**
Admin users in Magento should be able to create an order and take payment using Adyen payment method

**Magento version**
2.4.2

**Plugin version**
7.0.1

**Screenshots**
![Screenshot 2021-05-14 at 12 04 13](https://user-images.githubusercontent.com/16479189/118262309-9ad5a780-b4ac-11eb-885d-8b5df74ed151.png)
![Screenshot 2021-05-14 at 12 04 31](https://user-images.githubusercontent.com/16479189/118262319-a032f200-b4ac-11eb-9f1e-704cf1f39ac1.png)

**Desktop (please complete the following information):**
 - OS: MacOS Mojave
 - Browser Firefox
 - Version 88.0.1

**Additional context**

acampos1916:
Hello @jlczapski thanks for opening this issue. This is a known bug in versions 7.0.0 and 7.0.1, it is being fixed and will be released in 7.0.2.

acampos1916:
@jlczapski are you able to test if #1048 fixes this issue for you for card payments?

Issue #1040, opened 2021-05-10
[PW-4643] Critical error on create order by admin with Adyen 7.0.0


mirkocesaro:
**Describe the bug**
When an admin tries to create an order for a customer, there is a critical error in the log:

`report.CRITICAL: Invalid method Adyen\Payment\Block\Form\Cc::allowRecurring`

On developer mode, on the product submit to the cart, your journey will end on the page
https://{{base_url}}/{{admin}}/sales/order_create/processData/key/{{KEY}}/

With the following message:
```
1 exception(s):
Exception #0 (Magento\Framework\Exception\LocalizedException): Invalid method Adyen\Payment\Block\Form\Cc::allowRecurring
```

**To Reproduce**
Steps to reproduce the behaviour:
1. Set Magento in developer mode
2. Go to the admin
2. Create a new order
3. Add a product to the order by SKU
4. Click on "Add to Order"

**Expected behaviour**
The item is added to the order

**Magento version**
2.4.1

**Plugin version**
7.0.0

jlczapski:
We are also experiencing this issue on Magento 2.4.2, Adyen 7.0.0. Do you have any updates?

acampos1916:
Hello @mirkocesaro @jlczapski. Indeed as commented in #1045 there is as known bug with admin orders in 7.0.0 and 7.0.1. A fix is currently being worked on and will be released ASAP.

acampos1916:
@mirkocesaro are you able to test if #1048 fixes this issue for you for card payments?

PR #1048, opened 2021-05-18
[PW-4643] Implementing CC with web components for admin orders

acampos1916:
**Description**

The CC form for admin orders used an older version of the web components library. In this PR it is being updated to use the same version as the frontend checkout. The bundle.js was moved to the base area so it's available in frontend and adminhtml. The request body is taken from the statedata.

**Tested scenarios**
Refused and Authorised CC payments for 3DS1, 3DS2 (No auth needed) and non-3DS

TODO: Store payment method details? Installments?
We're disabling the payment method details save feature in the admin panel because OneClick payments require CVC, which is not used in MOTO.
We're using the installments feature from the Web Component.

**Fixed issue**:  #1040 #1045 PW-4643


cyattilakiss (dismissed):
There is 1 code smell that could be removed easily, pls check that, the rest looks good :)
Did you check w @KadoBOT the instalments options in the component?

AlexandrosMor (approved):

acampos1916:
> There is 1 code smell that could be removed easily, pls check that, the rest looks good :)
> Did you check w @KadoBOT the instalments options in the component?

Checked with Ricardo, the installment options are only modified in the `card.values` object, so what we're supplying now according to the docs is working.

KadoBOT (approved):
`installmentOptions` LGTM

cyattilakiss (approved):

Issue #995, opened 2021-03-29
[PW-3410] Amazon Pay (end checkout support)


fahu:
As Adyen offers full support for Amazon Pay (including recurring payments, https://docs.adyen.com/payment-methods/amazon-pay) I would expect the Adyen Magento 2 plugin to also support Amazon Pay.

Is there already a timeline / roadmap when Amazon Pay will be supported by the Magento 2 module?

msilvagarcia:
Hey @fahu !

It's definitely in our roadmap, but we don't have a timeline yet.

Cheers!
Marcos
Adyen

PR #1050, opened 2021-05-21
[PW-4531] Add support for AmazonPay

msilvagarcia:
This adds support for AmazonPay as a payment method

**Description**

The web component redirects the shopper to the Amazon website where they choose a payment method from their account. Amazon redirects the shopper to the merchant's website, adding a token in the URL which is used to finally submit the payment.

**Tested scenarios**

Payment with AmazonPay

**Fixed issue**:  #995


AlexandrosMor:
It is implemented #1091

PR #1053, opened 2021-05-25
[PW-4652] CC holder name config toggles

acampos1916:
**Description**
When using the Web Component is possible to toggle the holder name field and if it is required. Here a couple of configs are being added to control this feature.

**Tested scenarios**
has_holder_name toggles if the holder name is shown.
holder_name_required toggles if the holder name is required.

**Fixed issue**:  PW-4652

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
@@ -77,7 +77,6 @@ define(
                 this.vaultEnabler.isActivePaymentTokenEnabler(false);
 
                 this.checkoutComponent = new AdyenCheckout({
-                        hasHolderName: true,
                         locale: adyenConfiguration.getLocale(),
                         clientKey: adyenConfiguration.getClientKey(),
                         environment: adyenConfiguration.getCheckoutEnvironment(),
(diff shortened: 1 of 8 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -126,6 +126,9 @@ define(
                 self.cardComponent = self.checkoutComponent.create('card', {
                     enableStoreDetails: self.getEnableStoreDetails(),
                     brands: self.getAvailableCardTypeAltCodes(),
+                    hasHolderName: adyenConfiguration.getHasHolderName(),
```
cyattilakiss, line 129: there is a hasHolderName config already set in the line 80 in this file and also in the adyen-hpp-method.js file. I think those needs to be removed and replaced by this new setting.

acampos1916: I think I prefer removing the one from line 80, as it is a card specific config. What do you think?

About the one in adyen-hpp-method.js, is it necessary for hpp methods? If so I can move the config to `Advanced: Checkout Experience` so it applies not only for cards.

cyattilakiss: I don't think we actually need it for alternative payment methods
I'm also good with removing it from line 80

acampos1916: I just realized that it also affects PMs like bcmc, that use cards in a separate component. I've moved it to checkout experience and adyen_abstract

AlexandrosMor (approved):

cyattilakiss (approved):

PR #1059, opened 2021-06-02
[PW-4363] Map zh_Hans_* to zh_CN locale code

AlexandrosMor:
**Description**
In the checkout configuration the locale for the Chinese locale code is zh-CN as described in the documentation [here](https://docs.adyen.com/online-payments/components-web/localization-components#change-language). Same for Taiwan, too. The  Magento, returns the code for this same locale as  zh_Hans_* which is not matching with the Adyen's checkout

**Tested scenarios**

Tested with: NL, CN variants and TW

**Fixed issue**:

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1794,7 +1794,10 @@ class Data extends AbstractHelper
      */
     public function mapLocaleCode($localeCode)
     {
-        if ($localeCode && str_contains($localeCode, 'zh_Hans_')) {
+        if ($localeCode && $localeCode == 'zh_Hant_TW') {
+            $localeCode = 'zh-TW';
+        }
+        elseif ($localeCode && ($localeCode == 'zh_Hans_CN'||$localeCode== 'zh_Hant_HK')) {
             $localeCode = 'zh-CN';
         }
         return $localeCode;
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -144,6 +144,11 @@ class Data extends AbstractHelper
     private $componentRegistrar;
 
     /**
+     * @var \Adyen\Payment\Helper\Locale;
+     */
+    private $locale;
+
+    /**
      * Data constructor.
      *
      * @param \Magento\Framework\App\Helper\Context $context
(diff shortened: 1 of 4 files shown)


Helper/Data.php
```diff
@@ -1785,4 +1786,20 @@ public function getPspReferenceSearchUrl($pspReference, $liveEnvironment)
             $pspReference
         );
     }
+
+    /**
+     * Maps zh_Hans_* locale code to zh_CN
+     * @param $localeCode
+     * @return mixed|string
+     */
+    public function mapLocaleCode($localeCode)
```
acampos1916, line 1795: I think I would prefer to move this to a new locale helper, where the mappings can be added as part of an array constant, what do you think?

cyattilakiss: I think it makes sense since we try move from a generic helper class to multiple more specified ones.
Also the const array could make it more clean

AlexandrosMor: @acampos1916 @cyattilakiss Could you check again?

Helper/Locale.php
```diff
@@ -0,0 +1,60 @@
...
+    public function mapLocaleCode($localeCode)
+    {
+        if (in_array($localeCode, self::$chineseLocaleCodes)) {
+            $localeCode = self::ZH_CN;
+        } elseif (in_array($localeCode, self::$taiwanLocaleCodes)) {
+            $localeCode = self::ZH_TW;
+        }
+        return $localeCode;
+    }
+}
```
acampos1916, line 60: I was thinking more like this to make it more generic, new mappings could be added as new elements of the array. Haven’t tested it though

```suggestion
class Locale
{

    private static $localeMappings = array(
        'zh_Hans_CN' => 'zh-CN',
        'zh_Hant_HK' => 'zh-CN',
        'zh_Hant_TW' => 'zh-TW'
    );

    /**
     * @param $localeCode
     * @return mixed|string
     */
    public function mapLocaleCode($localeCode)
    {
        return !empty(self::$localeMappings[$localeCode]) ? self::$localeMappings[$localeCode] : $localeCode;
    }
}
```

AlexandrosMor: @acampos1916 Indeed this is more generic I will test it

AlexandrosMor:
@acampos1916 Tested with zh_Hans_CN, zh_Hant_HK, zh_Hant_TW and other random. It works

acampos1916 (dismissed):

cyattilakiss (approved):

acampos1916 (approved):

cyattilakiss (approved):

PR #1080, opened 2021-06-23
[PW-4922] Update checkout component to 4.5.0

AlexandrosMor:
**Description**

Update checkout component from 3.29.0 to 4.5.0
Replace the bundle files and update the endpoints in Data.php

**Tested scenarios**

Test checkout component 4 
Credit Card 
3ds1 
3ds2 

Oneclick
Credit Card 
3ds1 
3ds2 

Sepa
Klarna 
Klarna refused 
Ideal 
Ideal refused
Paypal 
Google pay  
Afterpay invoice 

Admin 
cc backend admin ok
cc  stored backend admin ok

**Fixed issue**:


Helper/Data.php
```diff
@@ -35,8 +35,8 @@ class Data extends AbstractHelper
     const TEST = 'test';
     const LIVE = 'live';
     // Only used for backend orders! Checkout in front-end is using different checkout version see web folder
```
cyattilakiss, line 37: For the functions using it we have a deprecation comment so I think it makes sense to add it here as well
```suggestion
    // Only used for backend orders! Checkout in front-end is using different checkout version see web folder, will be removed in v8.0.0
```

acampos1916 (approved):

cyattilakiss (approved):

Issue #1082, opened 2021-06-24
[PW-5016] Issue with getStreetAndHouseNumberFromAddress Not allowing checkout


stutaylor:
When checking out and entering card details users are being returned to the cart.

Error in system.log file 
(part of error)
Stack trace:
#0 /app/vendor/adyen/module-payment/Helper/Requests.php(192): Adyen\Payment\Helper\Address->getStreetAndHouseNumberFromAddress(Object(Magento\Sales\Model\Order\Address\Interceptor), '1', '3')

Resulting in 
AdyenLoggerTest.ERROR: Order not found. {"is_exception":false} []

The admin settings are set to map street 1 for the house number

This is happening on multiple installs of Adyen on magento 2.4.+

acampos1916:
Hello @stutaylor thanks for reporting this. Could you provide more information to try to reproduce? I see that "Order not found." is logged [here](https://github.com/Adyen/adyen-magento2/blob/master/Model/AdyenOrderPaymentStatus.php#L85) as part of the payment status check.

With Street line 1 for the house number I'm not able to reproduce in 2.4, maybe the full error has more info?

stutaylor:
Hi @acampos1916 

So the version of magento i am using is 2.4.2 -p1  on magento cloud

Here is the error in full from the system.log

[2021-06-24 16:48:02] report.CRITICAL: TypeError: Argument 1 passed to Adyen\Payment\Helper\Address::getStreetAndHouseNumberFromAddress() must be an instance of Magento\Payment\Gateway\Data\AddressAdapterInterface, instance of Magento\Sales\Model\Order\Address\Interceptor given, called in /app/vendor/adyen/module-payment/Helper/Requests.php on line 192 and defined in /app/vendor/adyen/module-payment/Helper/Address.php:41
Stack trace:
#0 /app/vendor/adyen/module-payment/Helper/Requests.php(192): Adyen\Payment\Helper\Address->getStreetAndHouseNumberFromAddress(Object(Magento\Sales\Model\Order\Address\Interceptor), '1', '3')
#1 /app/vendor/adyen/module-payment/Gateway/Request/AddressDataBuilder.php(64): Adyen\Payment\Helper\Requests->buildAddressData(Object(Magento\Sales\Model\Order\Address\Interceptor), Object(Magento\Sales\Model\Order\Address\Interceptor), 1, Array)
[...]

PR #1086, opened 2021-06-29
[PW-5016] Remove AddressAdapterInterface type declaration

cyattilakiss:
**Description**
The getBillingAddress() return type can be overwritten by third party
modules and that might break the interface's return type hint, therefore
the getStreetAndHouseNumberFromAddress should not require an address
that implements the AddressAdapterInterface

**Tested scenarios**

**Fixed issue**:  
#1082 [PW-5016] Issue with getStreetAndHouseNumberFromAddress Not allowing checkout

### DIFF
--- a/Helper/Address.php
+++ b/Helper/Address.php
@@ -39,7 +39,7 @@ class Address
      * @return array
      */
     public function getStreetAndHouseNumberFromAddress(
-        AddressAdapterInterface $address,
+        $address,
         $houseNumberStreetLine,
         $customerStreetLinesEnabled
     ): array {


Basvanderlouw (commented):
You might want to update the PHPDoc block as well.

cyattilakiss:
> You might want to update the PHPDoc block as well.

Hi @Basvanderlouw thanks for the review, the reason I kept the phpdoc is to keep the autocomplete for developers. Although it is possible with inline type hinting as well like `/** @var $address AddressAdapterInterface */`, do you prefer that over the phpdoc?

Basvanderlouw:
Hi @cyattilakiss, good point. I'd prefer the PHPDoc block over the inline type hint..

acampos1916 (approved):

AlexandrosMor (approved):

PR #1095, opened 2021-07-08
[PW-5000] Prefill house number field in the checkout component 

cyattilakiss:
**Description**

Use the house_number_street_line admin config field on the frontend to be able to prefill the checkout component input fields with the correct address line values.
Create new helper function getFormattedAddress() to make it easier retrieving the correct address details for the component while introducing the houseNumber value.
Use the correct address type (shipping/billing) for the checkout component config fields.
Add shippingAddress configuration to the component.

See commit messages for more detailed steps.

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
@@ -264,12 +264,17 @@ define(
                                 shopperDateOfBirth = customerData.dob;
 
                                 var formattedShippingAddress = {};
+                                var formattedBillingAddress = {};
 
                                 if (!!quote.shippingAddress()) {
                                     formattedShippingAddress = getFormattedAddress(quote.shippingAddress());
                                 }
 
-                                // Use shipping address details as default
+                                if (!!quote.billingAddress()) {
+                                    formattedBillingAddress = getFormattedAddress(quote.billingAddress());
+                                }
+
+                                // Use shipping address details as default and fall back to billing address if missing
                                 if (formattedShippingAddress) {
                                     country = formattedShippingAddress.country;
                                 }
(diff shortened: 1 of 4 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -245,37 +245,88 @@ define(
...
+
+                                    // address contains line items as an array, otherwise if string just pass along as is
+                                    if (Array.isArray(street)) {
+                                        // getHouseNumberStreetLine > 0 implies the street line that is used to gather house number
+                                        if (adyenConfiguration.getHouseNumberStreetLine() > 0) {
+                                            // removes the street line from array that is used to contain house number
+                                            houseNumber = street.splice(adyenConfiguration.getHouseNumberStreetLine() - 1, 1);
+                                        } else { // getHouseNumberStreetLine = 0 uses the last street line as house number
+                                            // removes last street line from array that is used to contain house number
+                                            houseNumber = street.pop();
```
acampos1916, line 309: I think this could be wrong in cases where the house number needs to be extracted using regex (e.g. only one street line enabled and input Simon Carmiggeltstraat 6 - 50), because if I'm not mistaken this is the raw input from Magento's form. Maybe this helper function can be adjusted and used to directly provide the house number instead https://github.com/Adyen/adyen-magento2/blob/develop/Helper/Address.php#L41

cyattilakiss: Thank you, very sharp! :)

The reason why I didn't use the helper Address.php#L41 is because of PWA implementations and one step checkouts. In case a merchant is using one step checkout then the shipping and billing address is editable on the same page and the frontend won't call the backend for the correct address details at every change. The same thing can happen with PWA when the merchant wants to minimize the number of the API calls.

We see that the regex solution is unreliable so I think it would not be a good idea to also introduce it on the frontend as a new helper but to improve the PR what do you think if in this line instead of just using the last line (which might be the same as the full street in case a merchant has only 1 street line) is if we check if there is only one and then we don't even prefill the houseNumberOrName and we leave it to the customer to fill in the correct format instead of filling it in with the wrong details.

acampos1916: Ah yes you're right, maybe the backend wont be queried at this point in other kinds of frontend implementations. I like your proposal of using the last line.

There's also the case of US addresses where the house number is at the beginning, but I think that can be dealt with the `getHouseNumberStreetLine()` config above provided that they have a separate field for it.

cyattilakiss: I will push a new commit with the changes later today!

acampos1916 (approved):

AlexandrosMor (approved):

Issue #1017, opened 2021-04-22
IE 11 Checkout Syntax Errors


itsbreadd:
**Describe the bug**
Syntax issues with Adyen in checkout on IE11 which is preventing checkout from rendering. Customers are unable to checkout even when not using Adyen as payment method.

**To Reproduce**
1. Ensure Adyen CreditCard API integration is configured and Adyen is enabled as a payment method
2. Navigate to the frontend (https://www.paulsmith.com/uk) on an IE11 browser (My ver: 18363.1440)
3. Add any product to the basket
4. Navigate to the checkout and monitor the console log
5. See multiple syntax errors

**Expected behavior**
Magento retrieves a list of available payment methods from Adyen

**Magento version**
Tested on both 2.3.4 (Live) & 2.4.2 (Staging)

**Plugin version**
Tested on both 6.6.5 (Live) & 7.0.0 (Staging)

**Desktop:**
 - OS: Windows 10
 - Browser: Internet Explorer
 - Version: 18363.1440

**Additional context**
Looking at the code we have noticed numerous trailing commas and unsupported syntax. There could be more, but we have identified the main offenders and have patched this issue to give support for IE11 browsers using the below:

````
diff --git a/view/frontend/web/js/model/adyen-payment-service.js b/view/frontend/web/js/model/adyen-payment-service.js
index 1e2532b..bb7fb0b 100644
--- a/view/frontend/web/js/model/adyen-payment-service.js
+++ b/view/frontend/web/js/model/adyen-payment-service.js
@@ -49,7 +49,7 @@ define(
 
                 return storage.post(
                     serviceUrl,
-                    JSON.stringify(payload),
+                    JSON.stringify(payload)
                 );
             },
             getPaymentMethods: function() {
@@ -83,9 +83,9 @@ define(
                 return storage.post(
                     serviceUrl,
                     JSON.stringify(payload),
-                    true,
+                    true
                 );
             }
         };
-    },
+    }
 );
diff --git a/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js b/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
index 3263ada..e8d3070 100755
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
@@ -82,9 +82,8 @@ define(
[...]

PR #1108, opened 2021-07-19
[PW-5107] Switch billing address to shipping address for amazon addressDetails

cyattilakiss:
**Description**

Use shipping address for amazon pay addressDetails config instead of billing address.


acampos1916 (commented):
We had an issue with IE11 and trailing commas before, any reason why we should add them here? https://github.com/Adyen/adyen-magento2/issues/1017

cyattilakiss:
> We had an issue with IE11 and trailing commas before, any reason why we should add them here? #1017

Thanks for the heads up, I will remove them!

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -427,11 +432,11 @@ define(
                                             showOrderButton: false,
                                             amount: {
                                                 currency: configuration.amount.currency,
-                                                value: configuration.amount.value
+                                                value: configuration.amount.value,
```
peterojo, line 435: One more
```suggestion
                                                value: configuration.amount.value
```

peterojo (dismissed):

acampos1916 (approved):

acampos1916 (approved):

peterojo (approved):

AlexandrosMor (approved):

PR #1109, opened 2021-07-19
[PW-4989] State data webapi endpoint

acampos1916:
**Description**
A new State Data webapi endpoint that receives a quote_id and a state data JSON to be stored in a new BD entity. This could be used in custom checkout flows and multishipping.

The stored state data is now checked if there's no frontend state data in the data assign observers.

If the payment has been Authorised the state data helper cleans the records for the quote.

**Tested scenarios**
Store state data and quote ID.
Use stored state data if no state data from the frontend is provided.
Clean state data records if the order is placed.

**Fixed issue**:  PW-4989


Api/Data/StateDataInterface.php
```diff
@@ -0,0 +1,94 @@
...
+
+    /*
+     * Quote ID.
+     */
+    const QUOTE_ID = 'quote_id';
+
+    /*
+     * Payment State Data.
+     */
+    const STATE_DATA = 'state_data';
```
cyattilakiss, line 46: shouldn't we provide some more information in the phpdocs (so it will also show up in the swagger doc) like `stringified json returned by the adyen checkout component`.

etc/webapi.xml
```diff
@@ -58,6 +58,13 @@
         </resources>
     </route>
 
+    <route url="/V1/adyen/me/stateData" method="POST">
```
cyattilakiss, line 61: what is the `me` in the url?

peterojo (approved):

Model/ResourceModel/StateData/Collection.php
```diff
@@ -0,0 +1,65 @@
...
+    public function _construct()
+    {
+        $this->_init(
+            StateData::class,
+            ResourceModel::class
+        );
+    }
+
+    /**
+     * Fetch the most recent state data with quote ID or return and empty array
```
acampos1916, line 42:
```suggestion
     * Fetch the most recent state data with quote ID or return an empty array
```

Model/AdyenStateData.php
```diff
@@ -0,0 +1,81 @@
...
+ * #############  #####( ######  #####. ######  #############  #############
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
```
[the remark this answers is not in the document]
acampos1916: Yeah copied from an outdated file. I've changed it now 👍

Api/Data/StateDataInterface.php
acampos1916: I've added a description for the function and the params, but only the function one shows up in Swagger. Do you know if this is done differently for the params?

etc/webapi.xml
acampos1916: I was experimenting with this. If "me" is used in the path then some params can be forced by M2, like customerId. But for now we don't need it
```suggestion
    <route url="/V1/adyen/stateData" method="POST">
```

peterojo (approved):

AlexandrosMor (approved):

PR #1111, opened 2021-07-20
[PW-3060] remove 3ds2 toggle config and usage

peterojo:
**Description**

Toggle for enabling/disabling 3ds2 was introduced during transition period from 3ds1 to 3ds2
Now we enable 3ds2 by default and remove this config option
**Tested scenarios**

3ds2 CC payment
**Fixed issue**:

### DIFF
--- a/Helper/Requests.php
+++ b/Helper/Requests.php
@@ -337,25 +337,19 @@ class Requests extends AbstractHelper
      */
     public function buildThreeDS2Data($additionalData, $storeId, $request = [])
     {
-        if ($this->adyenHelper->isCreditCardThreeDS2Enabled($storeId)) {
-            $request['additionalData']['allow3DS2'] = true;
-            $request['origin'] = $this->adyenHelper->getOrigin($storeId);
-            $request['channel'] = 'web';
-            $request['browserInfo']['screenWidth'] = $additionalData[AdyenCcDataAssignObserver::SCREEN_WIDTH];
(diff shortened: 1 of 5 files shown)


Helper/Requests.php
```diff
@@ -337,25 +337,19 @@ public function buildBrowserData($request = [])
...
+        $request['browserInfo']['language'] = $additionalData[AdyenCcDataAssignObserver::LANGUAGE];
+
+        if ($javaEnabled = $additionalData[AdyenCcDataAssignObserver::JAVA_ENABLED]) {
+            $request['browserInfo']['javaEnabled'] = $javaEnabled;
         } else {
-            $request['additionalData']['allow3DS2'] = false;
-            $request['origin'] = $this->adyenHelper->getOrigin($storeId);
-            $request['channel'] = 'web';
+            $request['browserInfo']['javaEnabled'] = false;
         }
```
cyattilakiss, line 353: Not 100% connected to this ticket but do we actually still need to add it or is the stateData enough since it's coming from the frontend in the state data?

peterojo: I noticed that this method is not called anywhere in the code, and changing some of the variables here did not affect the data in the outgoing API call. but I'm not sure whether there's something I'm missing.

peterojo: @acampos1916 what do you think? do you know if we can safely remove the `buildThreeDS2Data()` method?

acampos1916: It was being used before moving to v7, so it can be removed now @peterojo

cyattilakiss (approved):

acampos1916 (approved):

PR #1124, opened 2021-08-04
[PW-5211] Authorize web api requests

AlexandrosMor:
**Description**

Authorize webapi requests so the following are not used by guest users:
 payment-status
Origin key

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/etc/webapi.xml
+++ b/etc/webapi.xml
@@ -75,7 +75,7 @@
     <route url="/V1/adyen/orders/:orderId/payment-status" method="GET">
         <service class="Adyen\Payment\Api\AdyenOrderPaymentStatusInterface" method="getOrderPaymentStatus"/>
         <resources>
-            <resource ref="anonymous"/>
+            <resource ref="self"/>
         </resources>
     </route>
 </routes>
--- a/etc/webapi.xml
+++ b/etc/webapi.xml
@@ -68,7 +68,7 @@
     <route url="/V1/adyen/originKey" method="GET">
         <service class="Adyen\Payment\Api\AdyenOriginKeyInterface" method="getOriginKey"/>
         <resources>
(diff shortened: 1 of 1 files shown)


cyattilakiss (dismissed):

etc/webapi.xml
```diff
@@ -68,11 +68,11 @@
     <route url="/V1/adyen/originKey" method="GET">
         <service class="Adyen\Payment\Api\AdyenOriginKeyInterface" method="getOriginKey"/>
```
peterojo, line 69: I think we can remove the `originKey` endpoint entirely by now.

AlexandrosMor: Indeed we are not using it

peterojo: Removed.

Model/AdyenOrderPaymentStatus.php
```diff
@@ -75,12 +75,17 @@ public function __construct(
...
      * @param string $orderId
+     * @param string $shopperEmail
      * @return bool|string
      */
-    public function getOrderPaymentStatus($orderId)
+    public function getOrderPaymentStatus($orderId, $shopperEmail)
     {
         try {
             $order = $this->orderRepository->get($orderId);
+            // Validate shopper email against order email to prevent abuse
```
Morerice, line 85: Maybe update the comment to: `Validate shopper email against order email for extra security`?

Morerice (approved):

Model/AdyenOrderPaymentStatus.php
peterojo: Makes sense
```suggestion
            // Validate shopper email against order email for extra security.
```

Morerice (approved):

PR #1127, opened 2021-08-16
JSON Pretty log

jonathanribas:
**Description**
Using `print_r` for logs cause multiple lines on ECK, it's hard to debug when it is the case.

I've used `json_encode` as used on other Adyen logs.


acampos1916:
Thank you for this contribution @jonathanribas, I also think this would be an improvement to the logs readability.

acampos1916 (approved):

Morerice (approved):

PR #1128, opened 2021-08-17
[PW-4990] Cards multishipping support

acampos1916:
**Description**
This PR adds support for the Cards payment method in the multishipping checkout flow, some pointers:
* The component is mounted in the billing step and the state data input is validated by a mixin.
* When the orders are placed the response is now saved in a new Payment Response DB entity, pending cleanup.
* All orders will be placed regardless of payment response, the orders will be closed by a webhook in case of unsuccessful  payment.
* The success page, block class and template have been modified so that a "Complete Payment" button is shown for each payment that needs some kind of follow-up (i.e. the payment returned a non-final result code). In that case the component gets configured and mounted.

**To review**
Should the payment follow-up links be sent or shown somewhere so that the shopper doesn't lose access to them?
How to retry payment failures?
Installments?
Success phtml is being overriden.
Test 3DS fallback.

**Tested scenarios**
Happy and unhappy flows for non-3DS, 3DS1 and 3DS2.

**Fixed issue**:  PW-4990


view/frontend/templates/checkout/multishipping/success.phtml
```diff
@@ -0,0 +1,208 @@
...
+                                                        array_column(
+                                                            $paymentResponseEntities,
+                                                            'merchant_reference'
+                                                        )
+                                                    )]['response']
+                                                ); ?>"
+                                            <?php endif; ?>
+                                        ><?= $escaper->escapeHtml($block->getPaymentButtonLabel($orderId)); ?></button>
+                                <div class="adyen-component"></div>
+                                </span>
```
acampos1916, line 80: Add comment to indicate Adyen customization

view/frontend/web/template/payment/multishipping/form.html
```diff
@@ -0,0 +1,2 @@
+<div id="cardContainer"></div>
+<input name="payment[stateData]" id="stateData" type="hidden">
```
acampos1916, line 2: Duplicated, remove the other one

view/frontend/templates/checkout/multishipping/success.phtml
acampos1916: bbd019c

view/frontend/web/template/payment/multishipping/form.html
acampos1916: a61cbd5

Morerice (approved):

cyattilakiss (approved):

PR #1129, opened 2021-08-19
[PW-5238] Remove cc check so are not ignored for offer_closed

### LINKED ISSUE #1121: [PW-5238] OFFER_CLOSED notifications are ignored for orders paid with Google Pay
Hi,
server notifications with the event code `OFFER_CLOSED` don't affect orders paid with Google Pay.
The problem is a bit of logic that is missing here: https://github.com/Adyen/adyen-magento2/blob/7.1.0/Model/Cron.php#L1183
The order has payment method `adyen_google_pay`, but the server notification instead has the credit card brand in the `method` field, like for example `visa`, `mc`, `amex`.
So `$isOrderCc` is `false` and the order doesn't get canceled.
In my opinion `$isOrderCc` should be `true` when order payment method is `adyen_google_pay`.

**To Reproduce**
1. Make sure that open offers are configured to close automatically after a certain amount of time
2. Place an order with Google Pay, but don't complete the payment, so an open offer only is generated
3. Wait for the offer to close and receive the OFFER_CLOSED server notification
4. See that the order doesn't get canceled

**Expected behavior**
The order should be canceled.

**Magento version**
2.4.1

**Plugin version**
6.7.1 and 7.1.0

Thanks!

AlexandrosMor:
**Description**

For `OFFER_CLOSED` notifications the order cancel was skipped if the payment methods was not cc and cancelled only if the payment method of the notification matches the payment method of the order. The purpose of this is If an order was made with an Alternative payment method it will continue with the cancellation only if the payment method of the notification matches the payment method of the order. 
On the other hand, that is causing all the wallet payment methods to be skipped because the notification and the order are never matching. For example it could be `paywithgoogle` with `visa`.

Furthermore, The check for CC or one click was rejecting all wallet payment methods because the payment method code is  `adyen_hpp` and could trigger `OFFER_CLOSED` notification so should be also canceled. 

**Tested scenarios**

Google pay with 3ds1

**Fixed issue**:  fixes #1121

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -1181,7 +1181,8 @@ class Cron
                 $isOrderCc = strcmp(
                         $this->_paymentMethodCode(),
                         'adyen_cc'
-                    ) == 0 || strcmp($this->_paymentMethodCode(), 'adyen_oneclick') == 0;
+                    ) == 0 || strcmp($this->_paymentMethodCode(), 'adyen_oneclick') == 0
+                    || $this->isWalletPaymentMethod($this->_paymentMethodCode());
 
                 /*
                  * If the order was made with an Alternative payment method,
@@ -1190,13 +1191,12 @@ class Cron
                  */
                 if (!$isOrderCc && strcmp($notificationPaymentMethod, $orderPaymentMethod) !== 0) {
                     $this->_adyenLogger->addAdyenNotificationCronjob(
-                        "Order is not a credit card,
+                        "Order is not a credit card, wallet payment method
                     or the payment method in the notification does not match the payment method of the order,
                     skipping OFFER_CLOSED"
                     );
                     break;
                 }
-
                 if (!$this->_order->canCancel() && $this->configHelper->getNotificationsCanCancel(
                         $this->_order->getStoreId()
                     )) {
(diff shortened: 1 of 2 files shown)


Model/Cron.php
```diff
@@ -2243,4 +2243,15 @@ private function addNotificationErrorComment($errorMessage)
...
     }
+
+    /**
+     * Checks if a payment is wallet payment method
+     * @param $notificationPaymentMethod
+     * @return bool
+     */
+    private function isWalletPaymentMethod($notificationPaymentMethod): bool
+    {
+        $walletPaymentMethods = array('googlepay', 'paywithgoogle', 'wechatpayWeb', 'amazonpay', 'applepay');
```
Morerice, line 2254: Could this array be moved to a constant in the `Helper/PaymentMethods.php` class?

Model/Cron.php
```diff
@@ -1181,7 +1189,8 @@ protected function _processNotification()
                 $isOrderCc = strcmp(
                         $this->_paymentMethodCode(),
                         'adyen_cc'
-                    ) == 0 || strcmp($this->_paymentMethodCode(), 'adyen_oneclick') == 0;
+                    ) == 0 || strcmp($this->_paymentMethodCode(), 'adyen_oneclick') == 0
+                    || $this->paymentMethodsHelper->isWalletPaymentMethod($notificationPaymentMethod);
```
peterojo, line 1193: Won't it be better to add the wallet as a separate check? Like `$isWallet`. When you add this to the `$isOrderCc` variable it reads as if wallet payments are also CC payments.

Model/Cron.php
```diff
@@ -1190,13 +1199,12 @@ protected function _processNotification()
                  */
                 if (!$isOrderCc && strcmp($notificationPaymentMethod, $orderPaymentMethod) !== 0) {
```
peterojo, line 1200: See my other comment, this could be changed to something like `if (!$isOrderCc && !$isOrderWallet && ...`

peterojo (dismissed):

cyattilakiss (approved):

peterojo (approved):

Morerice (approved):

PR #1141, opened 2021-09-02
[PW-5364] - Support partial auth/capture notification processing

Morerice:
**Description**
Currently, when a notification of a partial auth/capture is received, the order's state is updated, even if the full amount of the auth/capture hasn't been processed yet.

The correct flow would be to update the order status only when the full amount has been authorized/captured.
Additionally, we should also add comments **WITHOUT UPDATING THE STATUS** whenever a new partial auth/capture is received.

**Steps for auth:**
1. Extract the functionality that is creating the Adyen Order Payment (**line 1648**) from the `_prepareInvoice` function. This should be done before the `__setPrePaymentAuthorized` call.
2. Use the _isTotalAmount function to check if the order status should be created and to see if an invoice should be created
3. Add comment to document that a partial auth was processed

**Steps for capture:**
1. Add a new status to the `adyen_order_payment` table to denote if the payment has been captured or not
1. Update the `_isTotalAmount` function to also be able to check if the full amount has been captured
1. Use this updated functionality to check if the status should be updated
2. If full amount has not been captured, add comment to document that a partial capture was processed

**Tested scenarios**
* Auto capture w/full amount authorization
* Auto capture w/multiple partial authorizations
* Single manual capture on a single authorization
* Multiple manual captures on multiple manual authorizations (1 to 1)

**Fixed issue**:


Model/Cron.php
```diff
@@ -1891,48 +1891,59 @@ protected function _getFraudManualReviewAcceptStatus()
...
+            ->getTotalAmount($paymentId, $captured);
 
         if ($res && isset($res[0]) && is_array($res[0])) {
             $amount = $res[0]['total_amount'];
             $formattedOrderAmount = $this->_adyenHelper->formatAmount($amount, $orderCurrencyCode);
             $this->_adyenLogger->addAdyenNotificationCronjob(
                 sprintf(
-                    'The grandtotal amount is %s and the total order amount that is authorised is: %s',
+                    'The grand total amount for order %s is %s and the currently authorised order amount is: %s',
+                    $orderId,
```
acampos1916, line 1924: I see $orderId is the DB ID, but the merchantReference/IncrementId would be more useful in this log message.

Model/Cron.php
```diff
@@ -1891,48 +1891,59 @@ protected function _getFraudManualReviewAcceptStatus()
...
                     $grandTotal,
                     $formattedOrderAmount
                 )
             );
 
             if ($grandTotal == $formattedOrderAmount) {
-                $this->_adyenLogger->addAdyenNotificationCronjob('AUTHORISATION has the full amount');
+                $this->_adyenLogger->addAdyenNotificationCronjob(sprintf(
+                    'Order %s has been fully authorized',
+                    $orderId,
```
acampos1916, line 1933: Same here

Morerice: 070ce8e

Model/Cron.php
Morerice: 070ce8e

AlexandrosMor (approved):

acampos1916 (approved):

AlexandrosMor (approved):

PR #1142, opened 2021-09-02
[PW-5170] Create invoices as pending during async capture and change state with notification #1140

### LINKED ISSUE #592: [PW-2771] Invoice paid even if the capture fails
**Magento version**: 2.3.3
**Plugin version**: 5.0.2
**Description**

When the invoice is created, the invoice state immediately is set to Paid. Anyway we only received the synchronous api response “capture-received”. This does not mean that payment has already captured. It just means that Adyen is taking care to capture the payment, and the result will be notified within a couple of minutes by the asynchronous notification. Only this will confirm if the payment has been successfully captured or not. 

Then the correct way to manage the invoice state by asynchronous notification is the following:

1) when the invoice is created the invoice state is “Pending”
2) when the asynchronous notification is received the invoice state changes to “Paid” if the payment was captured successfully. While if the capture failed, the invoice state changes to “Canceled”

A simple case: 
- Make a first partial invoice (status is paid)
- First invoice receives a capture notification with success=true
- Make a second invoice for the rest of the order (status is paid)
- Since my account does not have multi captures enabled, the second invoice receives a capture notification with success=false
- the order on Magento appears totally invoiced and paid, unless a comment on the history reporting the failure

AlexandrosMor:
**Description**

In case of `capture delay` `manual` for an order, the invoice will be created with status paid without processing the capture notification. With this pull request the invoice is created with status pending and it is updated to `Paid` by the capture notification. 
In case of `Capture Delay` is  set to `immediate` in the back office the invoice is automatically created with status `Paid` by processing the Authorisation notification because there is no capture notification.
 
**Tested scenarios**

CC and klarna for both auto and manual capture 
**Fixed issue**:  
fixes #592


Gateway/Response/PaymentCaptureDetailsHandler.php
```diff
@@ -41,6 +41,9 @@ public function handle(array $handlingSubject, array $response)
         // set pspReference as lastTransId only!
         $payment->setLastTransId($response['pspReference']);
 
+        if($response["response"]=='[capture-received]'){
```
acampos1916, line 44:
```suggestion
        // The capture request will return a capture-received message, but it doesn't mean the capture has been final
        // so the invoice is set to Pending
        if($response["response"]=='[capture-received]'){
```

Gateway/Response/PaymentCaptureDetailsHandler.php
```diff
@@ -41,6 +41,9 @@ public function handle(array $handlingSubject, array $response)
...
+            $payment->setIsTransactionPending(true);
+        }
```
acampos1916, line 48: I think the lines below should also be included in this if, as it doesn't make sense to close the transaction if the capture request wasn't received. And should we handle an else block to stop the invoice from being created if the response isn't [capture-received] ?

Gateway/Response/PaymentCaptureDetailsHandler.php
```diff
@@ -41,11 +41,17 @@ public function handle(array $handlingSubject, array $response)
...
-        $payment->setIsTransactionClosed(true);
-        $payment->setShouldCloseParentTransaction(false);
+        // The capture request will return a capture-received message, but it doesn't mean the capture has been final
+        // so the invoice is set to Pending
+        if($response["response"]=='[capture-received]'){
+            $payment->setIsTransactionPending(true);
+
+            /**
+             * close current transaction because you have capture the goods
+             * but do not close the authorisation becasue you can still cancel/refund order
```
acampos1916, line 51:
```suggestion
             * but do not close the authorisation because you can still cancel/refund order
```

Gateway/Response/PaymentCaptureDetailsHandler.php
Morerice: Is there a reason why strict comparison is not used?

Morerice (approved):

acampos1916 (approved):

PR #1143, opened 2021-09-07
[PW-5347] update tooltips for new customer area UI

peterojo:
**Description**

Update tooltips for the paths that have changed in new CA interface
**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/etc/adminhtml/system/adyen_split_payment.xml
+++ b/etc/adminhtml/system/adyen_split_payment.xml
@@ -29,7 +29,7 @@
         <comment>
             <![CDATA[
                 <p>
-                It is possible to do split payments with GiftCards please configure here what refund strategie you want to use
+                It is possible to do split payments with GiftCards please configure here what refund strategy you want to use.
                 </p>
             ]]>
         </comment>
--- a/etc/adminhtml/system/adyen_split_payment.xml
+++ b/etc/adminhtml/system/adyen_split_payment.xml
@@ -29,7 +29,7 @@
         <comment>
             <![CDATA[
                 <p>
-                It is possible to do split payments with GiftCards please configure here what refund strategy you want to use.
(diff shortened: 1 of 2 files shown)


AlexandrosMor (approved):

etc/adminhtml/system/adyen_split_payment.xml
```diff
@@ -29,7 +29,7 @@
         <comment>
             <![CDATA[
                 <p>
-                It is possible to do split payments with GiftCards please configure here what refund strategie you want to use
+                It is possible to do split payments with GiftCards please configure here what refund strategy you want to use.
```
acampos1916, line 32: Rephrasing it a bit. And Split Payment is now related to AFP, the right label would be Partial Payment
```suggestion
                It is possible to do partial payments with GiftCards. Set here which refund strategy to use.
```

Aleffio: Should the Label on line 27 and/or the filename be renamed as well?

AlexandrosMor (approved):

etc/adminhtml/system/adyen_split_payment.xml
acampos1916: Makes sense yes. Are you able to change the label and filename @peterojo ?

peterojo: I can but there's quite a lot of references to split payments in the codebase, and we would also need to test that it still works.
Screen Shot 2021-09-09 at 3 46 37 PM

I would rather make it a separate effort. Here we just wanted to correct a typo 😄

peterojo: I've created a ticket [PW-5431] to update all references to split payments

acampos1916 (approved):

PR #1154, opened 2021-09-15
[PW-5292] - Capture full order placed with partial payments

Morerice:
**Description**
If an order is placed with partial payment (e.g. gift card HPP + Cards), it will require multiple capture requests for each amount and payment method.

**Steps**

1. In the CaptureDataBuilder get all the `adyen_order_payment` entries linked to the order. 
3. If the total of all the `adyen_order_payment` entries does not match the order total, halt the invoicing process so that the invoice is not created. Log a message about it and show and error in the frontend (admin) saying that all partial payments need to be authorised first.
4. If the total matches:
    1. In `CaptureDataBuilder` check if multiple `adyen_order_payment` entries are linked to the order. If so, add them all to the returned request and add a specific key to denote multiple order payments
    1. In `TransactionCapture` check for this specific key. If it exists, call the `capture` api endpoint multiple times, using the data passed in the `CaptureDataBuilder`. If multiple capture calls were made, add the same key to the response array, along with the multiple responses from multiple requests
    1. Update the `CaptureResponseValidator`, `PaymentCaptureDetailsHandler` and `PaymentCommentHistoryHandler` to handle a response which contains multiple entries

**Tested scenarios**

* 1 authorisation with auto capture
* 1 authorisation with manual capture
* 2 partial authorisations with auto capture
* 2 partial authorisations (both can be manually captured) with manual capture


Gateway/Response/PaymentCaptureDetailsHandler.php
```diff
@@ -38,22 +52,65 @@ public function handle(array $handlingSubject, array $response)
...
-            $payment->setShouldCloseParentTransaction(false);
+        if (array_key_exists(TransactionCapture::MULTIPLE_AUTHORIZATIONS, $response)) {
+            $this->handleMultipleCaptureRequests($payment, $response);
+        } else {
+            // set pspReference as lastTransId only!
+            $payment->setLastTransId($response['pspReference']);
+
+            // The capture request will return a capture-received message, but it doesn't mean the capture has been final
+            // so the invoice is set to Pending
+            if ($response["response"] === '[capture-received]') {
```
acampos1916, line 63: Can we move `[capture-received]` to a constant? I see it is being used multiple times in 2 classes now

Morerice: bee1dc6

Gateway/Response/PaymentCaptureDetailsHandler.php
```diff
@@ -38,22 +52,65 @@ public function handle(array $handlingSubject, array $response)
...
+    /**
+     * Set payment to pending
+     *
+     * @param $payment
+     * @return mixed
+     */
+    private function setInvoiceToPending($payment)
+    {
+        $payment->setIsTransactionPending(true);
+        $payment->setIsTransactionClosed(true);
```
acampos1916, line 111: This was working before it was outsourced to this function, but does it make sense? Is it changing the transaction status twice or is it doing something else?

Morerice: 2b70e1c I don't think it does. I removed the `setIsTransactionClosed` call and everything still works fine. Also re-added the comment which was previously there

acampos1916 (approved):

peterojo (approved):

PR #1164, opened 2021-10-03
Feature/magento coding standards

cjnewbs:
**Description**
- I know it's a small PR but I saw the old style XSD namespace location and updated it to the newer version (https://devdocs.magento.com/guides/v2.4/extension-dev-guide/build/XSD-XML-validation.html),
- I have also updated the module to pass the PHPCS ruleset in the repo,
- I did start fixing more until I saw the ruleset in the repo only checks for Errors not warnings 😓 but I have reverted my changes. While going through them I did notice some things like `@deprecated` classes and think it might be worth updating to specify why the are deprecated but I am not sure of the reason so I left that alone.

**Tested scenarios**
- The only testable change is seeing that in an IDE that supports auto-completion of XML files with an associated XSD file that the auto-completion works,
- PHPCS scanner now completes with a full pass. 😀

### DIFF
--- a/Controller/Process/Result.php
+++ b/Controller/Process/Result.php
@@ -178,7 +178,6 @@ class Result extends \Magento\Framework\App\Action\Action
             $session = $this->_session;
             $session->getQuote()->setIsActive($setQuoteAsActive)->save();
             $this->_redirect($successPath, ['_query' => ['utm_nooverride' => '1']]);
-
         } else {
             $this->_adyenLogger->addAdyenResult(
                 sprintf(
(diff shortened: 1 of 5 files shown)


acampos1916 (approved):

cjnewbs:
Looks like the `MAGENTO_USERNAME` and `MAGENTO_PASSWORD` environment variables in CI are not populated 😱

peterojo (approved):

acampos1916:
> Looks like the `MAGENTO_USERNAME` and `MAGENTO_PASSWORD` environment variables in CI are not populated 😱

Hey @cjnewbs this is correct, the variables are only available to members of the org.

cjnewbs:
@acampos1916 Fair enough, I haven't used GitHub actions much so wasn't sure if it was something I was doing wrong. It occurred to me after it got merged that PHPCS is not actually part of your CI flow, not sure if that is something you maybe want to add as an additional step at some point.

Thanks for accepting my PR 🙂

Issue #1168, opened 2021-10-07
[PW-5546] Error: Call to undefined method PayPal\Braintree\Gateway\Data\Order\OrderAdapter


draper87:
**Describe the bug**
On a clean install of Magento 2.4.3 with no other third-party plugins other than Adyen 7.2.0, when we try to place an order with Adyen the result is a generic error message "Something went wrong with your request. Please try again later.". After checked the log file var/log/system.log we've discovered this:

`main.CRITICAL: Error: Call to undefined method PayPal\Braintree\Gateway\Data\Order\OrderAdapter::getQuoteId() in /var/www/html/vendor/adyen/module-payment/Gateway/Response/StateDataCleanupHandler.php:48
Stack trace:
`

After checking di.xml of Adyen It seems to me that both Adyen and Braintree are using a preference over the same Class
`Magento\Payment\Gateway\Data\Order\OrderAdapter`

I'm not sure what's going on here, how can I solve this without uninstall Braintree (actually I tried to disable Braintree's modules and Adyen stop to work).

**To Reproduce**
1. After adding a product to cart go to Checkout and fill customers data.
2. On payment page fill credit card number, expiry Date and CVC number.
3. Click on Place Order.

**Expected behavior**
I'm expecting to place the order without getting any error.

**Magento version**
2.4.3

**Plugin version**
7.2.0 (test mode)

**Screenshots**
![Screenshot from 2021-10-07 10-51-24](https://user-images.githubusercontent.com/61914749/136352053-bd637f7b-8282-4b30-a7cd-1852c6ce93cc.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - Browser Chrome 94

acampos1916:
Hello @draper87, thank you for this report. Maybe this could happen if the Braintree module is loaded after Adyen's, even if disabled. We are investigating this to find a different solution to getting the quoteId.

Djohn12:
Hello,

I can see this issue is closed but we're still reproducing the bug with backend orders on a magento 2.4.3-p1 instance using adyen/module-payment 8.1.1

Error log is : 
```
Error: Call to undefined method PayPal\Braintree\Gateway\Data\Order\OrderAdapter::getQuoteId() in /magento_code/vendor/adyen/module-payment/Gateway/Request/CcBackendAuthorizationDataBuilder.php:54
```

Do you have any workaround or fix for this issue ?
Thanks in advance

igi8819:
[...]

PR #1176, opened 2021-10-11
#1168: Fix for [PW-5546] Error: Call to undefined method PayPal\Brain…

diazwatson:
**Description**

Changes the module sequence to fix the raise condition generated when Paypal also tries to set a preference.
**Tested scenarios**

**Fixed issue**:  
https://github.com/Adyen/adyen-magento2/issues/1168

Credit goes to @cmtickle


diazwatson:
Hi @acampos1916 from what I see, #1182 does not solve the issue Commerce users face when installing the module.
It means the module will be still broken for everyone on Magento Commerce. 🤔

acampos1916:
Thanks @diazwatson, we'll check it out and will come back to you.

acampos1916:
@diazwatson we're going to apply your suggested fix and will later move the state data clean to a cron to remove the module sequence completely, thanks!

diazwatson:
Hey @acampos1916 I think Magento will just skip this sequence if modules:

- Amazon_Core
- PayPal_Braintree

are not present in the target installation, also neither of these modules is something Adyen module needs to work, so I don't think these need to be included in the composer file as dependencies.

acampos1916:
Closing this PR now as it has been addressed with #1192

Issue #1174, opened 2021-10-11
Payment refused generates 2 `order_canceled` compensation


jonathanribas:
**Describe the bug**

3DS 2.0 payment refused generates 2 `order_canceled` compensation, it means we oversell items.

**To Reproduce**
Steps to reproduce the behaviour:
1. Place an order with 3D2.0 card forcing a wrong authentification. We want a Payment Refused here.
2. Order is created in backoffice with a Payment Review status

You can see order is placed and cancelled on MSI `inventory_reservation` table:

MySQL [eu]> select * from inventory_reservation where metadata like '%2268272%';
+----------------+----------+-----+----------+-----------------------------------------------------------------------------+
| reservation_id | stock_id | sku | quantity | metadata                                                                    |
+----------------+----------+-----+----------+-----------------------------------------------------------------------------+
|             40 |        5 | 324 |  -1.0000 | {"event_type":"order_placed","object_type":"order","object_id":"2268272"}   |
|             41 |        5 | 324 |   1.0000 | {"event_type":"order_canceled","object_type":"order","object_id":"2268272"} |
+----------------+----------+-----+----------+-----------------------------------------------------------------------------

4. After Adyen notification cron job runs, it cancels the order and creates a new `order_canceled` compensation which increase saleable quantity

-1.000 + 1.0000 + 1.0000 = 1.0000

In this case, we do overselling! 

MySQL [eu]> select * from inventory_reservation where metadata like '%2268272%';
+----------------+----------+-----+----------+-----------------------------------------------------------------------------+
| reservation_id | stock_id | sku | quantity | metadata                                                                    |
+----------------+----------+-----+----------+-----------------------------------------------------------------------------+
|             40 |        5 | 324 |  -1.0000 | {"event_type":"order_placed","object_type":"order","object_id":"2268272"}   |
|             41 |        5 | 324 |   1.0000 | {"event_type":"order_canceled","object_type":"order","object_id":"2268272"} |
|             42 |        5 | 324 |   1.0000 | {"event_type":"order_canceled","object_type":"order","object_id":"2268272"} |
+----------------+----------+-----+----------+-----------------------------------------------------------------------------

**Expected behavior**

We should have only one `order_canceled` compensation. This way we are clean and we don't oversell items.

**Magento version**
Adobe Commerce 2.3.7 with MSI

**Plugin version**
7.0.1

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOSX
 - Browser Chrome
[...]

PR #1177, opened 2021-10-12
[PW-5637] Save order after cancellation of sync 3DS2 refused response

acampos1916:
**Description**

When a 3DS2 payment is refused the order is set to `canceled` but it wasn't being persisted.

**Tested scenarios**

Refused 3DS2 payments set the order to `canceled`.

**Fixed issue**:  #1174 PW-5637

### DIFF
--- a/Helper/Address.php
+++ b/Helper/Address.php
@@ -24,10 +24,25 @@
 
 namespace Adyen\Payment\Helper;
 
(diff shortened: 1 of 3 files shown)


jonathanribas:
Hi @acampos1916, I will test it but I have a question:
When this order cancellation will happen, will customer stay on checkout at payment step or he will be redirected to cart?

PR #1182, opened 2021-10-12
[PW-5546] - Verify if OrderAdapter passed is an instance of the Adyen OrderAdapter

Morerice:
**Description**
This ticket will add a verification in the `StateDataCleanupHandler` in order to verify that the obtained **OrderAdapter** is the one used by the Adyen plugin and not one linked to another third party module. If the expected class is not received, log a warning message and do not attempt to delete the entry in the `adyen_state_data` table

**Tested scenarios**
* Correct adapter is passed
* Message is logged when another adapter is passed

### DIFF
--- a/Gateway/Response/StateDataCleanupHandler.php
+++ b/Gateway/Response/StateDataCleanupHandler.php
@@ -23,7 +23,9 @@
 
 namespace Adyen\Payment\Gateway\Response;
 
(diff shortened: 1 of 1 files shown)


acampos1916:
FYI @Morerice https://github.com/Adyen/adyen-magento2/pull/1176#issuecomment-942137865 let's chat about further solutions for this.

PR #1212, opened 2021-10-26
[PW-5530] Remove paywithgoogle and googlepay 

AlexandrosMor:
**Description**

The reason to remove the `googlepay` or `paywithgoogle` is that the for google pay and apple pay we don't return anything related to google or apple since the actual payment is done with card. These variants are not included in the payments response or in the notification. The returned payment method is the cc type for example `visa` or `mc`

**Tested scenarios**

Tested with :
- paywithgoogle
- inspect the notification
- inspect the response
- Google pay token is set as visible false

**Fixed issue**:


acampos1916 (commented):
Also, please add tested scenarios to the PR, thanks!

Morerice (approved):

Gateway/Request/CheckoutDataBuilder.php
```diff
@@ -185,7 +185,7 @@ public function build(array $buildSubject)
         }
 
         $requestBody['additionalData']['allow3DS2'] = true;
-
+        //$requestBody['additionalData']['RequestedTestAcquirerResponseCode'] = 2;
```
acampos1916, line 188:
```suggestion
```

AlexandrosMor: oh no ! that was from testing amazonpay. Thanks @acampos1916

Helper/Vault.php
```diff
@@ -233,4 +235,14 @@ private function getExtensionAttributes(InfoInterface $payment)
         }
         return $extensionAttributes;
     }
+
+    /**
+     * @param $additionalData
+     * @return bool
+     */
+    private function isGooglePay($additionalData): bool
```
michaelpaul, line 243: What about moving this method to `\Adyen\Payment\Helper\PaymentMethods` so that we can reuse it?

AlexandrosMor: For now it is not used from other class and `\Adyen\Payment\Helper\PaymentMethods` I think is be responsible for  fetching the payment methods and functionality around it

michaelpaul: I saw that you also added an `isGooglePay` method in `Model/Billing/Agreement.php`.

AlexandrosMor: Check again ?

michaelpaul: Looks good to me! 👍

michaelpaul (dismissed):

Morerice (approved):

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -81,6 +81,7 @@ define(
                 'paypal',
                 'applepay',
                 'paywithgoogle',
+                'googlepay',
```
AlexandrosMor, line 84:
```suggestion
     
```

Helper/Vault.php
```diff
@@ -147,12 +147,6 @@ private function getVaultPaymentToken($payment, array $additionalData): PaymentT
             );
 
             $paymentToken->setGatewayToken($additionalData[self::RECURRING_DETAIL_REFERENCE]);
-
-            if (strpos($additionalData[self::PAYMENT_METHOD], "paywithgoogle") !== false
-                && !empty($additionalData['paymentMethodVariant'])) {
-                $additionalData[self::PAYMENT_METHOD] = $additionalData['paymentMethodVariant'];
-                $paymentToken->setIsVisible(false);
```
acampos1916, line 154: Removing this means that payments made with Google Pay will still be stored, just as not visible?
Google Pay supports recurring payments https://docs.adyen.com/payment-methods/google-pay/api-only#recurring-payments

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -81,6 +81,7 @@ define(
                 'paypal',
                 'applepay',
                 'paywithgoogle',
+     
                 'amazonpay'
```
acampos1916, line 85: `googlepay` payment method code still needs to show the pay button right?

acampos1916 (commented):
Left a few comments.

Helper/Vault.php
AlexandrosMor: I tested and the google pay  token is already set as visible false

PR #1214, opened 2021-10-27
[PW-5660] Installments adding to grand total the taxes

AlexandrosMor:
**Description**

 
The total amount displayed in the installment's section under the credit card is different with the total amount because it is not adding the taxes. This is solved by adding the correct grand total from magento quote.

**Tested scenarios**

- Order with tax
- Order without tax 
- Shipping
- Discount products
**Fixed issue**:


acampos1916 (commented):
Will it also include the +shipping and -discount amounts correctly?

AlexandrosMor:
@acampos1916 Yes, It calculates the tax with shipping and for discount items

acampos1916 (approved):

peterojo (approved):

PR #1215, opened 2021-10-27
[PW-5668] Patch for OFFER_CLOSED flow with cc

### LINKED ISSUE #1193: [PW-5668] Orders staying in Payment Review after OFFER_CLOSED since 7.0.1 -> 7.2.0 upgrade
**Describe the bug**
This has been occuring since upgrading to 7.2.0 from 7.0.1 - OFFER_CLOSED notifications should cancel the related order, however the order is staying in Payment Review.

Status history is added to the order RE the closed offer, however the order stays open.

Our downstream systems listen to the state of the M2 order and therefore we have a building list of orders that are pending in our ERP. 

**To Reproduce**
Place order through Adyen, wait for OFFER_CLOSED, Magento accepts the notification but doesn't cancel the order.

**Expected behavior**
Order is to be cancelled with the acknowledgement of the OFFER_CLOSED notification.

**Magento version**
2.3.7-p1

**Plugin version**
7.2.0

**Screenshots**

image

tnaber:
**Description**
When checking whether an OFFER_CLOSED notification should trigger a state change, alternative payment method names from the notification are compared with the original payment method name from the order, these should match. This also happened when the payment method was a type of cc, as these strings do not match with the notification strings (e.g. 'VI' vs 'visa') the OFFER_CLOSED notification did not trigger a state change. This PR fixes this.

To think about: should these cc strings be compared anyway or should this step be skipped as a whole (like it is now)?

**Tested scenarios**
- Sending an OFFER_CLOSED notification before the authorisation.
- OFFER_CLOSED for adyen_hpp closes the order.
- OFFER_CLOSED for adyen_cc closes the order.
- OFFER_CLOSED for wallet methods closes the order.

**Fixed issue**: 
Closes #1193

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -1212,12 +1212,18 @@ class Cron
                  * Returns if the payment method is wallet like wechatpayWeb, amazonpay, applepay, paywithgoogle
                  */
                 $isWalletPaymentMethod = $this->paymentMethodsHelper->isWalletPaymentMethod($orderPaymentMethod);
+
+                /*
+                 * Return if payment method is cc like VI, MI
+                 */
+                $isCCPaymentMethod = $this->_order->getPayment()->getMethod() === 'adyen_cc';
+
                 /*
                 * If the order was made with an Alternative payment method,
                 *  continue with the cancellation only if the payment method of
                 * the notification matches the payment method of the order.
                 */
-                if ( !$isWalletPaymentMethod && strcmp($notificationPaymentMethod, $orderPaymentMethod) !== 0) {
+                if ( !$isWalletPaymentMethod && !$isCCPaymentMethod && strcmp($notificationPaymentMethod, $orderPaymentMethod) !== 0) {
                     $this->_adyenLogger->addAdyenNotificationCronjob(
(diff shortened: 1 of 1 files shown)


acampos1916:
@tnaber Looks good. Just to make sure, can you test and add to the tested scenarios the following?

OFFER_CLOSED for adyen_hpp closes the order.
OFFER_CLOSED for adyen_cc closes the order.
OFFER_CLOSED for wallet methods closes the order.

acampos1916 (changes requested):
Just a clarification of the tested scenarios.

tnaber:
Scenarios tested and passed.

acampos1916 (approved):

peterojo (approved):

AlexandrosMor (approved):

dufrygrant:
Is there a reason this isn't being merged into the 7.x branch for those of us trying to maintain a stable platform and not needing to upgrade major versions every other month just to have a vendor-supported fix for these (ie instead of us doing/supporting manual patches to the module?)

tnaber:
This will also be implemented in an updated release of v7 :)

PR #1216, opened 2021-10-28
remove comma in sprintf, in order to support php version < 7.3 and adding imageUrl and productUrl to open invoice lines.

daanvinken:
**Description**
- Related to [PR-1200](https://github.com/Adyen/adyen-magento2/pull/1200). Same issue in sprintf for older php versions. 

**Tested scenarios**
- Ran default unit tests.

### DIFF
--- a/Gateway/Request/CheckoutDataBuilder.php
+++ b/Gateway/Request/CheckoutDataBuilder.php
@@ -211,6 +211,27 @@ class CheckoutDataBuilder implements BuilderInterface
     }
 
     /**
+     * @param string $item
+     * @return string
+     */
+    protected function retrieveImageUrl($item): string
+    {
+        $objectManager =\Magento\Framework\App\ObjectManager::getInstance();
+        $helperImport = $objectManager->get('\Magento\Catalog\Helper\Image');
+        $product = $item->getProduct();
+
+        $imageUrl = "";
+
+        if ($image = $product->getSmallImage()) {
+            $imageUrl = $helperImport->init($product, 'product_page_image_small')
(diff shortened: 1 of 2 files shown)


michaelpaul (dismissed):

acampos1916 (changes requested):
I see some of these changes are already being introduced in a different PR, committed by mistake maybe?

daanvinken:
Hi @acampos1916 ,

True, apologies.
As PR #1220 requires some changes, I will close PR #1220 such that we have two separate pull requests.

Cheers

Gateway/Request/CheckoutDataBuilder.php
```diff
@@ -210,6 +210,27 @@ private function getSelectedDebitBrand($brand)
...
     }
 
+    /**
+     * @param string $item
+     * @return string
+     */
+    protected function retrieveImageUrl($item): string
+    {
+        $objectManager =\Magento\Framework\App\ObjectManager::getInstance();
+        $helperImport = $objectManager->get('\Magento\Catalog\Helper\Image');
```
peterojo, line 220: @daanvinken please avoid use of the Object Manager and add `Magento\Catalog\Helper\Image` as a dependency to the constructor instead.
Also, please update the description of this PR since it is no longer only about removing commas

daanvinken: Will do. Apologies I was accidentally working in the same branch while pushing changes for Klarna. 
I scheduled some time Thursday 11-11 to clean this up.

PR #1218, opened 2021-10-29
[PW-5424] adyen giving web component

peterojo:
**Description**

Add adyen giving component to regular checkout page for supported payment methods.
- Save `donationToken` from adyen payments response 
- render giving component if enabled and donation token exists
- handle response from onDonate event and show result to shopper

Before giving
Screen Shot 2021-11-02 at 5 31 05 PM
After giving
Screen Shot 2021-11-02 at 5 36 17 PM

**Tested scenarios**

- CC, 3DS1 & 2 success and cancel flows
- Giving component is not rendered if disabled in administration
- Giving component is not rendered for non-supported PMs

### DIFF
--- a/Controller/Process/Result.php
+++ b/Controller/Process/Result.php
@@ -247,6 +247,11 @@ class Result extends \Magento\Framework\App\Action\Action
             ]
         );
 
+        // Save PSP reference from the response
+        if (!empty($response['pspReference'])) {
+            $this->payment->setAdditionalInformation('pspReference', $response['pspReference']);
+        }
+
         // Save payment token if available in the response
         if (!empty($response['additionalData']['recurring.recurringDetailReference']) &&
             $this->payment->getMethodInstance()->getCode() !== \Adyen\Payment\Model\Ui\AdyenOneclickConfigProvider::CODE) {
@@ -257,6 +262,11 @@ class Result extends \Magento\Framework\App\Action\Action
                 $this->_adyenHelper->createAdyenBillingAgreement($order, $response['additionalData']);
             }
             $this->orderResourceModel->save($order);
+        }
+
+        // Save donation token if available in the response
+        if (!empty($response['donationToken'])) {
+            $this->payment->setAdditionalInformation('donationToken', $response['donationToken']);
         }
 
         // update the order
--- a/Helper/Config.php
+++ b/Helper/Config.php
@@ -214,7 +214,7 @@ class Config
 
     public function adyenGivingEnabled($storeId)
     {
-        return $this->getConfigData('active', 'adyen_giving', $storeId);
+        return $this->getConfigData('active', self::XML_ADYEN_GIVING_PREFIX, $storeId);
     }
 
     public function getAdyenGivingConfigData($storeId)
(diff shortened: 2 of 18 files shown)


Helper/Config.php
```diff
@@ -212,6 +212,53 @@ public function getPWAOriginUrl($storeId)
         return $this->getConfigData(self::XML_PAYMENT_ORIGIN_URL, self::XML_ADYEN_ABSTRACT_PREFIX, $storeId);
     }
 
+    public function adyenGivingEnabled($storeId)
+    {
+        return $this->getConfigData('active', 'adyen_giving', $storeId);
```
acampos1916, line 217: `adyen_giving` is in a constant right?

etc/adminhtml/system/adyen_giving.xml
```diff
@@ -66,7 +66,7 @@
             <label>Suggested donation amounts</label>
             <backend_model>Adyen\Payment\Model\Config\Backend\DonationAmounts</backend_model>
             <config_path>payment/adyen_giving/donation_amounts</config_path>
-            <tooltip><![CDATA[Suggested amounts that the shopper can choose to donate. The minimum amount is 1 EUR.]]></tooltip>
+            <tooltip><![CDATA[Suggested amounts that the shopper can choose to donate. Amount should be specified in minor units, e.g 100 = 1 EUR. The minimum amount is 1 EUR.]]></tooltip>
```
acampos1916, line 69: I think that asking for minor units here can be confusing, is there something we could do to format it when it is about to be used? Using `\Adyen\Payment\Helper\Data::formatAmount` for example.

view/frontend/templates/checkout/success.phtml
```diff
@@ -45,4 +45,69 @@
...
+<script>
+    <?=$scriptString?>
+    require([
+        'jquery',
+        'Adyen_Payment/js/bundle',
+        'Adyen_Payment/js/model/adyen-payment-service'
+    ], function ($, adyenComponent, adyenPaymentService) {
+        function handleOnDonate(state, component) {
+            state.isValid // True or false. Specifies whether the shopper has selected a donation amount.
+            if (state.isValid) {
```
acampos1916, line 65: Seems like a typo
```suggestion
            // True or false. Specifies whether the shopper has selected a donation amount.
            if (state.isValid) {
```

acampos1916 (commented):
Left a few comments.

view/frontend/templates/checkout/success.phtml
peterojo: addressed in 0c4e426

etc/adminhtml/system/adyen_giving.xml
peterojo: addressed in 0c4e426

Helper/Config.php
peterojo: addressed in 0c4e426

Controller/Process/Result.php
```diff
@@ -259,6 +264,11 @@ protected function validateResponse($response)
             $this->orderResourceModel->save($order);
         }
 
+        // Save donation token if available in the response
+        if (!empty($response['donationToken'])) {
```
Morerice, line 268: Can you create some ENUMs with the response parameters used?

Morerice (approved):

AlexandrosMor (approved):

PR #1220, opened 2021-10-29
Klarna product and image url

daanvinken:
**Description**
The new API supports `productUrl` and `imageUrl` for Klarna. These will then be shown on the Klarna page after redirect. I've added a method to retrieve the imageUrl, which sends an empty string if no image was found. Additionally included both values in the lineItems.

**Tested scenarios**
- Klarna pay later payment
- Afterpay payment (does not support it but is ignored by API)
- Default unit tests

### DIFF
--- a/Gateway/Request/CheckoutDataBuilder.php
+++ b/Gateway/Request/CheckoutDataBuilder.php
@@ -222,9 +222,9 @@ class CheckoutDataBuilder implements BuilderInterface
 
         $imageUrl = "";
 
-        if ($product->getSmallImage()) {
+        if ($image = $product->getSmallImage()) {
             $imageUrl = $helperImport->init($product, 'product_page_image_small')
-                ->setImageFile($product->getSmallImage())
+                ->setImageFile($image)
                 ->getUrl();
         }
 
--- a/Gateway/Request/CheckoutDataBuilder.php
+++ b/Gateway/Request/CheckoutDataBuilder.php
@@ -211,6 +211,27 @@ class CheckoutDataBuilder implements BuilderInterface
     }
 
     /**
+     * @param string $item
+     * @return string
+     */
+    protected function retrieveImageUrl($item): string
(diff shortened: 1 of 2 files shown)


Gateway/Request/CheckoutDataBuilder.php
```diff
@@ -210,6 +210,27 @@ private function getSelectedDebitBrand($brand)
         return null;
     }
 
+    /**
+     * @param string $item
+     * @return string
+     */
+    protected function retrieveImageUrl($item): string
+    {
+        $objectManager =\Magento\Framework\App\ObjectManager::getInstance();
```
acampos1916, line 219: We shouldn't use ObjectManager, dependency injection would be the best approach to use the `\Magento\Catalog\Helper\Image` class here https://devdocs.magento.com/guides/v2.4/extension-dev-guide/depend-inj.html#constructor-injection

Gateway/Request/CheckoutDataBuilder.php
```diff
@@ -262,7 +282,9 @@ protected function getOpenInvoiceData($order): array
                 'description' => $item->getName(),
                 'quantity' => $numberOfItems,
                 'taxCategory' => $item->getProduct()->getAttributeText('tax_class_id'),
-                'taxPercentage' => $formattedTaxPercentage
+                'taxPercentage' => $formattedTaxPercentage,
+                'productUrl'=> $item->getProduct()->getUrlModel()->getUrl($item->getProduct()),
```
acampos1916, line 286: Have you tested this with products in different storeviews so that the correct URL is used?

acampos1916 (changes requested):
Nice contribution @daanvinken, just a couple of comments

daanvinken:
As this PR requires some changes and a commit of PR #1216 is still in here, I'll close this PR and create a new one later this week.

Issue #1223, opened 2021-11-03
[Storefront] returnUrl field in Magento's Dashboard


Fifciu:
**Is your feature request related to a problem? Please describe.**
In plugin's configuration we have 1 field and it's value is used for two purposes:
- as an `origin`
- to build `returnUrl`

It's problematic if the endpoint in our storefront starts with some prefix and we cannot change it as our system forces it. 

I have to redirect back after RedirectShopper process to the: `http://localhost:3000/api/adyen/process/result?someQueryData`. `/api` is our obligatory prefix. But its impossible for me to put there prefix part.    

If I set `Payment Origin URL` as `http://localhost:3000/api` then `orgin` is wrong.    
If I set `Payment Origin URL` as `http://localhost:3000` then `returnUrl` is wrong.

**Describe the solution you'd like**
A dedicated field for setting `returnUrl`.
So instead of:
```php
        $returnUrl = rtrim(
                $this->baseUrlHelper->getStoreBaseUrl($this->storeManager->getStore()->getId()), '/'
            ) . '/adyen/process/result?merchantReference=' . $order->getIncrementId();
```
We would have:
```php
$returnUrl = rtrim(
                $valueOfNewField), '/'
            ) . '?merchantReference=' . $order->getIncrementId();
```

acampos1916:
Thanks for opening this issue and for the contribution @Fifciu, #1224 looks very nice.

What is the use case for the `/api` part of your returnUrl? Maybe this is an issue very specific to your implementation and it that case there wouldn't be much use for other integrations.

Fifciu:
Each endpoint that does not render a view, starts with `/api` prefix so it does not pollute the main scope and it's just a space for Express API, independent from the frontend. The endpoint does its job and then redirects to the confirmation view in the frontend. 

It's the case especially related to PWAs - it doesn't have point to render additional view in PWA just to send a request and redirect. Secondly, different PWAs systems might have different requirements so giving the flexibility to provide returnUrl is very handy for every storefront integration. [Commercetools & Adyen integration](https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/docs/WebComponentsIntegrationGuide.md#step-5-make-a-payment) allows it and it works like a charm.

acampos1916:
Sounds good @Fifciu, then if I get it correctly, in your integration the `/api` space will forward the request to the Adyen plugin Result.php controller?

Fifciu:
[...]

PR #1224, opened 2021-11-03
feat: returnUrl field is settable via dashboard

Fifciu:
**Description**

It solves #1223 issue.

**Tested scenarios**
I manually tested it in my Storefront - returnUrl is properly set. Also I did test in M2 Frontend and it works fine.

### DIFF
--- a/etc/adminhtml/system/adyen_pwa.xml
+++ b/etc/adminhtml/system/adyen_pwa.xml
@@ -40,5 +40,11 @@
             <config_path>payment/adyen_abstract/payment_origin_url</config_path>
         </field>
 
+        <field id="payments_return_url" translate="label" type="text" sortOrder="11" showInDefault="1" showInWebsite="1" showInStore="1">
+            <label>Payment Return URL</label>
+            <tooltip>Only relevant if you process payments from an external URL different to that of Magento</tooltip>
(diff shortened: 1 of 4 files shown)


Fifciu:
@acampos1916 Will we merge it? I am going to release first version of Vue Storefront integration with Adyen Magento2 and I need this PR.

etc/adminhtml/system/adyen_pwa.xml
```diff
@@ -40,5 +40,11 @@
             <config_path>payment/adyen_abstract/payment_origin_url</config_path>
         </field>
 
+        <field id="payments_return_url" translate="label" type="text" sortOrder="11" showInDefault="1" showInWebsite="1" showInStore="1">
+            <label>Payment Return URL</label>
+            <tooltip>Only relevant if you process payments from an external URL different to that of Magento</tooltip>
```
peterojo, line 45: How about we make this a bit clearer
```suggestion
            <tooltip>The URL that redirect payment methods will return to after the shopper completes a payment in a 3rd party gateway. Default return URL endpoint is {YOUR_MAGENTO_BASE_URL}/adyen/process/result</tooltip>
```

michaelpaul (approved):

PR #1225, opened 2021-11-03
[PW-5332] - Set the adyen configured order status on: Manual Review 

Morerice:
**Description**
Set the adyen configured order status when an order is being manually reviewed and once this review is successful. The `fraud_manual_review_status` status is currently not being set anywhere in the code, whereas the `fraud_manual_review_accept_status` seems to be correctly used in the `Cron.php` file. Check [the docs](https://docs.adyen.com/risk-management/case-management#notifications) to verify how Adyen will tell us that the payment has to go trough manual review

**Tested scenarios**
* Manual review status set when the first manual review notification is received


acampos1916 (approved):

AlexandrosMor (dismissed):

AlexandrosMor:
@Morerice Just noticed there are 3 code smells

Morerice:
> @Morerice I just noticed there are 3 code smells

I fixed two of them and updated the if condition to be more readable.

acampos1916 (dismissed):

Model/Cron.php
```diff
@@ -1532,47 +1532,45 @@ protected function _authorizePayment()
...
-                'Ignore the pre authorized status because the order is ' .
-                'under manual review and use the Manual review status'
-            );
-        }
-
-        if ($isTotalAmountAuthorised) {
+        // IF manual review is NOT active OR no custom status is set on the Adyen config page AND total amount is authorised continue with authorisation
+        // ELSEIF total amount is NOT authorised
+        // ELSE add a comment and set the status to the defined fraudManualReviewStatus
+        if ($this->requireFraudManualReview || $fraudManualReviewStatus == "" && $isTotalAmountAuthorised) {
```
peterojo, line 1545: Could you use parentheses here so that we can be sure which condition is evaluated first.
```suggestion
        if (($this->requireFraudManualReview || $fraudManualReviewStatus == "") && $isTotalAmountAuthorised) {
```

Morerice: Updated in 88f808b

peterojo (approved):

acampos1916 (approved):

PR #1228, opened 2021-11-05
PW-5541: Allow notification username/password configuration per store

michaelpaul:
**Description**

This change allows the merchant to set a different webhook username/password per store. The other related settings like HMAC and IP check already supported multi-store.  

**Tested scenarios**

* Multi-store setup with different credentials

### DIFF
--- a/etc/adminhtml/system/adyen_required_settings.xml
+++ b/etc/adminhtml/system/adyen_required_settings.xml
@@ -38,12 +38,12 @@
             <config_path>payment/adyen_abstract/demo_mode</config_path>
             <tooltip><![CDATA[ In the test mode you must use test cards. See section Documentation & Support for the link to the test cards]]></tooltip>
         </field>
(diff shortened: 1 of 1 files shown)


Morerice (approved):

acampos1916:
These functions to fetch the user and password don't seem to use the storeId https://github.com/Adyen/adyen-magento2/blob/develop-6/Controller/Process/Json.php#L307-L308 how is the store-specific config fetched?

michaelpaul:
@acampos1916 Magento fetches the current store configuration by default when store ID is null. For example, here it's how `\Magento\Store\Model\StoreManager` works https://github.com/magento/magento2/blob/2.4.0/app/code/Magento/Store/Model/StoreManager.php#L155.

AlexandrosMor (approved):

peterojo (approved):

PR #1229, opened 2021-11-05
[PW-5712] - Check address issue on plugin v6

Morerice:
**Description**
This PR updates the regex used to identify the house number when it is entered in the same line as the street. Currently the houseNumber will be correctly sent to Adyen only if it is entered after the street. This PR will cater for the case where the number is also entered before the street.

Also when finalizing an invoice, check if the pay() method was called.

**Tested scenarios**
* House number entered before street
* House number entered after street

### DIFF
--- a/Helper/Invoice.php
+++ b/Helper/Invoice.php
@@ -118,7 +118,7 @@ class Invoice extends AbstractHelper
         foreach ($invoiceCollection as $invoice) {
             $parsedTransId = $this->adyenDataHelper->parseTransactionId($invoice->getTransactionId());
             // Loose comparison based on how Magento does the comparison in entity
-            if ($invoice->getState() == InvoiceModel::STATE_OPEN) {
+            if ($invoice->getState() == InvoiceModel::STATE_OPEN || !$invoice->wasPayCalled()) {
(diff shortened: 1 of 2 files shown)


Helper/Invoice.php
```diff
@@ -118,7 +118,7 @@ public function finalizeInvoices(Order $order, Notification $captureNotification
         foreach ($invoiceCollection as $invoice) {
             $parsedTransId = $this->adyenDataHelper->parseTransactionId($invoice->getTransactionId());
             // Loose comparison based on how Magento does the comparison in entity
-            if ($invoice->getState() == InvoiceModel::STATE_OPEN) {
+            if ($invoice->getState() == InvoiceModel::STATE_OPEN || !$invoice->wasPayCalled()) {
```
acampos1916, line 121: This probably is unrelated to the PR, should we mention it in the PR description?

acampos1916 (approved):

Helper/Invoice.php
Morerice: Updated description

peterojo (approved):

PR #1232, opened 2021-11-08
[PW-3872] Feature/adyen giving

peterojo:
**Description**

Add Adyen Giving

**Tested scenarios**


etc/adminhtml/system/adyen_giving.xml
```diff
@@ -0,0 +1,80 @@
...
+            <backend_model>Magento\Config\Model\Config\Backend\Image\Logo</backend_model>
+            <upload_dir config="system" scope_info="1">adyen</upload_dir>
+            <base_url type="media" scope_info="1">adyen</base_url>
+            <comment><![CDATA[Logo displayed on the Adyen Giving form.]]></comment>
+        </field>
+        <field id="donation_amounts" translate="label" type="text" sortOrder="40" showInDefault="1" showInWebsite="1" showInStore="1">
+            <label>Suggested donation amounts</label>
+            <backend_model>Adyen\Payment\Model\Config\Backend\DonationAmounts</backend_model>
+            <config_path>payment/adyen_giving/donation_amounts</config_path>
+            <tooltip><![CDATA[Suggested amounts that the shopper can choose to donate. The minimum amount is 1 EUR.]]></tooltip>
```
acampos1916, line 69:
```suggestion
            <tooltip><![CDATA[Suggested amounts that the shopper can choose to donate separated by commas. The minimum amount is 1 EUR.]]></tooltip>
```

Block/Checkout/Success.php
```diff
@@ -116,6 +141,50 @@ public function getAction()
...
+        return json_encode($this->getOrder()->getPayment()->getAdditionalInformation('donationToken'));
+    }
+
+    public function getDonationComponentConfiguration(): array
+    {
+        $storeId = $this->storeManager->getStore()->getId();
+        $imageBaseUrl = $this->storeManager->getStore()->getBaseUrl(\Magento\Framework\UrlInterface::URL_TYPE_MEDIA).'adyen/';
+        $donationAmounts = explode(',', $this->configHelper->getAdyenGivingDonationAmounts($storeId));
+        $donationAmounts = array_map(function ($amount) {
+            return $this->adyenHelper->formatAmount($amount, $this->getOrder()->getOrderCurrencyCode());
```
acampos1916, line 170: I think this is conflicting since we're requesting the suggested amounts in the config field in EUR. `getAdyenGivingDonationAmounts` will then return amounts in EUR and the order could be in a different currency. How about if we create a helper method that is fed the amounts and the order currency and converts them to the order currency? Similar to how is done in `Model/Config/Backend/DonationAmounts.php`

Gateway/Response/CheckoutDonationsDetailsHandler.php
```diff
@@ -0,0 +1,38 @@
...
+use Magento\Payment\Gateway\Response\HandlerInterface;
+
+class CheckoutDonationsDetailsHandler implements HandlerInterface
+{
+    /**
+     * @inheritDoc
+     */
+    public function handle(array $handlingSubject, array $response)
+    {
+        // @todo: PW-5424 Implement handle() method.
```
peterojo, line 36: this class turned out to be unnecessary in the implementation of #1218 
it can be deleted and removed from the di.xml

Block/Checkout/Success.php
acampos1916: Since this would require a conversion we can park it for now and inform in the config help tooltip that the order currency will be used.

acampos1916 (approved):

acampos1916 (approved):

Block/Checkout/Success.php
```diff
@@ -116,6 +141,50 @@ public function getAction()
...
+    }
+
+    public function adyenGivingEnabled(): bool
+    {
+        return (bool) $this->configHelper->adyenGivingEnabled($this->storeManager->getStore()->getId());
+    }
+
+    public function hasDonationToken()
+    {
+        return $this->getDonationToken() && 'null' !== $this->getDonationToken();
```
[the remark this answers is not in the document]
peterojo: I did it this way because the `->getAdditionalInformation('donationToken')` function could also return a string literal value of `'null'`, for some reason

Helper/Requests.php
```diff
@@ -374,4 +375,20 @@ public function buildRecurringData(int $storeId, $payment, $request = [])
 
         return $request;
     }
+
+    public function buildDonationData($buildSubject, $storeId): array
+    {
+        return [
+            "amount"=> $buildSubject['amount'],
+            "reference"=> Uuid::generateV4(),
+            'shopperReference' => $buildSubject['shopperReference'],
```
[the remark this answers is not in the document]
peterojo: Done in 115aacf

Morerice (dismissed):

acampos1916 (approved):

michaelpaul (approved):

michaelpaul (approved):

Morerice (approved):

Issue #1236, opened 2021-11-10
[PW-4710] Could be possibile to use the Manual review for immediate capture orders? 


mirkocesaro:
I noticed in the https://docs.adyen.com/plugins/magento-2/set-up-the-plugin-in-magento#manual-review section that it is not possible to use the manual review when it is configured the automatic capture.

At the moment, before processing an order, we have to check the Adyen platform to understand if there is a risk on it. 
This action is performed on all orders and it is done to prevent loss of money.

A possible solution could be to handle the manual review on the notification to set the 'Manual Review Status' when a risk is detected by Adyen and to set the 'Manual Review Accepted Status' after the notification of approval. 

In the current situation, the 'Manual Review Status' is set but for some reason, the processing status is back right after it. Is it a bug? As you can see in the following image the "Adyen holded" is my Manual Review Status, but the order's state and status are `processing`
<img width="861" alt="Screenshot " src="https://user-images.githubusercontent.com/1431255/141126004-7ba19957-3922-44f0-8526-d442c103a632.png">

Do you think is it a valid enhancement to work on? 

I also would like a confirmation that it is correct to don't perform any action when someone denies the review
<img width="794" alt="Screenshot deny" src="https://user-images.githubusercontent.com/1431255/141128992-aefedeac-7024-4077-a560-bf04d134bc1c.png">
and wait for the cancel notification.

Thanks,
Mirko

tnaber:
Hi @mirkocesaro,

This is a great suggestion. We have a ticket open for this internally which should be picked up soon. 

It is indeed correct to not perform any actions when someone denies a review. This is because the cancel or refund webhook should then be received by the plugin, this will handle the order cancellation in Magento.

Cheers,
Titus

mirkocesaro:
Hi @tnaber ,
thank you. I am checking the problem and basically, the Manual Review Status configuration uses the Magento\Sales\Model\Config\Source\Order\Status\NewStatus class, and this means that the status for the manual review is linked to the NEW state.

The problem with the immediate capture configuration is the invoice, after the creation of the invoice, Adyen module sets:
``` $invoice->getOrder()->setIsInProcess(true);```
you can see it here: \Adyen\Payment\Model\Cron::_createInvoice 

And this data on the order is used on the save method  to set the order in processing (see here \Magento\Sales\Model\ResourceModel\Order\Handler\State::check)

I am creating a PR with a fix. 

Let me know if it is ok.

Thanks,
Mirko

[...]

PR #1242, opened 2021-11-12
PW-4710 Fix manual review status for immediate capture

mirkocesaro:
**Description**
With this PR I would like to fix the bug on the Manual review not usable when Adyen is configured with immediate capture.

**Tested scenarios**

1. Set the Adyen Magento module and your Adyen Merchant account to use the `immediate` capture
2. Configure the Manual Review Status and the Manual Review Accepted Status configuration on Magento
3. Configure the risk management to require the manual review on the transactions
4. Create an order on Magento
5. Wait for the Adyen notification to be processed
6. Check the state of the order to be NEW and the status in the one configured in step 2 for Manual Review Status 
7. If it is ok, accept the risk on Adyen and wait the notification
8. After the notification is processed, you can see the order in state PROCESSING and status configured in step 2 for Manual Review Accepted Status

**Fixed issue**: https://github.com/Adyen/adyen-magento2/issues/1236

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -2122,6 +2122,10 @@ class Cron
                 if ($fraudManualReviewStatus != "") {
                     $status = $fraudManualReviewStatus;
                     $comment = "Adyen Payment is in Manual Review check the Adyen platform";
+                    //if the order requires the manual review, it is not possible to set the order in process because the
+                    //$fraudManualReviewStatus is a status assigned to NEW state.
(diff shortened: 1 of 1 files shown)


Morerice:
Functionality was updated in https://github.com/Adyen/adyen-magento2/pull/1274. If another issue still occurs, please open another issue relative to the updated code.

Thanks,
Jean
Adyen

PR #1245, opened 2021-11-15
GraphQL support

acampos1916:
This feature was initially added in #1136. Here it is being re-applied for version 8.1.0.


larsroettig:
@acampos1916 what means Version 8.1 when this will be released?

acampos1916:
Hello @larsroettig, this feature and 8.1.0 will be released no sooner than next week. Feel free to use this PR to test the GraphQL support for your project in the meantime.

larsroettig:
Hi @acampos1916, 
I found out we must definitely extend the Magento store config endpoint with the following fields

- Client key for Test (payment/adyen_abstract/client_key_test)
- Client key for Live (payment/adyen_abstract/client_key_live)

- Test/Production Mode  (payment/adyen_abstract/demo_mode)
- Test = 1
- Production Mode  = 0

Bartlomiejsz:
Hi @acampos1916, since this wasn't yet merged into mainline, I wonder if there are still possibilities to change schema of this module.

If possible, I would suggest to add `Adyen` prefix to each of the created input or type, because some of them are pretty general (like `paymentMethods` / `paymentMethodsArray`) so there is high possibility that at some time either some other module, or even Magento core will break this schema by introducing same fields.
Also, I believe according to Magento coding standards (not sure how strongly your codebase is connected to it), all inputs/types in graphql should be `PascalCase`, not `camelCase` - and in your `schema.graphqls` some of them are written in one way, and some in other.

I'm also upvoting @larsroettig comment above - I also had to add those fields manually to create payment method extension in PWA

Best regards

acampos1916:
Thanks for the feedback @Bartlomiejsz, those are valid points. Will include these adjustments in this branch as well as the config field extensions.

acampos1916:
@larsroettig @Bartlomiejsz I've added the suggestions in https://github.com/Adyen/adyen-magento2/pull/1279 and https://github.com/Adyen/adyen-magento2/pull/1280. Let me know if you're able to take a look and test.

Bartlomiejsz:
@acampos1916 looks good, besides three points:
- https://github.com/Adyen/adyen-magento2/pull/1279/files#diff-421c646372437d30d372541e053cbfda85cd596c031f607128706ce143ec2797L82 - this shouldn't be renamed as is extends `Order` type from Magento core
- https://github.com/Adyen/adyen-magento2/pull/1279/files#diff-421c646372437d30d372541e053cbfda85cd596c031f607128706ce143ec2797L86 - same here - `PaymentMethodInput` is also from Magento core and should stay as is
- https://github.com/Adyen/adyen-magento2/pull/1280/files#diff-421c646372437d30d372541e053cbfda85cd596c031f607128706ce143ec2797R129 - `demo_mode` should be also prefixed with `adyen_` since this field is being added to global `StoreConfig` type

acampos1916:
@Bartlomiejsz you're right, applied those adjustments, thanks!

Bartlomiejsz:
Hi @acampos1916, after you merged I noticed that there is one more place that you didn't change - https://github.com/Adyen/adyen-magento2/pull/1245/files#diff-c455c6c81db7e102ae72317546a6e16a5a556addf879c171099cf1cb640296ddR19 should contain also `adyen_demo_mode` as key.
Also, in `schema.graphqls` this field `adyen_demo_mode` should be set to `Boolean` since Magento handles this type automatically when retrieving data from config

larsroettig:
@acampos1916 is there an update on the release date ?

Morerice (approved):

peterojo (approved):

larsroettig:
Hi @acampos1916, i can still repoduce https://github.com/Adyen/adyen-magento2/issues/1267 checkout branch today

acampos1916:
Hey @larsroettig, thanks for letting us know, we'll do a short investigation on that issue you reported.

bramstroker:
We are also working on a PWA for a client of us and want to use the GraphQL API.
I am developing against this feature branch for the meantime, and I also have the same issue as https://github.com/Adyen/adyen-magento2/issues/1267, where the additional data is not save correctly. So when you do a `setPaymentMethodOnCart` mutation first, and `placeOrder` later than the additional data gets lost resulting in the place order failing.

Morerice:
> We are also working on a PWA for a client of us and want to use the GraphQL API. I am developing against this feature branch for the meantime, and I also have the same issue as #1267, where the additional data is not saved correctly. So when you do a `setPaymentMethodOnCart` mutation first, and `placeOrder` later than the additional data gets lost resulting in the place order failing.

Hi @bramstroker,

Did you also have this issue when you tried the request I mentioned in the  #1267 [comment](https://github.com/Adyen/adyen-magento2/issues/1267#issuecomment-1055530313)?

Thanks,
Jean
Adyen

bramstroker:
> Did you also have this issue when you tried the request I mentioned in the #1267 [comment](https://github.com/Adyen/adyen-magento2/issues/1267#issuecomment-1055530313)?
> 
> Thanks, Jean Adyen

@Morerice your approach with combining the two mutations into one request is working fine for me. However it is more cumbersome for us to implement in the frontend because we have a checkout containing multiple steps, so it would be nice when the persistence of the additional data is taken care of correctly when done in multiple requests.

Morerice:
> 

@bramstroker I understand your point. However, we want to move away from persisting some items that are in the additional data, which is why we are also combining these two requests in our current REST [implementation](https://docs.adyen.com/plugins/magento-2/magento-headless-integration#place-an-order). Hence we would like to have a similar flow for GraphQL.

It was originally working in the first 'version' of this PR because at the time we were still persisting all the information

Best regards,
Jean
Adyen

Bartlomiejsz:
Hi @Morerice, the issue with approach that you're suggesting is that it's different than default behavior of Magento PWA Studio (which I believe will be mostly used with your module). In PWA Studio there are two separate steps of checkout, on first the payment method is saved and on second order is being placed.
So it's not possible to use such flow without rebuilding large part of PWA Studio...

larsroettig:
> Hi @Morerice, the issue with approach that you're suggesting is that it's different than default behavior of Magento PWA Studio (which I believe will be mostly used with your module). In PWA Studio there are two separate steps of checkout, on first the payment method is saved and on second order is being placed. So it's not possible to use such flow without rebuilding large part of PWA Studio...


For a secure 3-D process, you need to create your own **placeOrder action** anyway. 

AdyeWorkflowPWAStudio

Morerice:
> @Morerice I understand. I will discuss with my frontend team if we can implement this way. Two other questions related to the GraphQL API which may also clear things up for others implementing this. Do you know if there is some kind of reference implementation for Hyva checkout (which we are using)? Als we are having a look into the adyen web components. https://docs.adyen.com/online-payments/web-components. Looking at the sequence diagram version 5.x is not really compatible with this adyen module and the GraphQL flow, and I should use 4.x, am I right about that? Sorry to bother you with this questions, but the documentation is conflicting and not really clear about this.

Hi @bramstroker,

No worries, feel free to ask any questions you might have. Regarding Hyva we currently have no reference for it. However this is something that is in our pipeline and we would like to have in the future.

Regarding the component, although the doc is unclear I can confirm that v5 also supports the 'older' flow, referenced on [this page](https://docs.adyen.com/online-payments/web-components/integrated-before-5-0-0) so it can also be used in this graphql version.

Best regards,
Jean
Adyen

cornedor:
If you are logged into an account, and start the order process and leave at the payment stage to continue on another device (for example, because you want to pay with a mobile app on your phone), the additional state must be persisted. This requires an additional backend endpoint to store this data in the user. This also applies to abandoned carts which you want to restore later on.

bramstroker:
@Morerice Thanks a lot for your answers, that clears out the skies for us.

Regarding the persistence of additional data. Please also see @cornedor remarks.
I also want to add that it's very confusing and bad practice imo to have an endpoint behaving differently depending on whether you combine two mutations or do a single mutation. These Graphql mutations should behave in a predictable and consistent way. When the input type describes that you can pass adyen additional data you also expect that to be persisted in the cart.

Morerice:
Hi guys,

I understand your concerns and also the use case mentioned by @bramstroker is very relevant. We need to investigate this internally, since ultimately the decision of being able to save (or not save) the stateData is not one that our team has the authority to decide on.

However in the meantime I would prefer if we merge this PR as is. We will then continue this discussion directly on issue #1267. We will obviously let you know of our proposed modification to this PR on that same issue, once we get some additional information about the stateData.

Thanks,
Jean
Adyen

cornedor:
> However in the meantime I would prefer if we merge this PR as is. We will then continue this discussion directly on issue #1267. We will obviously let you know of our proposed modification to this PR on that same issue, once we get some additional information about the stateData.

This might result in a breaking change after only a short time of usage.

Morerice:
> > However in the meantime I would prefer if we merge this PR as is. We will then continue this discussion directly on issue #1267. We will obviously let you know of our proposed modification to this PR on that same issue, once we get some additional information about the stateData.
> 
> This might result in a breaking change after only a short time of usage.

Hi @cornedor

If allowed to store the stateData, it should be possible for us to allow both solutions (the current one and the one where data is stored) simultaneously and mark this one as deprecated. Then on the next major version, we would remove the flow.

Thanks,
Jean
Adyen

PR #1251, opened 2021-11-16
[PW-5344] Change the UI of the overview of Magento2 config

AlexandrosMor:
**Description**

1. Remove different UI (Adyen logos and configure button) for the payment method sections. Keep same header UI.
2. Add new section under Required Settings called "Configure payment methods"
3. Move the credit card, stored payment methods and alternative payment method sections to the new section "Configure payment methods".
4. Change "Adyen all in one payment solution" header to "Adyen Payments"
5. Split the config into 3 sections: Required Settings, Configure payment methods, Advanced Settings (regroup the different advanced settings under this).

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/etc/adminhtml/system.xml
+++ b/etc/adminhtml/system.xml
@@ -33,19 +33,8 @@
                 <frontend_model>Magento\Paypal\Block\Adminhtml\System\Config\Fieldset\Group</frontend_model>
                 <include path="Adyen_Payment::system/adyen_getting_started.xml"/>
                 <include path="Adyen_Payment::system/adyen_required_settings.xml"/>
-                <include path="Adyen_Payment::system/adyen_advanced_order_processing.xml"/>
-                <include path="Adyen_Payment::system/adyen_advanced_notifications.xml"/>
-                <include path="Adyen_Payment::system/adyen_checkout_experience.xml"/>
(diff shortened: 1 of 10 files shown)


acampos1916 (approved):

etc/adminhtml/system/adyen_advanced_notifications.xml
```diff
@@ -24,7 +24,7 @@
 -->
```
peterojo, line 24: I think the word `advanced` is also redundant in the name of this file. It can be renamed to `adyen_payment_notifications.xml`

etc/adminhtml/system/adyen_advanced_order_processing.xml
```diff
@@ -24,7 +24,7 @@
 -->
```
peterojo, line 24: I think the word advanced is also redundant in the name of this file. It can be renamed to adyen_magento_order_processing.xml

etc/adminhtml/system/adyen_advanced_notifications.xml
AlexandrosMor: @peterojo check again ?

peterojo (approved):

acampos1916 (approved):

PR #1252, opened 2021-11-16
POS Cloud HTML receipt is escaped

pmzandbergen:
The receipt itself is HTML and should not be escaped, otherwise raw HTML will be disabled

**Description**
When using POS cloud Adyen will return a HTML receipt. This receipt is displayed in the backend and frontend as encoded HTML instead of HTML.

**Tested scenarios**
Create a POS payment, check frontend and backend.


peterojo (approved):

tnaber (commented):
Thanks for the contribution!

tnaber (approved):

PR #1254, opened 2021-11-17
Added productUrl and imageUrl to openInvoice fields (for Klarna)

daanvinken:
**Description**
The new checkout API supports `productUrl` and `imageUrl` for Klarna. These will then be shown on the Klarna page after redirect. I've added a method to retrieve the imageUrl, which sends an empty string if no image was found. Additionally included both values in the lineItems.

**Tested scenarios**
- Klarna pay later payment
- Afterpay payment (does not support it but values are ignored by API)
- Default unit tests
- Multi store view setup (with different URL).
    - As can be seen, productUrl is still correct and Images are cached from the same URL.

### DIFF
--- a/Gateway/Request/CheckoutDataBuilder.php
+++ b/Gateway/Request/CheckoutDataBuilder.php
@@ -56,6 +56,11 @@ class CheckoutDataBuilder implements BuilderInterface
      */
     private $chargedCurrency;
 
+    /**
+     * @var \Magento\Catalog\Helper\Image
+     */
+    private $imageHelper;
+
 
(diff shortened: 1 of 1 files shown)


acampos1916 (approved):

tnaber (approved):
Nice! However, in my no images appear on Klarna? (I see that the correct data is send so is this a problem on my end?)

daanvinken:
image
It's being shown in the app only. So hard to test for ourselves, however I have a merchant that'd be able to confirm. @tnaber

PR #1262, opened 2021-11-30
[PW-5924] fix check for unrendered LPMs in multishipping

peterojo:
**Description**

Fix how we set the state data for LPMs that do not have a rendered component.
[Currently](https://github.com/Adyen/adyen-magento2/blob/712da7349a0bc9f28ff09d07a03d24fda36072b3/view/frontend/web/js/view/payment/method-renderer/multishipping/adyen-hpp-method.js#L253) we check for the existence of a `component` in the LPM .
But this only works for LPMs which failed to render, because we [try to set](https://github.com/Adyen/adyen-magento2/blob/fbdbfad10b5a80b7c2798d061ec624d2f4560b95/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js#L774) the `component` property for every LPM.
This means that we are unable to click 'Proceed' in multishipping when we select for example Klarna or Bank Transfer.

***Fix:*** Updated the logic in `view/frontend/web/js/view/payment/method-renderer/multishipping/adyen-hpp-method.js` to check whether component is rendered in the page and set the stateData if it isn't rendered.
**Tested scenarios**

- Proceed with banktransfer, stateData is set manually
- Proceed with giftcard, stateData is set manually
- Proceed with Klarna, stateData is set manually
- Proceed with ideal, Sepa direct debit, stateData is generated from component and set


view/frontend/web/js/view/payment/method-renderer/multishipping/adyen-hpp-method.js
```diff
@@ -249,9 +249,10 @@ define([
         selectPaymentMethodType: function() {
             var self = this;
             $('#stateData').val('');
+            let paymentMethod = self.paymentMethod;
             let stateData;
-            if (!('component' in self)) {
-                let paymentMethod = self.paymentMethod;
+            if ($.trim($('#adyen-alternative-payment-container-' +
+                paymentMethod.methodIdentifier).html())==='') {
```
michaelpaul, line 255: What about this way? 
```suggestion
            const isEmptyContainer = $('#adyen-alternative-payment-container-' + paymentMethod.methodIdentifier).children().size() === 0;
            if (isEmptyContainer) {
```

tnaber (dismissed):

view/frontend/web/js/view/payment/method-renderer/multishipping/adyen-hpp-method.js
peterojo: works as well

michaelpaul (dismissed):

AlexandrosMor (dismissed):

michaelpaul (dismissed):

acampos1916 (dismissed):

michaelpaul (approved):

Morerice (approved):

AlexandrosMor (approved):

Issue #1139, opened 2021-08-27
[PW-5393] sales_order_payment table size keeps growing


jonathanribas:
**Is your feature request related to a problem? Please describe.**
As far as our e-commerce grows, we have noticed that `sales_order_payment` table is the largest one on our e-commerce platform.
After taking a look deeper in which fields contains a large amount of data, we have noticed that Adyen module is storing a lot of data on `additional_information` column.

**Describe the solution you'd like**
After placing payment, we may don't need anymore whole data which is stored on `additional_information` column. Most important data is PSP Reference and it is stored in a separate column.

acampos1916:
Hello @jonathanribas, thank you for opening this issue. This indeed could be a good improvement, we'll do an investigation about how to approach the `sales_order_payment` table cleanup after the data has been used.

jonathanribas:
Hi @tnaber, 
Have you guys planned to develop a script in order to move old data to new table?

jonathanribas:
Hi @tnaber, 
Have you guys planned to develop a script in order to move old data to new table?

tnaber:
Hi @jonathanribas, currently we were thinking about adding a function which would check the old location when there is no data available in the new location. This means we were not planning on moving the old data.

Morerice:
Hi @jonathanribas

Since [v8.1.0](https://github.com/Adyen/adyen-magento2/releases/tag/8.1.0) we have stopped saving the `stateData` field in the `sales_order_payment` table. We believe this field was the biggest contributor to the growth of this table.

Do you still see this issue happening when v8.1.0+ is being used?

Thanks,
Jean
Adyen

jonathanribas:
Hi @Morerice, yes it's better but there are still data like `url` that are not needed anymore after an order is placed. This takes place inside this table... For a merchant with a lot of transactions it's really an huge amount of data.
Hope you understand my point.

Morerice:
Hi @jonathanribas,

Yes I completely get your point. However at this point we don't really want to create a php script just for this specific purpose. Would an SQL command which would remove this specific data of orders that were created 3+ months before be something helpful?

Thanks,
Jean
Adyen

jonathanribas:
Hi @Morerice, yes it could be great!

Morerice:
Hi @jonathanribas,

[...]

PR #1265, opened 2021-12-01
[PW-5919] Implement new database to store and adjust dataflow for additionalData

### LINKED ISSUE #1233: [PW-5813] Persist 3DS and liabilityShift data in the payment table
We need to provide liability shift data to a third party integration (I imagine a reasonably common task).  This works for the 3DS2 cc payment flow, but with 3DS1 none of the additional data is persisted to the payment table.  We used to get around this using the 3DSuccess node in additional_information that was persisted in the redirect controller, but that has been removed in 7.x so we now have no way of knowing if liability is shifted for the order.

We would like the additional data (at least regarding liability shift status) to be persisted on the payment table during the 3DS1 flow, probably in additional_information which is where it is persisted for the 3DS2 flow

If there is an alternative to get this data for 3DS1 orders then please let us know.  We would like to avoid additional API requests though.

**Additional context**
Given the above, the ticket to remove the additional information data would actually cause us further problems unless the data removed does not include the liability/3DS information: https://github.com/Adyen/adyen-magento2/issues/1139

tnaber:
**Description**
In response to #1139, we decided to persist the additional data in a different way. Additionally, the additional data will be saved for dataflows (previously not done for hpp flow) to resolve #1233. This includes the additional data fetch by the paymentDetails call.

The PaymentResponseHandler Helper has been expanded to be used in every payment flow. This class implement the 'saveAdyenResponseData' function which saves a response corresponding the a payment to the adyen_payment_response table. Among other places, this function is called in the handlePaymentResponse function. This function was previously used to update the data in the sales_order_payment table, now it also updates or creates entries in the adyen_payment_response. The aim of this class is to create a more centralised class for processing the additionalData that is passed in every response by adyen. 

As indicated in a GH issue, the additionalData passed by Adyen should not overflow one of Magento's default tables. This is why the 'additionalInformation' and 'paymentId' column have been added to the adyen_payment_response table. Subsequently, the additionalData is not saved in the sales_order_payment table anymore. 
As a response can be saved without an order going through, a response does not necessarily have to be linked to a saved order. Thus, not always having a unique payment id assigned. As a workaround the combination of the merchant_reference and the store_id can be used a unique identifier.

Known weaknesses and strengths have been added in the comments.

**Tested scenarios**
- 3DS2
- 3DS1
- CC
- Alternative PMs

**Fixed issue**:  
#1233 
#1139


tnaber:
Known weaknesses:
- Additional data also in raw response
- Raw response and result_code overwritten on new response for payment
- No payment_id update for standard cc method
- No official foreign key on (store_id, merchant_reference)
- No unique key on payment id
- Payment id defaults to 0 (instead of null, is this a bug?)
- Did not find a place where the additionalData is actually requested in our own code, so if it is ever asked for I overlooked it :(

Morerice:
Due to a change of direction in order to avoid saving unnecessary data to the database, this PR will be closed.

Thanks,
Jean
Adyen

PR #1268, opened 2021-12-02
[PW-5915] Add address regex for special characters for version 8

AlexandrosMor:
**Description**

Polish addresses are not parsed properly because of the special characters.
Improving the regex which extract the house number from the street line if needed fixes that issue
**Tested scenarios**

- Address without special characters
- Address with special characters


peterojo (approved):

Morerice (dismissed):

Helper/Address.php
```diff
@@ -45,8 +45,8 @@ public function __construct(AdyenLogger $logger)
     }
 
     // Regex to extract the house number from the street line if needed (e.g. 'Street address 1 A' => '1 A')
-    const STREET_FIRST_REGEX = "/(?<streetName>[a-zA-Z0-9.'\- ]+)\s+(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)$/";
-    CONST NUMBER_FIRST_REGEX = "/^(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)\s+(?<streetName>[a-zA-Z0-9.'\- ]+)/";
+    const STREET_FIRST_REGEX = "/(?<streetName>[A-Za-zÀ-ÖØ-öø-ÿ0-9.'\- ]+)\s+(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)$/";
```
michaelpaul, line 48: How about if we consider these strings as unicode and match them as alphanumeric?  
```suggestion
    const STREET_FIRST_REGEX = "/(?<streetName>[[:alnum:].'\- ]+)\s+(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)$/u";
```
We can add unit tests for these cases at `Test/Unit/Helper/AddressTest.php`.

michaelpaul (changes requested):

michaelpaul (approved):

Morerice (approved):

PR #1274, opened 2021-12-03
[PW-4710] - Refactor case management (manual review) functionality

Morerice:
**Description**
This ticket will refactor the functionality which is currently used to handle case management (manual review) orders. It will also fix issues related to case management on orders when auto capture is enabled.

**Tested scenarios**

* Manual review required on auto capture / manual capture (manual review statuses set)
* Manual review required on auto capture / manual capture (manual review statuses **NOT** set)
* Manual review required on manual capture (invoice is generated **before** manual review is accepted)
* Payment w/o manual review
* Partial payment w/o manual review

**Fixed issue**: 
#1236


Model/Cron.php
```diff
@@ -2121,24 +2091,24 @@ protected function finalizeOrder($manualReviewComment = true, $createInvoice = f
...
-            // If manual review is true AND manual review status is set
-            if ($manualReviewComment && $this->requireFraudManualReview) {
-                // check if different status is selected
-                $fraudManualReviewStatus = $this->getFraudManualReviewStatus();
-                if ($fraudManualReviewStatus != "") {
-                    $status = $fraudManualReviewStatus;
-                    $comment = "Adyen Payment is in Manual Review check the Adyen platform";
+            // If a status is set, add comment, set status and update the state based on the status
+            // Else add comment and if on auto capture, set state to payment review (replicate previous functionality)
+            // TODO: Check if state should be set to PAYMENT_REVIEW
```
acampos1916, line 2096: How should we check this?

acampos1916 (approved):

Helper/CaseManagement.php
```diff
@@ -0,0 +1,184 @@
...
+
+        if (!empty($reviewRequiredStatus)) {
+            $order->addStatusHistoryComment(__($manualReviewComment), $reviewRequiredStatus);
+            $this->adyenLogger->addAdyenNotificationCronjob(sprintf(
+                'Order %s is pending manual review. The following status will be set: %s',
+                $order->getIncrementId(),
+                $reviewRequiredStatus
+            ));
+        } else {
+            $order->addStatusHistoryComment($manualReviewComment);
```
peterojo, line 121:
```suggestion
            $order->addStatusHistoryComment(__($manualReviewComment));
```

Helper/CaseManagement.php
```diff
@@ -0,0 +1,184 @@
...
+    public function markCaseAsAccepted(Order $order, $comment): Order
+    {
+        $reviewAcceptStatus = $this->configHelper->getFraudStatus(
+            Config::XML_STATUS_FRAUD_MANUAL_REVIEW_ACCEPT,
+            $order->getStoreId()
+        );
+
+        // Empty used to cater for empty string and null cases
+        if (!empty($reviewAcceptStatus)) {
+            $order->addStatusHistoryComment($comment, $reviewAcceptStatus);
```
peterojo, line 148:
```suggestion
            $order->addStatusHistoryComment(__($comment), $reviewAcceptStatus);
```

Helper/CaseManagement.php
```diff
@@ -0,0 +1,184 @@
...
+            $this->adyenLogger->addAdyenNotificationCronjob(sprintf(
+                'Created comment history for this notification linked to order %s with status update to: %s',
+                $order->getIncrementId(),
+                $reviewAcceptStatus
+            ));
+        } else {
+            $order->addStatusHistoryComment($comment);
```
peterojo, line 155:
```suggestion
            $order->addStatusHistoryComment(__($comment));
```

Model/Cron.php
```diff
@@ -319,6 +326,11 @@ class Cron
      */
     protected $invoiceHelper;
 
+    /**
+     * @var CaseManagement
+     */
+    protected $caseManagementHelper;
```
peterojo, line 332: Why not private?

peterojo (changes requested):
Just a few minor things

Model/Cron.php
```diff
@@ -2300,4 +2270,17 @@ private function getVirtualStatus($status)
...
     }
+
+    /**
+     * Call the caseManagement helper function and log the passed message
+     *
+     * @param bool $autoCapture
+     * @param string $logComment
+     * @param ...$logValues
+     */
+    private function markPendingReviewAndLog(bool $autoCapture, string $logComment, ...$logValues): void
```
michaelpaul, line 2281: What about calling these methods directly instead of adding this extra helper?

Model/Cron.php
```diff
@@ -2121,24 +2091,24 @@ protected function finalizeOrder($manualReviewComment = true, $createInvoice = f
...
+                $this->_adyenLogger->addAdyenNotificationCronjob(
+                    'Order status was changed to authorised status: ' . $status
+                );
+            } else {
+                $this->_order->addStatusHistoryComment(__($comment));
+                if ($this->isAutoCapture) {
+                    $this->_order->setState(Order::STATE_PAYMENT_REVIEW);
                 }
+                $this->_adyenLogger->addAdyenNotificationCronjob(
+                    'Order was finalized. Authorised status not set'
```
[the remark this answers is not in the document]
Morerice: https://github.com/Adyen/adyen-magento2/pull/1274/commits/74d66b9a058723e7b582048bc510e1c39e2e1e20

Model/Cron.php
Morerice: Updated in https://github.com/Adyen/adyen-magento2/pull/1274/commits/74d66b9a058723e7b582048bc510e1c39e2e1e20

Helper/CaseManagement.php
```diff
@@ -0,0 +1,184 @@
...
+        if (!array_key_exists(self::FRAUD_MANUAL_REVIEW, $additionalData)) {
+            return false;
+        }
+
+        // Strict comparison to 'true' since it will be sent as a string
+        if ($additionalData[self::FRAUD_MANUAL_REVIEW] === 'true') {
+            return true;
+        }
+
+        return false;
```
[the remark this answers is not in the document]
Morerice: I would prefer to keep it as it is for easier readability. But if you guys agree on it I can change it.

Model/Cron.php
Morerice: I just wanted to group them together since I always want to log something when the `markCaseAsPendingReview` is called. I thought it would be a bit cleaner like this but again, I can remove it if you guys want.

Helper/CaseManagement.php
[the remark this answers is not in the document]
michaelpaul: Minor style issue, either way is fine.

Model/Cron.php
michaelpaul: This API felt a bit strange to me. We are mixing business logic with logging which is something secondary. `CaseManagement` is a helper class with a logger, maybe we leave logging for `markCaseAsPendingReview`?

Morerice: That's a very good point. I included the logging in the business logic to ensure that each specific case will have some different info. I could've done this on this function (in Cron.php) but then I would have had to replicate the different scenarios (that are in the helper) in Cron.php, in order to be able to log the different messages

If Cron.php was in a good state I would have done this (or something similar) but due to it's current state I preferred to omit this for now.

peterojo (dismissed):

Helper/CaseManagement.php
```diff
@@ -0,0 +1,184 @@
...
+use Magento\Framework\App\Helper\Context;
+use Magento\Sales\Model\Order;
+
+/**
+ * Helper class for anything related to Case Management (Manual Review)
+ *
+ * Class AdyenOrderPayment
+ * @package Adyen\Payment\Helper
+ */
+class CaseManagement extends AbstractHelper
```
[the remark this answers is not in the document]
Morerice: I added it due to best practises, but as you said since none of its methods are utilized I removed it in https://github.com/Adyen/adyen-magento2/pull/1274/commits/dd34f924da2fe3cfc45587c3f0794ab05b1fa566

mirkocesaro (changes requested):
I think `_setPrePaymentAuthorized` and `_prepareInvoice` methods should be reviewed to work together properly with the configuration available on the admin.

Helper/CaseManagement.php
```diff
@@ -0,0 +1,184 @@
...
+
+        $reviewRequiredStatus = $this->configHelper->getFraudStatus(
+            Config::XML_STATUS_FRAUD_MANUAL_REVIEW,
+            $order->getStoreId()
+        );
+
+        // Set is in process to false since on Auto Capture, the payment would have already gone trough on the Adyen
+        // platform. The invoice should also be already generated.
+        if ($autoCapture) {
+            $order->setIsInProcess(false);
```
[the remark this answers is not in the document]
Morerice: Hi @mirkocesaro 

I've tested it on my end and the order is actually set to the `new` state, once the Authorization w/case management webhook is processed. If the `$order->setIsInProcess(false);` line is omitted, the order won't be set to the `new` state because of this [if condition](https://github.com/magento/magento2/blob/3d4095f9f4e691e82d002aedc5a228de1128cd1d/app/code/Magento/Sales/Model/ResourceModel/Order/Handler/State.php#L27).

I've also attached a screenshot showing two orders:
* On order **48** there are no modifications to the PR
* On order **49** I commented out the `$order->setIsInProcess(false);`

Screen Shot 2021-12-08 at 9 20 25 AM

Model/Cron.php
```diff
@@ -1527,38 +1527,27 @@ protected function _refundOrder()
...
-                $this->_setPrePaymentAuthorized();
-            } else {
-                $this->_order->addStatusHistoryComment(__(sprintf(
-                    'Partial authorisation w/amount %s %s was processed',
-                    $this->_currency,
-                    $this->_adyenHelper->originalAmount($this->_value, $this->_currency)
-                )), false);
-            }
+        if ($isTotalAmountAuthorised) {
+            $this->_setPrePaymentAuthorized();
```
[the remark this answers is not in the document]
Morerice: I added a small fix to this in this [commit](https://github.com/Adyen/adyen-magento2/pull/1274/commits/89c62c85f8d9dbd412dc299ee9028648808802c3). Although I'm aware that this if condition isn't the cleanest solution, this check will eventually be done better once we handle other PRs related to the refactoring of the `Cron.php` file, which will also include the functions you mentioned in the parent comment. This should be done in the near future.

peterojo (approved):

Helper/CaseManagement.php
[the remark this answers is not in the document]
mirkocesaro: Hi @Morerice ,
as you noticed, when the `payment_pre_authorized` status with `processing` state is configured, the result is not correct. Of course, the status of the manual review is set correctly set but the wrong state shows a wrong set of status in the update from the admin. This is exactly the problem I found with my configuration. 

And this is the reason I think _setPrePaymentAuthorized and _prepareInvoice methods should be reviewed to work together.

Model/Cron.php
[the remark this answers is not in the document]
mirkocesaro: This could work but it is not really clean, especially if in the feature PR (or other opened PR), for some reason, someone tries to save the order after state is set to 'processing'.

I will update my local env with this to make other checks.  

I'll keep you updated,
Thanks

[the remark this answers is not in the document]
Morerice: Hi @mirkocesaro 

I agree that this is not the cleanest solution, but it is something that will be updated eventually. This PR is the first attempt to refactor the code related to case management and webhook handling. I'm sure that after this first iteration, we will have more PRs to clean up more related code and appropriately handle the edge cases. If no further issues are found, we plan to merge this PR to the develop branch by the end of this week.

In the meantime, we greatly appreciate all your feedback!

Jean,
Adyen

Model/Cron.php
Morerice: We checked this by comparing the previous functioning flows and ensuring that with this PR, those flows end up in the same state. Comment removed in https://github.com/Adyen/adyen-magento2/commit/8da51746ce736d1804db34d8d426325b11d56448

michaelpaul (approved):

peterojo (approved):

PR #1286, opened 2021-12-10
FAM-1166: change order by field

Rafal-Dzierzbinski:
Our main issue is that, functionality "Refund from last payment first" is not working correctly.

config

We noticed that we have different refunds sequence, on identical orders with partial payments.

adyen_notification

As you can see, on the ps above, in order with merchant_reference 4110003215 refunds are arranged correctly, but in orders (4110002771, 4110002744) they are wrong.

adyen/module-payment/Model/ResourceModel/Order/Payment/Collection.php

This is a result of sorting data from adyen_order_payment table by create_at, because sometimes both refunds are processed at same second. This create issue, when system is fast enought to process both refunds at same second, payments are returned in wrong order (the configuration in admin is not taken into consideration is this case). Example is bellow.

adyen_order_payment

In the second query, you can clearly see, that result is not ordered correctly.


tnaber (approved):
Thanks for the contribution!

PR #1290, opened 2021-12-15
fix for capture offline issue

hankat:
**Description**
There is a possibility to create an invoice with capture offline in Magento. 
image

This capture does not trigger any request to Adyen, so for that invoice, there is no PSP reference. When that invoice is the partial one and the second invoice is created with capture online, then an unexpected error is shown in the order history comments section:
image

It is fixed by adding an additional check if the PSP reference exists before comparing its value.

**Tested scenarios**
1. Create a partial invoice with capture offline.
2. Create a second partial invoice with capture online.
3. An error is displayed in the order history comments section: "The order failed to update. Notice: Undefined index: pspReference in /app/vendor/adyen/module-payment/Helper/Invoice.php on line 211"

**Fixed issue**:

### DIFF
--- a/Helper/Invoice.php
+++ b/Helper/Invoice.php
@@ -206,7 +206,7 @@ class Invoice extends AbstractHelper
 
         foreach ($invoiceCollection as $invoice) {
             $parsedTransId = $this->adyenDataHelper->parseTransactionId($invoice->getTransactionId());
-            if ($parsedTransId['pspReference'] === $originalReference) {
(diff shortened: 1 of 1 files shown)


AlexandrosMor (approved):

peterojo (approved):

peterojo:
Hi @hankat thanks for your contribution, but this won't be necessary anymore as this function has been removed in #1295 
I will now close this PR but please don't hesitate to make another PR if you discover any related or unrelated issues

PR #1294, opened 2021-12-16
Added updated regex for number first

### LINKED ISSUE #1281: Parsing address incorrect when address contains non-latin characters
**Describe the bug**
Adyen Address Helper parses the billing/shipping address in order to get street name and house number.
See `\Adyen\Payment\Helper\Address::getStreetAndHouseNumberFromArray()`

The problem is, that regexp it uses, supports only a-zA-Z alphabet:
```
const STREET_FIRST_REGEX = "/(?<streetName>[a-zA-Z0-9.'\- ]+)\s+(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)$/";
CONST NUMBER_FIRST_REGEX = "/^(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)\s+(?<streetName>[a-zA-Z0-9.'\- ]+)/";

```
As the result, addresses that contain other letters (in our case ä, ü, ß etc. on german website) will be parsed incorrectly.
For example, address "Burgstraße 4" will give street name "e" and house number 4.

The issue is applicable to **all payment methods**.
The issue becomes high prio, when PayPal is being used on checkout, because it supports Shipping address and shopper just can't change the address, that came from the shop, i.e. from Adyen plugin.

So far, we fixed it on our side by rewriting `\Adyen\Payment\Helper\Address` and replacing regular expressions to the following:

```
const STREET_FIRST_REGEX = "/(?<streetName>[a-zA-Z0-9.'\- ]+)\s+(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)$/";CONST NUMBER_FIRST_REGEX = "/^(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)\s+(?<streetName>[a-zA-Z0-9.'\- ]+)/";
```
So, basically we accept any characters till the number met.

If you have a better idea, please share/

**To Reproduce**
1. Go to checkout
2. Enter the address, where street name contains something different than a-zA-Z (non-latin, german letters etc.)
3. Use any adyen payment method
2. Check adyen request json logs how `billingAddress` and `deliveryAddress` are being transferred to Adyen.

**Expected behavior**
`billingAddress.street` and `deliveryAddress.street` transferred correctly

**Actual behavior**
`billingAddress.street` and `deliveryAddress.street` are cut by letters than out of the range a-zA-Z

**Magento version**
2.4.2

**Plugin version**
7.3.0 (but the problem exists in all versions, even in 8.0.0)

**Additional context**
Issue can be fixed by clients by rewriting adyen helper address in local module:

`    <preference for="Adyen\Payment\Helper\Address" type="COMPANY\Adyen\Payment\Helper\Address" />`

```
namespace COMPANY\Adyen\Payment\Helper;

use Magento\Payment\Gateway\Data\AddressAdapterInterface;
use Magento\Sales\Api\Data\OrderAddressInterface;

class Address extends \Adyen\Payment\Helper\Address
{
    // Regex to extract the house number from the street line if needed (e.g. 'Street address 1 A' => '1 A')
    const STREET_FIRST_REGEX = "/(?<streetName>.+)[\s\,]+(?<houseNumber>\d{1,10}((\s)?.+)?)$/";
    CONST NUMBER_FIRST_REGEX = "/^(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)\s+(?<streetName>.+)/";
    ...
}

```

tnaber:
In #1281 it was noticed that the updated regex was not implemented for the house number first scenario, this has been added now. Note: a slightly different regex has been used as the [:alnum:] did not seem to work in combination with the ^ operator.

**Fixed issue**:  
Fixes #1281


Helper/Address.php
```diff
@@ -46,7 +46,7 @@ public function __construct(AdyenLogger $logger)
 
     // Regex to extract the house number from the street line if needed (e.g. 'Street address 1 A' => '1 A')
     const STREET_FIRST_REGEX = "/(?<streetName>[[:alnum:].'\- ]+)\s+(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)$/u";
-    const NUMBER_FIRST_REGEX = "/^(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)\s+(?<streetName>[a-zA-Z0-9.'\- ]+)/";
+    const NUMBER_FIRST_REGEX = "/^(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)\s+(?<streetName>[A-Za-zÀ-ÖØ-öø-Ÿ0-9.'\- ]+)/";
```
michaelpaul, line 49: I think alphanumeric match is possible here too, try it together with the unicode modifier (`u`):
```suggestion
    const NUMBER_FIRST_REGEX = "/^(?<houseNumber>\d{1,10}((\s)?\w{1,3})?)\s+(?<streetName>[[:alnum:].'\- ]+)/u";
```

AlexandrosMor: Indeed we did the same for `STREET_FIRST_REGEX`

michaelpaul: @tnaber What do you think about my suggestion? I think it's better than using specific character ranges which might fail for other languages.

michaelpaul (changes requested):

michaelpaul (approved):

peterojo (approved):

Morerice (approved):

PR #1297, opened 2021-12-21
[PW 6035] - Case management & partial captures support (v6)

Morerice:
**Description**
This PR will implement the changes introduced in #1295 and #1274 on the v6 branch.

**Tested scenarios**

Manual Capture

* Single capture w/card
* Single capture w/multiple payments
* Multiple captures w/single payment
* Multiple captures w/multiple payments

Automatic Capture

* Single payment
* Multiple payments


Gateway/Http/Client/TransactionCapture.php
```diff
@@ -124,4 +128,19 @@ private function placeMultipleCaptureRequests(Modification $service, $requestCon
...
+     * Copy data from the request to the response. This data will be used later when handling the response
+     *
+     * @param array $response
+     * @param array $request
+     * @return array
+     */
+    private function copyParamsToResponse(array $response, array $request): array
+    {
+        $response[self::CAPTURE_AMOUNT] = $request['modificationAmount']['value'];
+        $response[self::ORIGINAL_REFERENCE] = $request['originalReference'];
```
acampos1916, line 142: Should we check if these variable are there before assigning them?

Gateway/Request/CaptureDataBuilder.php
```diff
@@ -100,19 +108,21 @@ public function build(array $buildSubject)
...
-
         $payment = $paymentDataObject->getPayment();
         /** @var Order $order */
         $order = $payment->getOrder();
+        /** @var \Magento\Sales\Model\Order\Invoice $latestInvoice */
+        $latestInvoice = $order->getInvoiceCollection()->getLastItem();
+        $invoiceAmountCurrency = $this->chargedCurrency->getInvoiceAmountCurrency($latestInvoice);
+        $currency = $invoiceAmountCurrency->getCurrencyCode();
+        $amount = $this->adyenHelper->formatAmount($invoiceAmountCurrency->getAmount(), $currency);
+        $orderAmountCents = $this->adyenHelper->formatAmount($order->getGrandTotal(), $currency);
```
acampos1916, line 119: I think that here it is possible that `$order->getGrandTotal()` won’t match `$currency`, since the latter was fetched with the charged currency helper

Gateway/Request/CaptureDataBuilder.php
```diff
@@ -204,28 +215,47 @@ protected function getOpenInvoiceData($payment)
...
      *
      * @param $payment
      * @param $currency
      * @param $adyenOrderPayments
-     * @return array[]
+     * @param $captureAmount
+     * @return array
      */
-    private function buildMultipleCaptureData($payment, $currency, $adyenOrderPayments)
+    public function buildPartialOrMultipleCaptureData($payment, $currency, $adyenOrderPayments, $captureAmount): array
```
acampos1916, line 228: Was this made public so that we can cover it with a test?

Helper/AdyenOrderPayment.php
```diff
@@ -84,92 +89,108 @@ public function __construct(
...
+            if ($this->invoiceHelper->isFullInvoiceAmountManuallyCaptured($invoice)) {
+                $invoiceAmountCents += $this->adyenDataHelper->formatAmount(
+                    $invoice->getGrandTotal(),
+                    $invoice->getOrderCurrencyCode()
+                );
             }
         }
 
-        return $requireManualCapture;
+        $orderAmountCents = $this->adyenDataHelper->formatAmount($order->getGrandTotal(), $order->getOrderCurrencyCode());
```
acampos1916, line 177: Should we use the charged currency helper to get the order data?

Helper/AdyenOrderPayment.php
```diff
@@ -218,39 +239,22 @@ public function createAdyenOrderPayment(Order $order, Notification $notification
...
-            $orderPayment->setCaptureStatus(Payment::CAPTURE_STATUS_MANUAL_CAPTURE);
-            $this->orderPaymentResourceModel->save($orderPayment);
+        $adyenOrderPaymentsTotal = 0;
+        foreach ($adyenOrderPayments as $adyenOrderPayment) {
+            $adyenOrderPaymentsTotal += $adyenOrderPayment[OrderPaymentInterface::AMOUNT];
         }
 
-        return $orderPayment;
+        $adyenOrderPaymentsTotalCents = $this->adyenDataHelper->formatAmount($adyenOrderPaymentsTotal, $order->getOrderCurrencyCode());
+        $orderAmountCents = $this->adyenDataHelper->formatAmount($order->getGrandTotal(), $order->getOrderCurrencyCode());
```
acampos1916, line 256: Same as above

Gateway/Http/Client/TransactionCapture.php
Morerice: I can add a check just in case, but since they will always be set in the `CaptureDataBuilder` and even if they are not set, the adyen capture request will fail, which will cause this code snippet to be skipped, I'm not sure if such a check would be worth it.

Gateway/Request/CaptureDataBuilder.php
Morerice: That may be possible, but the issue with using the `ChargedCurrency` helper here is that we would need to use the `getOrderAmountCurrency` function, which in turn utilizes the `charged_currency` config value which is not on v6.

Helper/AdyenOrderPayment.php
Morerice: See https://github.com/Adyen/adyen-magento2/pull/1297#discussion_r777335221

Helper/AdyenOrderPayment.php
Morerice: See https://github.com/Adyen/adyen-magento2/pull/1297#discussion_r777335221

Gateway/Request/CaptureDataBuilder.php
Morerice: Fixed in https://github.com/Adyen/adyen-magento2/pull/1297/commits/ba4d8fd6c2aa9cc2c508a648d410aae3bc7b7bb7

michaelpaul (approved):

peterojo (approved):

michaelpaul (approved):

peterojo (approved):

PR #1303, opened 2021-12-30
[PW-1796] Add SSN check for Boleto

### LINKED ISSUE #542: [PW-1796] Validation for social security number is missing
**Magento version**: 2.3.2
**Plugin version**: 4.3.0
**Description**
1- Boleto enabled from backoffice 
2- Proceed to checkout 
3- choose Boleto as a payment method and complete an invalid Social security number (ex: 5 digits)

actual behaviour: the order is placed without errors
expected behaviour: the Social number should be validated and an error message should be thrown

The issue is available on other payment methods as well but seems to work on cash on delivery

tnaber:
Adds checker for SSN
image

Used this RegEx: https://gist.github.com/wiltonbsilva/cff6ed2970d11e8f8934

**Tested scenarios**
Adding wrong and correct SSN

**Fixed issue**:  
Fixes #542


AlexandrosMor (dismissed):

michaelpaul (changes requested):
The `validate-ssn` validator expects the North American Social Security Number format  (e.g. 123-45-6789) which is different from Brazilian format (e.g. 123.123.123-12).

I tested this only switching currency to BRL. Maybe changing another configuration would work?  

I also tested Drop-In at https://www.mystoredemo.io/ and it does validation and formatting in Brazilian format. Maybe that's something Components can do as well.

michaelpaul (approved):

acampos1916 (approved):

AlexandrosMor (approved):

PR #1309, opened 2022-01-06
[PW-5724] Refactor Cron and use webhook module

peterojo:
**Description**

- Moved Cron.php into the cron folder with the other cron jobs, and renamed it to WebhookProcessor
- Created Webhook helper to moved all notification processing logic to helper
- Refactored notification processing logic to use transition state provided by the [Adyen PHP Webhook Module](https://github.com/Adyen/php-webhook-module)

**Tested scenarios**

- [x] Authorisation: true
- [x] Authorisation: false
- [x] Authorisation: true (with manual processing)
- [x] RefundFailed
- [x] Refund
- [x] Pending
- [x] ManualReviewReject
- [x] ManualReviewAccept
- [x] Capture (triggered by invoice)
- [x] Capture (triggered from CA)
- [x] OfferClosed
- [x] Cancelled
- [x] CancelOrRefund (cancel)
- [x] CancelOrRefund (refund)
- [ ] RecurringContract

##### To be skipped
- [x] Authorised:false (less than 10 minutes old)
- [x] Offer closed (less than 10 minutes old)


Model/Cron.php
```diff
@@ -1,2229 +0,0 @@
...
-        return false;
-    }
-
-    public function execute()
-    {
-        // needed for Magento < 2.2.0 https://github.com/magento/magento2/pull/8413
-        $renderer = Phrase::getRenderer();
-        if ($renderer instanceof Placeholder) {
-            $this->_areaList->getArea(Area::AREA_CRONTAB)->load(Area::PART_TRANSLATE);
-        }
```
peterojo, line 531: Got rid of this since we don't support 2.2 anymore

Cron/WebhookProcessor.php
```diff
@@ -0,0 +1,126 @@
...
+            if ($notification->isDuplicate(
+                    $notification->getPspreference(),
+                    $notification->getEventCode(),
+                    $notification->getSuccess(),
+                    $notification->getOriginalReference(),
+                    true
+                )
+            ) {
+                $this->adyenLogger
+                    ->addAdyenNotificationCronjob("This is a duplicate notification and will be ignored");
```
Morerice, line 107: Can we add some information to the message, such as notification id?

michaelpaul:
I think we should keep our business logic in model classes instead of a "helper" class. We could create a new folder say `Model/Webhook` and break `Helper/Webhook.php` into smaller pieces inside of that folder. These classes may be more testable, so adding some basic unit tests would be a nice extra.

Helper/Webhook.php
```diff
@@ -0,0 +1,1883 @@
...
+                if ($notification->isSuccessful()) {
+                    // do not skip this notification but process it now
+                    return false;
+                }
+
+                // AUTHORISATION success=false notifications needs to be at least 10 minutes old to be processed
+                $authorisationSuccessFalseMinDate = new \DateTime('-10 minutes');
+                $createdAt = \DateTime::createFromFormat('Y-m-d H:i:s', $notification->getCreatedAt());
+
+                $minutesUntilProcessing = ($createdAt->format('U') - $authorisationSuccessFalseMinDate->format('U')) / 60;
```
acampos1916, line 396: Can we consolidate the minutesUntilProcessing calc so that we don't repeat the code for both OFFER_CLOSED and AUTHORISATION?

Helper/Webhook.php
```diff
@@ -0,0 +1,1883 @@
...
+        switch ($notification->getEventCode()) {
+            case Notification::PENDING:
+                if ($this->configHelper->getConfigData(
+                    'send_email_bank_sepa_on_pending',
+                    'adyen_abstract',
+                    $this->order->getStoreId()
+                )
+                ) {
+                    // Check if payment is banktransfer or sepa if true then send out order confirmation email
+                    if ($this->isBankTransfer($notification->getPaymentMethod()) || $notification->getPaymentMethod() == 'sepadirectdebit') {
```
acampos1916, line 488: Does this mean that SEPA confirmation emails will only be sent when the order is in PENDING and we receive a webhook with PENDING event code?

peterojo: Not **only** if, confirmation email is sent in the `authorizePayment` method as well, this case only makes sure we also send the emails for bank transfer and sepadirect even if the notification event code is PENDING.
BTW this only checks for the Notification event code not the order state, and only if the state remains unchanged from the webhook module

Helper/Webhook.php
```diff
@@ -0,0 +1,1883 @@
...
+    /**
+     * @param Notification $notification
+     * @param $previousAdyenEventCode
+     * @throws LocalizedException
+     */
+    private function cancelPayment(Notification $notification, $previousAdyenEventCode): void
+    {
+        $ignoreHasInvoice = true;
+        // if payment is API check, check if API result pspreference is the same as reference
+        if ($notification->getEventCode() == NOTIFICATION::AUTHORISATION) {
```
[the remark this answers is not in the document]
peterojo: f3950c1

Helper/Webhook.php
peterojo: f3950c1

Cron/WebhookProcessor.php
peterojo: f3950c1

michaelpaul (approved):

acampos1916 (approved):

PR #1318, opened 2022-01-14
[PW-5815]Handle declined flow  

AlexandrosMor:
**Description**

With this pr we are adding the `handleDeclineFlow` which redirects from checkout page back to the amazon pay page to choose another card. This response includes pspreference and others declined payment information. 
Testing using the amazon pay declined cards returns only an error message `AmazonPay internal error: Unknown issue has occured while processing. `  For this case the component is remount with the current return url and amazonsessionkey

**Tested scenarios**

- [x] Happy flow
- [x] Choose a declined card
- [x] 3ds Declined - wrong password
- [x] 3DS Happy flow
- [x] From Declined to happy flow
- [x] CC
- [x] klarna

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
@@ -753,10 +753,7 @@ define(
                             var self = this;
                             await self.handleOnSubmit(state.data, amazonPayComponent);
                         } catch (error) {
-                            debugger;
                             amazonPayComponent.handleDeclineFlow();
-                            //go to failed page
-                            self.handleOnSubmit(state.data, amazonPayComponent);
                         }
                     };
                     if (formattedShippingAddress &&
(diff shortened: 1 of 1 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -751,7 +747,13 @@ define(
                     configuration.productType = 'PayAndShip';
                     configuration.checkoutMode = 'ProcessOrder';
                     configuration.returnUrl = location.href;
-
+                    configuration.onSubmit = async(state, amazonPayComponent) => {
+                        try {
+                            await self.handleOnSubmit(state.data, amazonPayComponent);
```
peterojo, line 752: I think this would still work the same without the async call right?

michaelpaul:
Since `placeRedirectOrder` became async, don't we also need to update this [call](https://github.com/Adyen/adyen-magento2/blob/e83eb5775618c4b4065f971343ea1fff6fd02026/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js#L297)?

What about the multi-shipping scenario in this [case](https://github.com/Adyen/adyen-magento2/blob/e83eb5775618c4b4065f971343ea1fff6fd02026/view/frontend/web/js/view/payment/method-renderer/multishipping/adyen-hpp-method.js#L226)?

PR #1319, opened 2022-01-19
Set amount authorized in sales_order_payment for hpp payment methods

AlexandrosMor:
**Description**

`amount_authorized` field in `sales_order_payment` table was not getting populated for Adyen HPP 
but the credit cards are already initialized in magento `Payment.php` so the table is updating properly
With this PR when receiving a `AUTHORISATION` notification and the amount is full authorised then the amount is updated in the `sales_order_payment` table

**Tested scenarios**

Tested for CC and IDeal

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -1521,6 +1521,8 @@ class Cron
         if ($isFullAmountAuthorized) {
             $this->_setPrePaymentAuthorized();
             $this->_prepareInvoice();
+            //set authorized amount in sales_order_payment
+            $this->_order->getPayment()->setAmountAuthorized($this->orderAmount);
         } else {
             $this->addProcessedStatusHistoryComment();
         }


Model/Cron.php
```diff
@@ -1521,6 +1521,8 @@ protected function _authorizePayment()
         if ($isFullAmountAuthorized) {
             $this->_setPrePaymentAuthorized();
             $this->_prepareInvoice();
+            //set authorized amount in sales_order_payment
+            $this->_order->getPayment()->setAmountAuthorized($this->orderAmount);
```
Morerice, line 1525: This will only set the field once the full amount has been authorized. However when a partial payment is processed the field won't be updated. Can we update the else statement to update this field by adding the amount of this webhook to the existing amount authorized?

AlexandrosMor: I believe we need to update the amount only in case is full authorised similar to cc that is the reason why I added only after the amount is full authorised

peterojo: We're going to have a conflict with #1309 here though

AlexandrosMor: Good point @peterojo lets work on this after you merge the cron refactor

acampos1916: I think this can be picked up again, but in a different file as Cron.php is being removed

michaelpaul:
PR moved to https://github.com/Adyen/adyen-magento2/pull/1394.

Morerice:
> PR moved to #1394.

Closing this PR since it was moved to another existing one.

PR #1331, opened 2022-01-24
[PW-5815] Handle declined flow

AlexandrosMor:
**Description**

With this pr we are adding the `handleDeclineFlow` which redirects from checkout page back to the amazon pay page to choose another card. For this case the response includes the declined payment information. 
Testing using the amazon pay declined cards returns only an error message `AmazonPay internal error: Unknown issue has occurred while processing. `  For this case the component is remounted with the current return url using the url `amazonsessionkey` 

**Tested scenarios**

- [x] Amazon pay Happy flow 
- [x] Choose a declined card
- [x] Decline from Adyen with test code 
- [x] 3DS Declined - wrong password
- [x] 3DS Happy flow Amazon pay
- [x] From Declined to happy flow
- [x] CC 3DS2
- [x] CC

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
@@ -310,7 +310,7 @@ define(
 
                 return result;
             },
-            placeRedirectOrder: function(data, component) {
+            placeRedirectOrder: async function(data, component) {
                 var self = this;
 
                 // Place Order but use our own redirect url after
@@ -318,14 +318,11 @@ define(
                 $('.hpp-message').slideUp();
                 self.isPlaceOrderActionAllowed(false);
 
-                $.when(
+               await $.when(
                     placeOrderAction(data,
                         self.currentMessageContainer),
                 ).fail(
                     function(response) {
-                        if (component && component.props.methodIdentifier === 'amazonpay') {
-                            component.handleDeclineFlow();
-                        }
                         self.isPlaceOrderActionAllowed(true);
                         fullScreenLoader.stopLoader();
                         self.showErrorMessage(response);
@@ -460,7 +457,7 @@ define(
                     mount(actionNode);
                 }
             },
-            handleOnSubmit: function(state, component) {
+            handleOnSubmit: async function(state, component) {
                 if (this.validate()) {
                     var data = {};
                     data.method = this.getCode();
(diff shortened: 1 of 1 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -750,7 +747,18 @@ define(
                 if (paymentMethod.methodIdentifier.includes('amazonpay')) {
                     configuration.productType = 'PayAndShip';
                     configuration.checkoutMode = 'ProcessOrder';
-                    configuration.returnUrl = location.href;
+                    var url = new URL(location.href);
+                    if (url.searchParams.has('amazonCheckoutSessionId')) {
+                        url.searchParams.delete('amazonCheckoutSessionId');
+                    }
```
michaelpaul, line 753: We can omit the `has` check and always `delete`. If the parameter does not exist, `delete` accomplished it's job the easy way. :) 
```suggestion
                    url.searchParams.delete('amazonCheckoutSessionId');
```

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -460,7 +457,7 @@ define(
                     mount(actionNode);
                 }
             },
-            handleOnSubmit: function(state, component) {
+            handleOnSubmit: async function(state, component) {
```
michaelpaul, line 460: Is this method different from the one we use at line `125`? If it's the same, we might need an `await` there too.

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -789,25 +797,24 @@ define(
                                 currency: configuration.amount.currency,
                                 value: configuration.amount.value
                             },
-                            returnUrl: location.href,
                             showChangePaymentDetailsButton: false
                         }).mount(containerId);
                         amazonPayComponent.submit();
                         result.component = amazonPayComponent;
-                    } else {
```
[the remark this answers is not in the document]
AlexandrosMor: Yes the `else` was removed intentional. In any time it should create and mount the checkout component. The issue was that after amazon pay the component was stuck with the sessionId and it was not recreated so the button was missing for declined flow

michaelpaul (approved):

acampos1916 (approved):

AlexandrosMor:
Thank you for reviewing. We scheduled a meeting with amazon and after testing with them I am going to merge this

PR #1337, opened 2022-02-01
PW-5343 Remove Boleto as a separate payment method

AlexandrosMor:
**Description**

The boleto payment method is not part of the alternative payment methods configuration


etc/adminhtml/system/adyen_hpp.xml
```diff
@@ -71,6 +123,7 @@
                 <config_path>payment/adyen_hpp/country_code</config_path>
             </field>
         </group>
+        <!--counrty specific settings-->
```
peterojo, line 126:
```suggestion
        <!--country specific settings-->
```

AlexandrosMor: good catch 👍🏻

peterojo (approved):

michaelpaul (approved):

PR #1351, opened 2022-02-09
[PW-5070] Redo: Fix coupon counted as 'used' on canceled order in 3DS2 and rerouted 3DS1 flow

tnaber:
**Description**
Same as #1305, however this PR was reverted as it was suspected to introduce breaking changes. 

**Tested scenarios**

**Fixed issue**:


AlexandrosMor (approved):

AlexandrosMor (dismissed):

peterojo (dismissed):

Controller/Process/Result.php
```diff
@@ -515,6 +527,7 @@ protected function validatePayloadAndReturnResponse($result)
 
         try {
             $response = $service->paymentsDetails($request);
+            $this->paymentResponseHandler->handlePaymentResponse($response, $this->payment, $order);
```
michaelpaul, line 530: This controller already has a method `_validateUpdateOrder` to handle the payments response. I think that bringing the whole logic of `handlePaymentResponse` here might not be good. The alternative solution would require some code duplication but since it is a different flow, it could be necessary.

michaelpaul (approved):

acampos1916 (approved):

PR #1355, opened 2022-02-10
[PW-5070-v6] Coupon code decrement bug fix for failing 3DS flows

tnaber:
**Description**
Same as #1351, but for v6.


Controller/Process/Redirect.php
```diff
@@ -245,6 +260,32 @@ public function execute()
 
                     $this->messageManager->addErrorMessage("3D-secure validation was unsuccessful");
 
+                    // Cancel order
+                    // Always cancel the order if the payment has failed
+                    if (!$order->canCancel()) {
```
Morerice, line 265: Since this snippet is duplicated, could you abstract it a new helper class?

Morerice (approved):

Issue #1336, opened 2022-01-27
Show holder name field for card payment methods visible


dronerdk:
**Describe the bug**
If we set the "Show holder name field for card payment methods" to "No" in checkout experince configuration the name field in checkout will still be visible.

**To Reproduce**
1. set Show holder name field for card payment methods to No
2. Cache clean
3. Put product in cart and go to checkout and select credit card payment
4. Name field still visible

**Expected behavior**
The name field should not be visible if configuration is set to no

**Magento version**
2.4.3

**Plugin version**
8.1.0 + 8.1.1
![Skærmbillede 2022-01-27 kl  13 46 15](https://user-images.githubusercontent.com/15999222/151362066-2f295b0b-0542-49f1-bbcd-d3cc4aebab17.png)
![Skærmbillede 2022-01-27 kl  13 44 22](https://user-images.githubusercontent.com/15999222/151362071-0c69b7d1-4570-4d9e-a622-4d2f89f46ab6.png)

Morerice:
Hi @dronerdk,

Thanks for opening this issue and providing clear steps to replicate the problem. We will investigate this and come back to you with any updates.

Regards,
Jean
Adyen

Morerice:
Hi @dronerdk,

We have created a PR to fix this issue. It has already been tested on our setup, but feel free if you want to carry out some additional tests on your end.

Regards,
Jean
Adyen

PR #1373, opened 2022-02-18
#Show Holder Name Field Fix - When Config is set to NO/YES in Magento admin

dejankar:
**Description**
- Javascript expects bool to be returned on the frontend and to avoid type conversion we can get bool type directly from config.

- getConfigData() function variable flag is false by default, so IsSetFlag function is ignored in this case. 

- Since has_holder_name & has_holder_name required are yes/no magento fields return, $this->scopeConfig->isSetFlag()will return true/false by default if flag is set to true under getHasHolderName() & getHasHolderNameRequired().

**Tested scenarios**
 - Tested both scenarios when the Show holder name field for card payment methods is set to NO and YES. 
 - The frontend form is rendered properly without the need for type conversion in the javascript. 
 -
**Fixed issue**:  
#1336
#1345


Morerice:
Hi @dejankar,

Thanks for opening this PR. As you rightly pointed out this is a cleaner solution to my older [PR](https://github.com/Adyen/adyen-magento2/pull/1345). Hence, could you please also revert the changes of that PR and include it in this PR? Since those changes will be redundant once we merge your PR.

Also could you please remove the [binary file](https://github.com/Adyen/adyen-magento2/blob/333430b54b5661c9b695511253e37b50177857ca/.DS_Store) which I think was accidentally added?

Best regards,
Jean
Adyen

dejankar:
> Hi @dejankar,
> 
> Thanks for opening this PR. As you rightly pointed out this is a cleaner solution to my older [PR](https://github.com/Adyen/adyen-magento2/pull/1345). Hence, could you please also revert the changes of that PR and include it in this PR? Since those changes will be redundant once we merge your PR.
> 
> Also could you please remove the [binary file](https://github.com/Adyen/adyen-magento2/blob/333430b54b5661c9b695511253e37b50177857ca/.DS_Store) which I think was accidentally added?
> 
> Best regards, Jean Adyen

Hi Jean, 

I hope this is what you mean by reverting changes from your commit and adding it to my commit? 

In case I'm wrong please specify the exact commit numbers that need to be changed or reverted. 

Best Regards,
Dejan

Morerice (approved):

Morerice:
Hi @dejankar,

Yes, that was exactly what I meant. Thanks a lot for your quick response!

Best regards,
Jean
Adyen

acampos1916 (approved):

PR #1374, opened 2022-02-18
[PW-6300] Fix CC alignment issue in checkout for v8.1

peterojo:
**Description**

- Added icon to cc-form template after `getPaymentMethods()` promise resolves
- Fix alignment for POS cloud also
**Tested scenarios**
Screen Shot 2022-02-18 at 2 43 13 PM
Screen Shot 2022-02-18 at 2 43 22 PM


acampos1916 (approved):
Is this also working as expected for multishipping?

peterojo:
@acampos1916 we don't show the CC icon at all in multishipping be cause we simple use the default form template, so there was no alignment issue. Tested and multishipping works as expected.

AlexandrosMor (approved):

PR #1378, opened 2022-02-24
[PW-5275] Reactivate cart after closing a modal

### LINKED ISSUE #1346: [PW-5275] Quote cleared after cancelling 'Payconiq by Bancontact' order
**To Reproduce**
1. Add a product to your cart
2. Go to the checkout 'Payconiq by Bancontact'
3. A Payconiq popup will open:
CleanShot 2022-02-03 at 16 54 14
4. Close the Payconiq iframe

**Expected behavior**
The quote will persist and the customer can choose another payment method

**Actual behaviour**
The quote is cleared and an error will appear when choosing another payment method:
CleanShot 2022-02-03 at 16 55 59

**Magento version**
Magento 2.4.3-p1

**Plugin version**
adyen/module-payment 8.1.1

**Additional contect**
I tried enabling the 'clone quote on failed payment' setting, but this makes no difference.
CleanShot 2022-02-03 at 17 02 23

tnaber:
**Description**
When a payment is processed by opening a modal, the payment would not be properly cancelled when the modal is closed using the cross at the top right corner. This PR fixes this by adding a proper closing script in a separate file called `adyen-payment-modal`, additionally the payment is cancelled in the PaymentDetailHandler helper. This is needed as the payment will be cancelled before making the actual call to Adyen.

I have not managed to properly test this for oneclick as I don't know which oneclick method might trigger this modal.

**Tested scenarios**
- [x] cc modal
- [x] hpp modal
- [x] oneclick modal, is this modal ever used?
- [x] multishipping
- [x] admin order

**Fixed issue**:  
Closes #1346


view/frontend/web/template/payment/hpp-form.html
```diff
@@ -23,9 +23,9 @@
 -->
 
 <!-- TODO check if it can be outsources to an adyen-methods template to be used by all the methods-->
```
peterojo, line 25: @tnaber I think you can also get rid of the above TODO line since this is now different for each PM

peterojo (approved):

Issue #1377, opened 2022-02-22
[PW-6306] Klarna Payment method throws an error when the storeview currency isn't the same as the base currency.


DennisVendic:
**Describe the bug**
When placing an order with Klarna in an different currency (displayed currency) then the base currency,  an error is thrown and the user can't proceed with the order. 

**Prerequisites**
- U need to have a Magento 2 store with at least two storeviews
- The base currency of the installation is set to Euro
- The displayed currency of one of the storeviews is set to CHF
- The Adyen charged currency for both storeviews is set to Euro

**To Reproduce**
1. Navigate to the storeview set to CHF
2. Add a product to the cart
3. Go to the checkout step 1 and fill in dummy details
4. Proceed the checkout step 2 (payment methods)
5. Select Klarna payment method "Sofort bezahlen mit Klarna." 
6. Try to finalize the order

**Expected behavior**
The order should be added to Magento 2 and the client should be able to proceed with the payment.

**Actual Result**
A generic error is thrown "Error with payment method please select different payment method." and the user isn't able to proceed with the payment. 

Error added to the logfile: `[2022-02-22 10:22:59] .CRITICAL: Placing an order with quote_id SXqeA6f1qBAZDKvr6WSa7SQcjGtswX9G is failed: Error with payment method please select different payment method. [] []`

**Magento version**
2.4.3-p1

**Plugin version**
8.1.1

**Screenshots**
![CleanShot 2022-02-22 at 11 23 01](https://user-images.githubusercontent.com/100120970/155112992-93bfa21a-9f84-42cb-9dec-128f2e9e6de1.png)

**Desktop (please complete the following information):**
 - MAC OS, Google Chrome 98.0.475.80

tnaber:
Hi @DennisVendic,

Thank your for opening this ticket. We are experiencing multiple issues in this area and will be opening an internal investigation.

Cheers,
Titus

DennisVendic:
@tnaber Do you have any update for us? We still aren't able to proceed payments.

tnaber:
Hi @DennisVendic,

 #1388 should resolve this issue and is currently up for review. When this is merged it will be added in the next release.
 If you would like to use the solution earlier you can temporarily add the PR to your own codebase, copy the lines of code over or replace your version of the `Helper/ChargedCurrency.php` with the one in the PR.

Cheers,
Titus

Issue #1356, opened 2022-02-12
[PW-6306] Open invoice data incorrect with non- base currency: line amountExcludingTax is in base currency while amountIncludingTax is in FC


ThijsdejongSNC:
**Describe the bug**
In case of a foreign currency, the amountExcludingTax + taxAmount does not equal the amountIncludingTax causing klarna error "Acquirer Error (BAD_VALUE : order_lines[0].total_tax_amount)". 

**To Reproduce**
1. Create a M2 store with EUR as base currency and SEK as optional currency. Enable klarna as option for SEK / Sweden
2. Select SEK as the current currency in the front end
3. Add a random configurable product to card
4. Go to checkout and enter a Swedish address (so klarna with SEK is an option).
5. Checkout  and assert "Acquirer Error (BAD_VALUE : order_lines[0].total_tax_amount)" error or debug getOpenInvoiceData in CheckoutDataBuilder and watch the data in $formFields['lineItems'].

**Expected behavior**
In the line item data, the amountExcludingTax should be equal to amountIncludingTax minus taxAmount.

**Magento version**
2.4.3-p1

**Plugin version**
7.3.5

**Screenshots**

![image](https://user-images.githubusercontent.com/12978537/153707500-5418262c-b7c0-4682-a80e-ef256f81ae37.png)

![image](https://user-images.githubusercontent.com/12978537/153707205-ede0066a-a28f-4ef7-8f82-9ff765a33cd9.png)

**Additional context**
* tax rate for this shipping combination is set to 26%
* Additional small remark: the shipping line item calculates the vat rate back and, trough rounding, can calulate a slightly different rate with lots of decimals
![image](https://user-images.githubusercontent.com/12978537/153707408-fef19c0a-03ea-4b77-9645-5a595055c2d6.png)

ThijsdejongSNC:
I've looked into this a little deeper, In Adyen\Payment\HelperChargedCurrency:
`
        return new AdyenAmountCurrency(
            $item->getPrice(),
            $item->getQuote()->getQuoteCurrencyCode(),
            $item->getDiscountAmount(),
            $item->getPriceInclTax() - $item->getPrice(),
            null,
            $item->getPriceInclTax()
        );`

$item->getPrice() Always returns the price in base currency.

tnaber:
Hi @ThijsdejongSNC,

[...]

PR #1381, opened 2022-02-25
 Klarna Payment method throws an error when the display currency isn't the same as the base currency FIX

dejankar:
**Description**
- Klarna throws an error when the display currency is different than the base currency. 
- Function getQuoteItemAmountCurrency() under Helper/ChargedCurrency 
is returning back wrong data about item price because of the function item->getPrice(); 
- The root of the issue is because $item->getPrice() is pulled in from quote_item database entry and price is never stored in display currency but always in base currency if you have a multi-currency store. 
- so under Magento/Tax/Mode/Sales/Total/Quote/CommonTaxCollector.php -> updateItemTaxInfo() ->
$quoteItem->setPrice($baseItemTaxDetails->getPrice()) quote item price is always pulled in from the baseItemTaxDetails which holds the base price in base currency for that product. 

Solution: 
$quoteItem->setRowTotal($itemTaxDetails->getRowTotal()) is always taken from display item, but it always hold the whole item row totals. 
That's why we need to divide the $item->getRowTotal()/$item->getQty() so we can get the correct price per single quantity as needed for Klarna request. 

This is only applicable when item data is pulled from quote!!!

**Tested scenarios**
1. After fixes Klarna payment methods work. 
2. Tested an order containing taxable items and non-taxable items.
3. Tested orders with discount and tax together. 
4. Tested multiple currency scenarios as well. 

**Fixed issue**:  
#1377
#1356


tnaber:
Hi @dejankar,

Thanks a lot for this contribution! I will be merging it into a branch and open a PR on that myself. In this way I can run our CI and potentially extend your solution.

Cheers,
Titus

dejankar:
> Hi @dejankar,
> 
> Thanks a lot for this contribution! I will be merging it into a branch and open a PR on that myself. In this way I can run our CI and potentially extend your solution.
> 
> Cheers, Titus

Hi Titus, 

It's my pleasure to help! Let me know if you extend this solution and I hope we see it live soon. 

Best Regards,
Dejan

PR #1367, opened 2022-02-17
[PW-6029] - Allow for configuration of Checkout frontend region

astiskala:
**Description**

Adyen has region-specific frontend resources available in Australia and the US via the checkoutshopper-live-au.adyen.com and checkoutshopper-live-us.adyen.com domains. These can be used by setting the `environment` parameter in the Component environment configuration can be set to **live-au** or **live-us**.

Currently, Magento always uses the `live` environment. This is not only not optimal if the bulk of shoppers come from a different region, but can also lead to failures in some scenarios (e.g. Native 3DS2 flows can break when the frontend and backend regions are not in sync).

**Tested scenarios**

* Confirmed that configuration displays correct and can be set
* Confirmed that environment is set correctly in the Checkout
* No new failing unit tests

PR #1352, opened 2022-02-09
[PW-6031] add basket data to risk data builder

peterojo:
**Description**

Add [risk basket data](https://api-explorer.is.adyen.com/api-explorer/#/Payment/v68/post/capture__reqParam_additionalData-AdditionalDataRisk-riskdata----customFieldName---) and [level 2/3 data](https://api-explorer.is.adyen.com/api-explorer/#/Payment/v68/post/capture__reqParam_additionalData-AdditionalDataLevel23-enhancedSchemeData-customerReference) to /payment and /authorise requests.
Create new builder for level2/3 data
update risk data builder with basket data
Create admin config for sending additional risk data

**Fixed issues**
Closes #522

PR #1384, opened 2022-03-02
Release 8.2.0

AlexandrosMor:
**Features**
- #1331
- #1245
- #1395
- #1413
- #1429 
- #1433
- #1352
- #1367
- #1371
- #1390
- #1410
- #1414
- #1423 

**Fixes**

- #1338
- #1339 
- #1340
- #1357
- #1334
- #1365
- #1366
- #1373 
- #1374
- #1351
- #1389
- #1409
- #1397
- #1394 
- #1405
- #1436
- #1428

**Other**

- #1337
- #1349
- #1353
- #1309
- #1380
- #1391
- #1403
- #1418 
- #1419


peterojo (dismissed):

dronerdk:
Time frame for release?

Morerice:
> Time frame for release?

Hi @dronerdk,

Release should be out between this week and next week.

Regards,
Jean
Adyen

Observer/AdyenHppDataAssignObserver.php
```diff
@@ -118,12 +123,17 @@ public function execute(Observer $observer)
         if (!empty($stateData)) {
             $stateData = $this->checkoutStateDataValidator->getValidatedAdditionalData($stateData);
         }
+
+        if ($additionalData[self::BRAND_CODE] === Data::SEPA) {
```
evgv, line 127: Hi

Currently I'm working on GraphQL implementation Adyen plugin with our frontend, so I pulled develop and everything was working well, but after these update and adding these code

if ($additionalData[self::BRAND_CODE] === Data::SEPA) { $additionalDataToSave = $this->getSepaAdditionalDataToSave($stateData); }

into the AdyenHppDataAssignObserver it stopped working because after second time when this event is triggering the additional data received from payment object (table) there is no brand_code field because in the payment table no such field, this data stored in additional_information field so as the temporary solution we can add these code

if (isset($additionalData['additional_information'])) { $additionalData = array_merge($additionalData, $additionalData['additional_information']); }

right above the previous the

$additionalData = DataArrayValidator::getArrayOnlyWithApprovedKeys( $additionalData, self::$approvedAdditionalDataKeys );

to add brand_code from additional_information

this error can be reproduced within HPP method as PP for example

UPD

or we could check the cc_type field instead of brand_code because in case of HPP it stored brand code value, for example for PP cc_type = paypal

Morerice: Hi @evgv,

Thanks for your comment. Just to check that we are on the same page, you're saying that a simple solution would be to check the array before accessing the `self::BRAND_CODE` right? So something like:

```
if (array_key_exists(self::BRAND_CODE, $additionalData) && $additionalData[self::BRAND_CODE] === Data::SEPA) {
```

Thanks,
Jean
Adyen

evgv: Yep, you are right!

Morerice: Check added in https://github.com/Adyen/adyen-magento2/pull/1395/commits/102c0725267ed7f0e849078cae9769c96c46f94f

peterojo (approved):

michaelpaul (approved):

michaelpaul (dismissed):

Gateway/Validator/CheckoutResponseValidator.php
```diff
@@ -120,21 +126,23 @@ public function validate(array $validationSubject)
...
-
             if (!empty($response['error'])) {
                 $this->adyenLogger->error($response['error']);
             }
 
-            throw new \Magento\Framework\Exception\LocalizedException(__($errorMsg));
+            if (!empty($response['errorCode']) && !empty($response['error']) && in_array($response['errorCode'], self::ALLOWED_ERROR_CODES, true)) {
+                $errorMsg = __($response['error']);
+            } else {
+                $errorMsg = __('Error with payment method please select different payment method.');
```
acampos1916, line 142:
```suggestion
                $errorMsg = __('Error with payment method, please select a different payment method.');
```

michaelpaul (approved):

peterojo (approved):

PR #1389, opened 2022-03-10
[PW-6287] - Check if LPM supports recurring before adding recurring info

Morerice:
**Description**
Currently we are attempting to tokenize any alternative payment methods based on the settings related to stored card payments. The functionality should be divided and a list of payment methods that support recurring should be created. However from this list, only SEPA should be enabled for now.

Also add a new `Token Type` option to identify if the created token for alternative payment methods is a [CardOnFile](https://docs.adyen.com/online-payments/tokenization/create-and-use-tokens#save-for-one-off) or a [Subscription](https://docs.adyen.com/online-payments/tokenization/create-and-use-tokens?tab=subscriptions_2#save-for-subscriptions) token.

**Note that when upgrading the current `Store alternative payment methods` option will be set to *No*. This is done to ensure that once enabled, the admin will also choose what type of token should be created.**

**Tested scenarios**
* Card payment
* Card payment (CardOnFile token created)
* Card payment (Subscription token created)
* Tokenized card payment (CardOnFile)
* SEPA payment
* SEPA payment (CardOnFile token created)
* SEPA payment (Subscription token created)

### DIFF
--- a/Helper/PaymentMethods.php
+++ b/Helper/PaymentMethods.php
@@ -31,6 +31,7 @@ use Magento\Framework\App\Helper\AbstractHelper;
 class PaymentMethods extends AbstractHelper
 {
     const ADYEN_HPP = 'adyen_hpp';
+    const ADYEN_CC = 'adyen_cc';
 
     const METHODS_WITH_BRAND_LOGO = [
         "giftcard"
@@ -490,5 +491,63 @@ class PaymentMethods extends AbstractHelper
             'alipay_hk'
         ];
         return in_array($notificationPaymentMethod, $walletPaymentMethods);
+    }
+
+    /**
+     * Check if the payment method used was a card
+     *
+     * @param $payment
+     * @return bool
+     */
+    public function isCardPayment($payment): bool
+    {
+        return $payment->getMethod() === self::ADYEN_CC;
+    }
+
+    /**
+     * Check if the passed payment method supports recurring functionality.
+     * If a payment method is in this list it does not necessarily imply that it is already supported
+     * by the plugin
+     *
+     * @param string $lpm
+     * @return bool
+     */
+    public function paymentMethodSupportsRecurring(string $paymentMethod): bool
+    {
+        $paymentMethodRecurring = [
+            'ach',
+            'amazonpay',
+            'applepay',
+            'directdebit_GB',
+            'bcmc',
+            'dana',
+            'dankort',
+            'eps',
+            'gcash',
+            'giropay',
+            'googlepay',
+            'paywithgoogle',
+            'gopay_wallet',
+            'ideal',
+            'kakaopay',
+            'klarna',
+            'klarna_account',
+            'klarna_b2b',
+            'klarna_paynow',
+            'momo_wallet',
+            'paymaya_wallet',
+            'paypal',
+            'sepadirectdebit',
+            'trustly',
+            'twint',
+            'uatp',
+            'billdesk_upi',
+            'payu_IN_upi',
+            'vipps',
+            'yandex_money',
+            'zip'
+        ];
+
+        return in_array($paymentMethod, $paymentMethodRecurring);
     }
 }
--- a/Helper/PaymentMethods.php
+++ b/Helper/PaymentMethods.php
@@ -32,6 +32,7 @@ class PaymentMethods extends AbstractHelper
 {
     const ADYEN_HPP = 'adyen_hpp';
     const ADYEN_CC = 'adyen_cc';
+    const ADYEN_ONE_CLICK = 'adyen_oneclick';
 
     const METHODS_WITH_BRAND_LOGO = [
         "giftcard"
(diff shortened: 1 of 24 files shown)


Helper/PaymentMethods.php
```diff
@@ -487,4 +490,65 @@ public function isWalletPaymentMethod($notificationPaymentMethod): bool
...
+            'paymaya_wallet',
+            'paypal',
+            'trustly',
+            'twint',
+            'uatp',
+            'billdesk_upi',
+            'payu_IN_upi',
+            'vipps',
+            'yandex_money',
+            'zip',*/
```
acampos1916, line 548: The intention for this comment is to serve as a guide for PM codes that could be enabled for recurring in customizations, right?
If so, would it be better to move it to the function's doc block?

Model/Billing/Agreement.php
```diff
@@ -262,6 +276,46 @@ public function setCcBillingAgreement($contractDetail, $storeOneClick, $storeId)
...
+        );
+
+        $this->setAgreementLabel($label);
+        $recurringType = $this->configHelper->getAlternativePaymentMethodTokenType($storeId);
+
+        $agreementData = [
+            'bank' => [
+                'ownerName' => $savedPaymentData['ownerName'],
+                'iban' => $savedPaymentData['iban'],
+            ],
```
acampos1916, line 309: Same here. If this is so specific for SEPA I would suggest renaming the function to setSepaBillingAgreement as an alternative for the above.

Setup/UpgradeData.php
```diff
@@ -169,4 +182,42 @@ public function updateSchemaVersion800(ModuleDataSetupInterface $setup)
...
+    {
+        $configDataTable = $setup->getTable('core_config_data');
+        $pathStoreAlternativePaymentMethod = 'payment/adyen_hpp_vault/active';
+        $connection = $setup->getConnection();
+
+        $select = $connection->select()
+            ->from($configDataTable)
+            ->where(
+                'path = ?',
+                'payment/adyen_hpp_vault/active'
```
acampos1916, line 202:
```suggestion
                $pathStoreAlternativePaymentMethod
```

Model/Billing/Agreement.php
Morerice: Updated function name [ea60870](https://github.com/Adyen/adyen-magento2/pull/1389/commits/ea608700caec84cf2a23fc8e72a37dd8570f8d7b)

Model/Billing/Agreement.php
```diff
@@ -262,6 +276,46 @@ public function setCcBillingAgreement($contractDetail, $storeOneClick, $storeId)
...
+            ->setMethodCode(PaymentMethods::ADYEN_ONE_CLICK)
+            ->setReferenceId($additionalData['recurring.recurringDetailReference']);
+
+        $variant = $additionalData['paymentMethod'];
+
+        $label = __(
+            '%1, %2',
+            $savedPaymentData['ownerName'],
+            $savedPaymentData['iban']
+        );
```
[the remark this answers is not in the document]
Morerice: Updated function name [ea60870](https://github.com/Adyen/adyen-magento2/pull/1389/commits/ea608700caec84cf2a23fc8e72a37dd8570f8d7b)

Setup/UpgradeData.php
Morerice: Updated in [5bee1ec](https://github.com/Adyen/adyen-magento2/pull/1389/commits/5bee1eca1388d16f6858c89bcf668df474611f35)

Helper/PaymentMethods.php
Morerice: Updated in [8717edd](https://github.com/Adyen/adyen-magento2/pull/1389/commits/8717edd3b1595cae4346274f1449a671c3168516)

acampos1916 (dismissed):

Helper/Requests.php
```diff
@@ -329,17 +337,15 @@ public function buildBrowserData($request = [])
     }
 
     /**
+     *
```
peterojo, line 340: maybe you forgot to write some docs for this method? or why the extra lines?

Logger/Handler/AdyenWarning.php
```diff
@@ -0,0 +1,41 @@
+<?php
```
peterojo, line 1: Cool! 😎

Helper/Requests.php
Morerice: Good catch 💪  Updated in [c4ba1d3](https://github.com/Adyen/adyen-magento2/pull/1389/commits/c4ba1d3f827be53bb3aeddde5bd63f90f5ae9cbb)

acampos1916 (approved):

evgv:
Hi

Currently I'm working on GraphQL implementation Adyen plugin with our fronted, so I pulled `develop` and everything was working well, but after these update and adding these code

`if ($additionalData[self::BRAND_CODE] === Data::SEPA) {
            $additionalDataToSave = $this->getSepaAdditionalDataToSave($stateData);
        }`

into the `AdyenHppDataAssignObserver` it stopped working because after second time when this event is triggering the additional data received from payment object (table) there is no `brand_code` field because in the `payment` no such filed, this data stored in `additional_information` field so as the temporary solution we can add these code 

`if (isset($additionalData['additional_information'])) {
            $additionalData = array_merge($additionalData, $additionalData['additional_information']);
        }`
        
right above the previous condition to add `brand_code` from `additional_information`

this error can be reproduced within HPP method as PP for example

PR #1393, opened 2022-03-15
[PW-6297] Required settings - management api

AlexandrosMor:
**Description**

With [Management api](https://docs.adyen.com/api-explorer/#/ManagementService/v1/overview) the merchant can configure and manage the Adyen merchant account configuration. In this pull request we updated the magento configuration panel so by setting the `xapi key` the merchant could retrieve the 
- Associated merchant accounts
- The Client key

If the xapi key is not stored then the merchant can add it in the input field and it is used to make the management api call. 
The actual call is a `GET` to `\me`
**Tested scenarios**

- Manual testing if the call returns the correct values.
- Adding wrong xapi key
- Choosing between the merchant accounts


view/adminhtml/templates/config/required_settings.phtml
```diff
@@ -0,0 +1,98 @@
...
+        function(jQuery)
+        {
+            jQuery('#adyen_payments_configuration_errors').hide();
+            let progressSpan = jQuery('#progress');
+
+            jQuery('#adyen_configure_merchants').click(function () {
+                let xapi_key = '';
+                let demoMode = document.querySelector(
+                    "div.adyen_required_config_settings select[name*='demo_mode']").value;
+                if (demoMode == 1) {
```
michaelpaul, line 15:
```suggestion
                const $required_settings = jQuery('div.adyen_required_config_settings'); 
                let demoMode = parseInt($required_settings.find("select[name*='demo_mode']").val());
                if (demoMode === 1) {
```

michaelpaul (changes requested):
There is a small [build error](https://github.com/Adyen/adyen-magento2/runs/5558191891?check_suite_focus=true#step:7:942): 

```pre
Adyen\Payment\Controller\Adminhtml\Configuration\MerchantAccounts
		Incompatible argument type: Required type: \Magento\Backend\App\Action\Context. Actual type: \Magento\Framework\App\Action\Context; File: 
/var/www/html/vendor/adyen/module-payment/Controller/Adminhtml/Configuration/MerchantAccounts.php
```

michaelpaul (dismissed):

Helper/Data.php
```diff
@@ -1721,6 +1721,16 @@ public function createAdyenRecurringService($client)
         return new \Adyen\Service\Recurring($client);
     }
 
+    /**
+     * @param $client
+     * @return \Adyen\Service\Management
+     * @throws \Adyen\AdyenException
+     */
+    public function createManagementService($client)
```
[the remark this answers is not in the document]
AlexandrosMor: Well I thought about it but I kept it similar to rest of the services. This way I will not create confusion

michaelpaul (dismissed):

Helper/Data.php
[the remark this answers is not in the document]
peterojo: I agree with @Morerice I think we ought to be cleaning up this Data helper, which at the moment is a bit bloated, and move functions to singular responsibility classes

Model/Config/Source/MerchantAccounts.php
```diff
@@ -0,0 +1,53 @@
...
+        $this->_adyenHelper = $adyenHelper;
+    }
+
+    /**
+     * @return array
+     */
+    public function toOptionArray()
+    {
+        $merchantAccount = $this->_adyenHelper->getAdyenMerchantAccount('adyen_cc');
+        return $merchantAccount?[$merchantAccount] : null;
```
peterojo, line 51:
```suggestion
        return $merchantAccount?[$merchantAccount] : [];
```

Helper/Data.php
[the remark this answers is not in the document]
AlexandrosMor: @peterojo @Morerice ok I removed it :)

Controller/Adminhtml/Configuration/MerchantAccounts.php
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * @return \Magento\Framework\Controller\Result\Json
+     */
+    public function execute()
+    {
+        $resultJson = $this->resultJsonFactory->create();
+        try {
+            $xapikey = $this->getRequest()->getParam('xapikey', '');
+            $demoMode = $this->getRequest()->getParam('demoMode', '');
+            if (str_contains($xapikey, '*')) {
```
acampos1916, line 57: This is checking if the value from the frontend is masked/obscured no? If so I think this would ignore real API keys that happen to have a * character.

Model/Config/Adminhtml/MerchantAccounts.php
```diff
@@ -0,0 +1,52 @@
+<?php
+namespace Adyen\Payment\Model\Config\Adminhtml;
+
+class MerchantAccounts extends \Magento\Config\Block\System\Config\Form\Field
+{
+    protected $_template = 'Adyen_Payment::config/required_settings.phtml';
+
+    public function render(\Magento\Framework\Data\Form\Element\AbstractElement $element)
+    {
+        $element->unsScope()->unsCanUseWebsiteValue()->unsCanUseDefaultValue();
```
acampos1916, line 10: What is this doing?

Model/Config/Adminhtml/MerchantAccounts.php
```diff
@@ -0,0 +1,52 @@
...
+        return parent::render($element);
+    }
+
+    protected function _getElementHtml(\Magento\Framework\Data\Form\Element\AbstractElement $element)
```
acampos1916, line 14: This function seems identical to the one in the parent class, do we need to overwrite it?

Model/Config/Adminhtml/MerchantAccounts.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public function getButtonHtml()
+    {
+        $button = $this->getLayout()->createBlock(
+            'Magento\Backend\Block\Widget\Button'
+        )->setData(
+            [
+                'id' => 'adyen_configure_merchants',
+                'label' => __('Get your merchant accounts from Adyen'),
```
acampos1916, line 31:
```suggestion
                'label' => __('Get available merchant accounts from Adyen'),
```

etc/adminhtml/system/adyen_required_settings.xml
```diff
@@ -100,7 +91,24 @@
             </tooltip>
             <config_path>payment/adyen_abstract/client_key_live</config_path>
         </field>
-        <field id="live_endpoint_url_prefix" translate="label" type="text" sortOrder="90" showInDefault="1"
+        <field id="notification_username" translate="label" type="text" sortOrder="80" showInDefault="1"
+               showInWebsite="1" showInStore="1">
+            <label>Notification User Name</label>
```
acampos1916, line 96: to match docs
```suggestion
            <label>Notification Username</label>
```

acampos1916 (changes requested):

acampos1916 (changes requested):
Some comments from the previous review were not resolved

peterojo (approved):

acampos1916 (approved):

AlexandrosMor:
This pull request should remain open for now.

PR #1394, opened 2022-03-15
[PW-6238] Set authorized amount in sales_order_payment

michaelpaul:
**Description**

The `amount_authorized` field in `sales_order_payment` table was not getting populated for Adyen HPP, but the credit cards are already initialized in Magento `Payment.php` so the table is updating properly.

With this PR when receiving a `AUTHORISATION` notification and the amount is full authorised, then the amount is updated in the `sales_order_payment` table.

**Tested scenarios**

Tested for CC and IDeal.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,6 @@
 name: Main CI workflow
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   build:
--- a/Helper/Webhook.php
+++ b/Helper/Webhook.php
@@ -884,6 +884,12 @@ class Webhook
         if ($isFullAmountAuthorized) {
             $this->setPrePaymentAuthorized();
             $this->prepareInvoice($notification);
+
+            // Set authorized amount in sales_order_payment
+            $orderAmountCurrency = $this->chargedCurrency->getOrderAmountCurrency($this->order, false);
+            $orderAmount = $orderAmountCurrency->getAmount();
+            $this->order->getPayment()->setAmountAuthorized($orderAmount);
+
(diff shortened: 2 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -1,6 +1,6 @@
 name: Main CI workflow
 
-on: [push, pull_request]
+on: [pull_request]
```
michaelpaul, line 3: Running this workflow on `push` + `pull_request` causes duplicated runs when members of the repo `push` to a `pull_request`.

AlexandrosMor (dismissed):

Helper/Webhook.php
```diff
@@ -884,6 +884,12 @@ private function authorizePayment(Notification $notification)
         if ($isFullAmountAuthorized) {
             $this->setPrePaymentAuthorized();
             $this->prepareInvoice($notification);
+
+            // Set authorized amount in sales_order_payment
+            $orderAmountCurrency = $this->chargedCurrency->getOrderAmountCurrency($this->order, false);
+            $orderAmount = $orderAmountCurrency->getAmount();
+            $this->order->getPayment()->setAmountAuthorized($orderAmount);
```
Morerice, line 891: Can we move this change outside of this if condition? 
In that way, if this is a partial payment, the amount authorized will be updated on each partial payment webhook, instead of it being updated only once all the partial payments have been processed

tnaber (dismissed):

Helper/Webhook.php
michaelpaul: Sure, I moved it outside the full amount `if`.

Morerice (approved):

acampos1916 (approved):

PR #1397, opened 2022-03-16
[PW-5810] Enforce correct states when multiple states are assigned to a status

### LINKED ISSUE #1173: [PW-5810] Unpredictable order state change
**Describe the bug**

When a payment is authorised (without a capture), if the selected status for this is assigned to more than one state, the behaviour is unpredictable as either of the available states could be selected.

`    /**
     * Set State from Status
     */

    protected function _setState($status)
    {
        $statusObject = $this->_orderStatusCollection->create()
            ->addFieldToFilter('main_table.status', $status)
            ->joinStates()
            ->getFirstItem();

        $this->_order->setState($statusObject->getState());
        $this->_adyenLogger->addAdyenNotificationCronjob('State is changed to  ' . $statusObject->getState());
    }`

We currently have a status 'processing_not_captured' assigned to both STATE_PENDING_PAYMENT and STATE_PROCESSING, this is to allow for multiple order flows across different stores.

As the relationship status and state do not have a one to one relationship, it is unsafe to rely on getFirstItem() as the admin user has no way to see which state will be used.

**To Reproduce**

1. Create a new order status (e.g processing_not_captured) and assign it to the processing state
2. Configure the Adyen module so that authorised orders should use this status
3. An authorisation in a notification will now set the order state to processing and status to processing_not_captured.
4. This order now appears in the sales reporting.

but

5. Additionally assign the order status to an additional state (e.g pending_payment)
6. Now, although the status is assigned to both pending_payment and processing, an authorisation in a notification will now set the order state to pending_payment and status to processing_not_captured.

This then leads to issues, for example in our situation the sales report module ignores any orders with a state of pending_payment

It is unsafe to choose the order state based on the status because of this, as from the admin configuration perspective there is no indication this could happen

Screenshot 2021-10-11 at 11 01 16
Screenshot 2021-10-11 at 11 03 05

**Expected behavior**

Either allow the user to choose a state in the configuration, or ensure that payments are kept in a consistent state.

**Magento version**
2.4.1

**Plugin version**
6.6.6

tnaber:
**Description**
This PR addresses the issue from #1173.
When configuring statuses, a status can be assigned to multiple states. 
_E.g. the status 'Waiting for capture' can be assigned to both the 'pending_payment' and 'processing' state. When configuring "Order status: payment authorisation" it is possible to select statuses that lead to the 'processing' state, thus 'Waiting for capture' is shown in the dropdown. When this status is selected it is possible for the flow to enter the 'pending_payment' state. This should not be possible._

Further explanation of the problem and the corresponding code can be found in #1173. This PR proposes a solution in which the possible states are passed as a parameter when the `setState` function is called in the Webhook module. Before selecting a state to enter, the list of possible states will be filtered on this parameter. 
_E.g. this will prevent the 'pending_payment' state from the example to be removed._

When multiple states are possible to enter, and the passed status is also tied to all of these states _(e.g. if 'Waiting for capture' would be assigned to both 'new' and 'processing' on "Order status: payment authorisation")_. The first one that fits will be selected. This means the unpredictability is not fully removed, however this should be clear to the user as the 'Waiting for capture' label will appear multiple times in the dropdown.

**Tested scenarios**
- [x] Tested for multiple states

**Fixed issue**: 
Fixes #1173

### DIFF
--- a/Helper/Webhook.php
+++ b/Helper/Webhook.php
@@ -1643,8 +1643,6 @@ class Webhook
      */
     private function setState($status, $possibleStates)
     {
-        $this->logger->addAdyenNotificationCronjob('Looking for states: ' . json_encode($possibleStates));
-
         // Loop over possible states, select first available status that fits this state
         foreach ($possibleStates as $state) {
             $statusObject = $this->orderStatusCollection->create()
--- a/Helper/Webhook.php
+++ b/Helper/Webhook.php
@@ -1208,11 +1208,13 @@ class Webhook
             'adyen_abstract',
             $this->order->getStoreId()
         );
+        // Possible states entered by the configured status
+        $possibleStates = ['new', 'processing'];
 
         // only do this if status in configuration is set
         if (!empty($status)) {
             $this->order->setStatus($status);
-            $this->setState($status, ['new', 'processing']);
+            $this->setState($status, $possibleStates);
 
             $this->logger->addAdyenNotificationCronjob(
                 'Order status is changed to Pre-authorised status, status is ' . $status
@@ -1571,6 +1573,8 @@ class Webhook
             'adyen_abstract',
             $order->getStoreId()
         );
+        // Possible states entered by the configured status
+        $possibleStates = ['processing'];
 
         // virtual order can have different status
         if ($order->getIsVirtual()) {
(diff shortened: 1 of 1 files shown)


Helper/Webhook.php
```diff
@@ -1208,11 +1208,13 @@ private function setPrePaymentAuthorized()
             'adyen_abstract',
             $this->order->getStoreId()
         );
+        // Possible states entered by the configured status
+        $possibleStates = ['new', 'processing'];
```
acampos1916, line 1212: I would propose to move these states and the finalizeOrder ones to a constant multidimensional array in this class

acampos1916 (changes requested):

AlexandrosMor (approved):

acampos1916 (approved):

PR #1404, opened 2022-03-18
PW-6442: Remove clone quote and authentication from redirect handler

michaelpaul:
**Description**

The redirect handler endpoint cannot require authentication because in the [headless](https://docs.adyen.com/plugins/magento-2/magento-headless-integration) scenario, customers don't always use the session based login. In this case authentication is managed client side and provided via HTTP headers which won't ever be shared with Adyen servers. 

This endpoint can be accessible to anyone and should have minimal functionalities. Specially, do not perform order mutations according to our [recommendation](https://docs.adyen.com/online-payments/web-drop-in/advanced-use-cases/redirect-result):
> Because a synchronous result is not always available, for example if the shopper didn't return to your website, we strongly recommend that you only use it to present the payment result to your shopper. Use the webhooks to update your order management system.  

For this reason, the clone quote (mutation) functionally based on a provided order id must be removed. The quote restore on error will still work for non-headless stores or the ones using sessions. If a headless storefront does not use sessions, it can handle the redirect result client side via checkout component by following this [guide](https://docs.adyen.com/online-payments/web-drop-in#handle-redirect-result).   

**Tested scenarios**

* 3DS1
* iDeal

### DIFF
--- a/etc/adminhtml/system/adyen_checkout_experience.xml
+++ b/etc/adminhtml/system/adyen_checkout_experience.xml
@@ -42,12 +42,6 @@
             <config_path>payment/adyen_abstract/return_path</config_path>
             <tooltip><![CDATA[The path the customer will be redirected to when payment was <b>not</b> successful. Default is <i>checkout/cart</i>.]]></tooltip>
         </field>
-        <field id="clone_quote" translate="label" type="select" sortOrder="30" showInDefault="1" showInWebsite="1" showInStore="1" canRestore="1">
-            <label>Clone quote on failed payment</label>
-            <config_path>payment/adyen_abstract/clone_quote</config_path>
-            <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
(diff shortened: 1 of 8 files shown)


etc/adminhtml/system/adyen_checkout_experience.xml
```diff
@@ -42,12 +42,6 @@
             <config_path>payment/adyen_abstract/return_path</config_path>
             <tooltip><![CDATA[The path the customer will be redirected to when payment was <b>not</b> successful. Default is <i>checkout/cart</i>.]]></tooltip>
         </field>
-        <field id="clone_quote" translate="label" type="select" sortOrder="30" showInDefault="1" showInWebsite="1" showInStore="1" canRestore="1">
```
AlexandrosMor, line 45: Why the clone quote has to do with the shopper session?

tnaber (approved):

AlexandrosMor (approved):

dufrygrant:
I notice this was tested with 3DS1 and iDeal.
Were any checks performed on this with 3DSv2 (which is essentially the standard now)?

PR #1405, opened 2022-03-18
PW-6442: Remove authentication from redirect handler and refactor clone quote [v8]

### LINKED ISSUE #1392: [PW-6442] Restore cart does not work
Making your own contribution is easy, encouraged and greatly appreciated! We will put effort into reviewing and merging your PR quickly. For more info, please refer to the contribution guidelines: https://github.com/Adyen/adyen-magento2/blob/develop/CONTRIBUTING.md

**Describe the bug**
Call the restoreQuote() in Adyen\Payment\Model\Api\AdyenPaymentDetails always return `false` in case of refused or cancelled payment, because it uses sessions but GraphQL does't it uses the token auth instead off cookies.

**To Reproduce**
1. Go to checkout with Adyen Drop in for example, choose PP or Klarna or any HPP method
2. Click on `Place Order` and go to the payment service page, then click 'Zuruk' for Klarna or `Back` for PP and you will return to the redirect URL (you Magento checkout page)
3. Then call the `adyenPaymentDetails` GraphQL query to check the payment status
4. It will return you 500 error but it's ok, let it be, but also before the throwing the error it should restore the quote from the last session order id but there is no any last order id in session because GraphQL doesn't use it.

**Expected behavior**
After return to the Magento checkout page and run the `adyenPaymentDetails` query, quote should be restored from the last order id that you have in query parameters.

**Magento version**
2.4.3-p1

**Plugin version**
8.2.0 

**Screenshots**

 - OS: Ubuntu 20.04 LTS
 - Browser Chrome

michaelpaul:
**Description**

The redirect handler endpoint cannot require authentication because in the [headless](https://docs.adyen.com/plugins/magento-2/magento-headless-integration) scenario, customers don't always use the session based login. In this case authentication is managed client side and provided via HTTP headers which won't ever be shared with Adyen servers. 

This endpoint can be accessible to anyone and should have minimal functionalities. Specially, do not perform order mutations according to our [recommendation](https://docs.adyen.com/online-payments/web-drop-in/advanced-use-cases/redirect-result):
> Because a synchronous result is not always available, for example if the shopper didn't return to your website, we strongly recommend that you only use it to present the payment result to your shopper. Use the webhooks to update your order management system.  

For this reason, the clone quote (mutation) functionally based on a provided order id must be removed. The quote will remain active giving the possibility to make another attempt on error. Cloning also becomes unnecessary, since the same quote will be reused over all attempts.    

**Tested scenarios**

* 3DS1
* iDeal

Fixes #1392


AlexandrosMor (dismissed):

tnaber (approved):

Helper/Quote.php
```diff
@@ -130,4 +93,22 @@ public function getIsQuoteMultiShippingWithMerchantReference(string $merchantRef
 
         return $quote->getIsMultiShipping();
     }
+
+    /**
+     * Try to disable a quote after successful payment
+     * @param $quoteId
+     */
+    public function disableQuote($quoteId)
```
Morerice, line 101: In both the catch and the if statement could you either log a message, or elso throw the exception so that it is handled somewhere else instead of failing silently?

Also can you return the quote so that we can easily unit test the function?

Morerice (changes requested):

Helper/Quote.php
michaelpaul: The side effect of failing to disable a quote is that the cart will remain active. If that occurs often, bug reports can easily describe such symptom without logs. Throwing an exception will push the issue to merchants and it's a breaking change. 

This method performs a command, so ideally it should not return anything. An easy way to test is to spy the dependency interactions (e.g. with `quoteRepository` and `cartRepository`).

Morerice: Yea I agree that if we just threw the exception it will push the issue elsewhere, but if we then catch this exception wherever the `disableQuote` function is being called and log a message, the merchant should see no change from his side right?

michaelpaul: I merged `develop` into this branch and that broke GraphQL tests. Do you have any idea why @Morerice?

AlexandrosMor (approved):

Controller/Process/Result.php
```diff
@@ -246,25 +226,7 @@ public function execute()
...
-                $this->_session->restoreQuote();
-                $this->_adyenLogger->addAdyenResult(
-                    'Error when trying to create a new quote, ' .
-                    'the previous quote has been restored instead: ' . $e->getMessage()
-                );
-            }
-        } else {
-            $this->_session->restoreQuote();
-        }
+        $this->_session->restoreQuote();
```
Morerice, line 229: Can this cause any issues to headless merchants, since the session won't be available?

Observer/SubmitQuoteObserver.php
```diff
@@ -0,0 +1,23 @@
...
+ * Keep cart active based on payment method
+ */
+class SubmitQuoteObserver implements ObserverInterface
+{
+    public function execute(Observer $observer)
+    {
+        /** @var Quote $quote */
+        $quote = $observer->getEvent()->getQuote();
+        $method = $quote->getPayment()->getMethod();
+        if (in_array($method, ['adyen_hpp', 'adyen_cc'], true)) {
```
Morerice, line 19: Can we check if `adyen_oneclick` can also be included here? An example scenario would be a 3ds redirection, using a one click method.

michaelpaul: Method added!

Controller/Process/Result.php
michaelpaul: Not an issue. Using session classes still works in headless requests, clients just ignore the effects by not using cookies.

Morerice (approved):

AlexandrosMor (approved):

PR #1422, opened 2022-03-30
[PW-6473] Required settings merchant account returns array 

AlexandrosMor:
**Description**

In required settings the merchant account resets every time the merchant clicks the saving button and change any other other value in the config. That caused because the merchant account was populated with select value and not with value after calling `get /me` With this pr we return the correct array format which is 
`array('value' => '<value>', 'label' => '<label>')`

**Tested scenarios**
- CC payment
- Save values in config

### DIFF
--- a/Model/Config/Source/MerchantAccounts.php
+++ b/Model/Config/Source/MerchantAccounts.php
@@ -48,6 +48,8 @@ class MerchantAccounts implements \Magento\Framework\Data\OptionSourceInterface
     public function toOptionArray()
     {
         $merchantAccount = $this->_adyenHelper->getAdyenMerchantAccount('adyen_cc');
-        return !empty($merchantAccount) ? array($merchantAccount => $merchantAccount) : [];
+        return $merchantAccount ? array(
+            ['value' => $merchantAccount, 'label' => $merchantAccount]
+        ) : [];
     }
 }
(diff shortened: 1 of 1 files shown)


Model/Config/Source/MerchantAccounts.php
```diff
@@ -48,6 +48,6 @@ public function __construct(
     public function toOptionArray()
     {
         $merchantAccount = $this->_adyenHelper->getAdyenMerchantAccount('adyen_cc');
-        return $merchantAccount ? [$merchantAccount] : [];
+        return !empty($merchantAccount) ? array($merchantAccount => $merchantAccount) : [];
```
acampos1916, line 51: We discussed this and found that this works, but now I also see that we did it differently for the ChangedCurrency https://github.com/Adyen/adyen-magento2/blob/main/Model/Config/Source/ChargedCurrency.php

Could you check if that approach also works here? It is important that this doesn't break in all M2 versions.

acampos1916 (changes requested):

Model/Config/Source/MerchantAccounts.php
peterojo: Won't this return `['MyMerchantAccountName' => 'MyMerchantAccountName']`? I thought the idea was to return something like `['value' => 'MyMerchantAccountName', 'label' => 'Merchant Account']`

AlexandrosMor: It is modified. Check again ? @peterojo @acampos1916

peterojo (approved):

acampos1916 (approved):

Issue #1267, opened 2021-12-02
[PW-6382] GraphQL PlaceOrder dont work


larsroettig:
**Describe the bug**
* it is not possible to check out via `adyen_cc` via *GraphQLResolver* \
  `Magento\QuoteGraphQl\Model\Resolver\PlaceOrder::resolve`

**To Reproduce**
Execute Query `setPaymentMethodOnCart`

```graphql
 mutation setPaymentMethodOnCart(
        $cart_id: String!
        $payment_method: String!
        $additional_data: AdyenAdditionalDataCc
    ) {
        setPaymentMethodOnCart(
            input: {
                cart_id: $cart_id
                payment_method: {
                    code: $payment_method
                    adyen_additional_data_cc: $additional_data
                }
            }
        ) {
            cart {
                id
                selected_payment_method {
                    code
                    title
                }
            }
        }
    }

```

**Variables:**
```
{
	cart_id: 'my_cart_id',
	payment_method: 'adyen_cc',
	additional_data: {
		cc_type: 'mc',
		stateData:  JSON.stringify(onChangeData)
	}
}
```

Execute Query `placeOrder`
```graphql
    mutation placeOrder($cartId: String!) {
        placeOrder(input: { cart_id: $cartId }) @connection(key: "placeOrder") {
            order {
                order_number
            }
        }
    }
```

**Variables:**
```
{
cartId: 'my_cart_id'
}
```

**Error**
```
AdyenLoggerTest.INFO: JSON Response to Adyen:{"status":422,"errorCode":"14_006","message":"Required object 'paymentMethod' is not provided.","errorType":"validation","pspReference":"851638437550915E"} {"is_exception":false} []
``` 

I started with debugging session at `Magento\QuoteGraphQl\Model\Resolver\PlaceOrder::resolve`
he loads the payment correct and pushes it correct to Adyen logic looks like the module lost this information a plugin observer etc.

**Expected behavior**
Should be possible to place an order with `adyen_cc`

**Magento version**
2.4.3

**Plugin version**
`dev-feature/graphql-reapply`

**Screenshots**

acampos1916:
Hey @larsroettig, thanks for reaching out with this issue.

Sounds very similar to what @Bartlomiejsz reported a few weeks back. This seems related to the stateData not being added to the quote/order payment entity `additional_information` column, which should be done in the AdyenCcDataAssignObserver.php class when the `payment_method_assign_data_adyen_cc` is dispatched.

[...]

PR #1425, opened 2022-03-30
Add Plugin to handle place order with PCI compliance

larsroettig:
**Description**
[PW-6382] GraphQL PlaceOrder dont work

The recommended solution combine payment and place orders mutations, works not reliable. In Magento 2.4.3, the Magento Place Order Revolver loads the payment from the database without taking the singleton state. This behaviour triggers a reset of the Adyen\Payment\Helper\StateData this leads to a problem that an order can not place without the additional Data. 

As discussed, we can not save this information in the database regarding PCI compliance and security. Our approach is the enrich loaded payment to prevent Adyen\Payment\Helper\StateData is reset before placing an order. 

**Tested scenarios**
* Executed 
```php
mutation setPaymentMethodAndPlaceOrder(
    $cartId: String!
    $stateData: String!
) {
    setPaymentMethodOnCart(
        input: {
            cart_id: $cartId
            payment_method: {
                code: "adyen_cc"
                adyen_additional_data_cc: {
                    cc_type: "VI"
                    stateData: $stateData
                }
            }    
        }
    ) {
        cart {
            selected_payment_method {
                code
                title
            }
        }
    }

    placeOrder(
        input: {
            cart_id: $cartId
        }
    ) {
        order {
            order_id
            adyen_payment_status {
                isFinal
                resultCode
                additionalData
                action
            }
        }
    }
}
```
* Make sure that encprected card is not saved 

**Fixed issue**:  #1267

### DIFF
--- a/Plugin/GraphQlEnrichPaymentWithStateData.php
+++ b/Plugin/GraphQlEnrichPaymentWithStateData.php
@@ -1,0 +1,79 @@
+<?php
+/**
+ *                       ######
+ *                       ######
+ * ############    ####( ######  #####. ######  ############   ############
+ * #############  #####( ######  #####. ######  #############  #############
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
+ *                                      ######
+ *                               #############
+ *                               ############
+ *
+ * Adyen Payment module (https://www.adyen.com/)
+ *
+ * Copyright (c) 2022 Adyen BV (https://www.adyen.com/)
+ * See LICENSE.txt for license details.
+ *
+ * Author: Lars Roettig <l.roettig@techdivisivion.com>
+ */
+declare(strict_types=1);
+
+namespace Adyen\Payment\Plugin;
+
+use Adyen\Payment\Helper\StateData;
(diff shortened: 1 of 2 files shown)


larsroettig:
@Morerice can pls make a review of this

Plugin/GraphQlEnrichPaymentWithStateData.php
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        $additionalData = $payment->getAdditionalData() ?? [];
+
+        if (!is_array($additionalData)) {
+            return $payment;
+        }
+
+        $additionalData[AdyenCcDataAssignObserver::STATE_DATA] = $stateData;
+        $payment->setAdditionalData($additionalData);
```
michaelpaul, line 75: Hello @larsroettig, bringing state data back to the payment object could make it persistent eventually. Can we guarantee that this won't happen?

larsroettig: @michaelpaul  the `adyen_state_data` was empty but we can bassicly remove this after the event was handled by `\Adyen\Payment\Observer\AdyenCcDataAssignObserver::execute` the problem is that place order trigger this observer agian without addtional data

michaelpaul: Updating the `if` is a nice and simple solution. I think @Morerice is cooking a PR for it or maybe we do it here in your PR as well.

larsroettig:
Can be closed better soultion in terms of PCT aproval was implement in https://github.com/Adyen/adyen-magento2/pull/1429

PR #1428, opened 2022-03-31
[PW-6400] - The POST Donations API can be called multiple times

candemiralp:
**Description**

The POST Donations API could be called multiple times from the internal REST endpoint.

'donationToken' in the payment's 'additionalInformation' field is removed after each API call. Therefore, duplicate requests fail since there aren't any corresponding donationToken in the database.

**Tested scenarios**

**Fixed issue**:  PW-6400


Model/Api/AdyenDonations.php
```diff
@@ -96,6 +107,32 @@ public function donate($payload)
...
+        return $donationResult;
+    }
+
+    private function incrementTryCount($order)
+    {
+        $donationTryCount = $order->getPayment()->getAdditionalInformation('donationTryCount');
+
+        if (!$donationTryCount) {
+            $order->getPayment()->setAdditionalInformation('donationTryCount', 1);
+            $order->save();
```
acampos1916, line 124: save can be called for both IF and ELSE if moved after the closing bracket

Model/Api/AdyenDonations.php
```diff
@@ -96,6 +107,32 @@ public function donate($payload)
...
+        }
+        else {
+            $donationTryCount += 1;
+            $order->getPayment()->setAdditionalInformation('donationTryCount', $donationTryCount);
```
acampos1916, line 128: 'donationTryCount' being used multiple times should be a constant

Model/Api/AdyenDonations.php
```diff
@@ -80,10 +80,21 @@ public function donate($payload)
         /** @var Order */
         $order = $this->orderFactory->create()->load($this->checkoutSession->getLastOrderId());
 
+        $this->incrementTryCount($order);
```
peterojo, line 83: Shouldn't we be saving this `donationTryCount` only if a donation has failed?

peterojo: Seems to me if a donation has succeeded we can just delete the token end of story

candemiralp: If the donation fails, it just throws an exception in the execute method and doesn't return here. So I decided to count at the beginning to be on the safe side for the failing scenarios.

Model/Api/AdyenDonations.php
candemiralp: https://github.com/Adyen/adyen-magento2/pull/1428/commits/b6068376284c73fd88555611dc96e08375aba649

Model/Api/AdyenDonations.php
candemiralp: https://github.com/Adyen/adyen-magento2/pull/1428/commits/b6068376284c73fd88555611dc96e08375aba649

Model/Api/AdyenDonations.php
candemiralp: https://github.com/Adyen/adyen-magento2/pull/1428/commits/454cf56730d6f05442ed9a0a73d4d8095ee42285

peterojo (approved):

acampos1916 (approved):

PR #1443, opened 2022-04-08
Required settings - notifications with management API

peterojo:
**Description**

Create / update webhook credentials in CA, when they are saved in M2 admin
- Added webhook URL to notification required settings
- Added new config to save webhook_id if it doesn't exists
- Use webhook_id to update existing webhook credentials 

**Tested scenarios**

- save new webhook username, password or URL (no saved webhook_id)
- update webhook username, password or URL (when webhook_id exists)
- save other config (webhook credentials unchanged)


peterojo:
new version of php-api-library has to be released for workflow tests to pass

Helper/ManagementHelper.php
```diff
@@ -40,35 +43,86 @@ class ManagementHelper
...
+        $management = new Management($client);
+        $params = [
+            'url' => $url,
+            'username' => $username,
+            'password' => $password,
+            'communicationFormat' => 'json',
+            'active' => true,
+        ];
+        $webhookId = $this->configHelper->getWebhookId($storeId);
+        if ($webhookId) {
```
acampos1916, line 112: Should we check for !empty here?
```suggestion
        if ($webhookId) {
```

acampos1916 (approved):

Issue #1316, opened 2022-01-11
Revamp the recurring/stored payment method section


aziya-im:
The sections terminology does not align with the documentation. 
Stored the token in both Paypal BAs and Magento Vault. That would enable merchants to use Paypal BAs for subscriptions and Vault for express checkout/Instant purchase.
Don't check for these stored tokens in checkout, instead we mount the OneClick component if there are OneClick methods in the paymentMethods response.

PR #1458, opened 2022-04-13
[PW-6348] - Update recurring admin interface and recurring related payment requests (v7)

Morerice:
**Description**
This PR will update the card recurring related admin interface, to make it more clear which recurring parameters are currently being sent to the Adyen platform.

The admin interface will have the following fields:

* Enabled (Yes/No)
* Mode (Magento Vault / Adyen token)
* Token Type (CardOnFile / Subscription)
* Title
* Sort order
* Share Adyen Tokens between stores

When adding the recurringData to the request (in `Requests::buildRecurringData`) the data should be added only IF `$stateData['storePaymentMethod']` is true (this is the 'save card for my next payment' output of the checkbox on the frontend).

**Note when upgrading:**

* If tokenization is not enabled, no changes will be made
* If tokenization is enabled using Magento Vault, setup will be automatically configured
* If tokenization is enabled using Adyen Tokenization, setup will be set to use [CardOnFile](https://docs.adyen.com/api-explorer/#/CheckoutService/latest/post/payments) (One-Click) tokens.

**Tested scenarios**

* Payment to create Vault token
* Payment w/Vault Token
* Payment to create Adyen token (CardOnFile)
* Payment w/Adyen Token (CardOnFile)
* Payment to create Adyen token (Subscription)
* Payment to create Adyen SEPA token
* Payment w/Adyen SEPA token
* Upgrading with Adyen Tokenization configured
* Upgrading with Magento Vault configured

**Fixed Issue**

#1316


etc/adminhtml/menu.xml
```diff
@@ -2,6 +2,6 @@
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Backend:etc/menu.xsd">
     <menu>
         <add id="Adyen_Payment::adyen" title="Adyen" module="Adyen_Payment" sortOrder="100" parent="Magento_Backend::system" resource="Adyen_Payment::adyen"/>
-        <add id="Adyen_Payment::notifications_overview" title="Notifications Overview" module="Adyen_Payment" sortOrder="10" parent="Adyen_Payment::adyen" action="adyen/notifications/overview" resource="Adyen_Payment::notifications_overview"/>
+        <add id="Adyen_Payment::notifications_overview" title="Webhooks Overview" module="Adyen_Payment" sortOrder="10" parent="Adyen_Payment::adyen" action="adyen/notifications/overview" resource="Adyen_Payment::notifications_overview"/>
```
peterojo, line 5: Since we're making such visible changes, perhaps this should be a 7.4.0 release instead of 7.3.8?

acampos1916 (approved):

etc/adminhtml/menu.xml
Morerice: It's not a feature, just a title change so I don't think a feature release would be appropriate.

peterojo (approved):

PR #1503, opened 2022-05-10
[PW-6585] Management API: Fetch merchant accounts with /merchants endpoint

AlexandrosMor:
**Description**

For ticket PW-6297 we built the config to use the /me endpoint to fetch merchant accounts and client key fields.
However, when a API Credential has access to "All associated merchant accounts" , the 'associatedMerchantAccounts' fields in response to "GET /me" is empty. This is the reason why we would like to move to /merchants endpoint

**Tested scenarios**

Tested with pagination by decreasing the number of merchants per page 
and without pagination

### DIFF
--- a/Helper/ManagementHelper.php
+++ b/Helper/ManagementHelper.php
@@ -90,7 +90,7 @@ class ManagementHelper
         }
         return [
             'clientKey' => $responseMe['clientKey'],
-            'associatedMerchantAccounts' => $associatedMerchantAccounts,
+            'associatedMerchantAccounts' => $associatedMerchantAccounts
         ];
     }
 
--- a/Helper/ManagementHelper.php
+++ b/Helper/ManagementHelper.php
@@ -76,7 +76,7 @@ class ManagementHelper
         $responseMe = $management->me->retrieve();
         $associatedMerchantAccounts = [];
         $page = 1;
-        $pageSize = 2;
+        $pageSize = 100;
         //get the associated merchant accounts using get /merchants.
         $responseMerchants = $management->merchantAccount->list(["pageSize" => $pageSize]);
         while (count($associatedMerchantAccounts) < $responseMerchants['itemsTotal']) {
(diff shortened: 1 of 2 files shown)


Helper/ManagementHelper.php
```diff
@@ -63,19 +63,34 @@ public function __construct(StoreManager $storeManager, Data $adyenHelper, Confi
...
-        $response = $management->me->retrieve();
+        $responseMe = $management->me->retrieve();
+        $associatedMerchantAccounts = [];
+        $page = 1;
+        $pageSize = 100;
+        //get the associated merchant accounts using get /merchants.
+        $responseMerchants = $management->merchantAccount->list(["pageSize" => $pageSize]);
+        foreach ($responseMerchants['data'] as $value) {
+            array_push($associatedMerchantAccounts, $value['id']);
+        }
```
acampos1916, line 84: Is this foreach necessary if we also use the one below?

AlexandrosMor: yes it is. The `$responseMerchants['_links']['next']` is missing if there is only one page.

michaelpaul, line 84: We can also write:
```suggestion
        $associatedMerchantAccounts = array_column($responseMerchants['data'], 'id');
```

Helper/ManagementHelper.php
```diff
@@ -63,19 +63,34 @@ public function __construct(StoreManager $storeManager, Data $adyenHelper, Confi
...
+        while (isset($responseMerchants['_links']['next'])) {
+            $responseMerchants = $management->merchantAccount->list(["pageSize" => $pageSize, "pageNumber" => ++$page]);
+            foreach ($responseMerchants['data'] as $value) {
+                array_push($associatedMerchantAccounts, $value['id']);
+            }
```
michaelpaul, line 89: Then here we merge other pages:
```suggestion
            $associatedMerchantAccounts += array_column($responseMerchants['data'], 'id');
```

Helper/ManagementHelper.php
acampos1916: Right, the suggestions from @michaelpaul are what I'm referring to, I think we can do without this loop

candemiralp (approved):

Issue #1264, opened 2021-12-01
Adyen 7.3*/8.0 with Magento 2.4.4-beta*


FredericMartinez:
Hello folks!

adyen-magento 7.3.1 or 8.0 is not compatible with Magento 2.4.4-beta* on PHP 8.0

**To Reproduce**

```
composer create-project --repository-url=https://repo.magento.com/ magento/project-enterprise-edition=2.4.4-beta2 ee_244-beta2
composer require adyen-magento
php bin/magento setup:di:compile
```

Error:

```
Fatal error: Declaration of Adyen\Payment\Logger\Handler\AdyenBase::isHandling(array $record) must be compatible with Monolog\Handler\AbstractHandler::isHandling(array $record): bool in vendor/adyen/module-payment/Logger/Handler/AdyenBase.php on line 35
```

**Expected behavior**
Adyen should be compiled in `generated` without errors

**Magento version**
2.4.4-beta2

**Plugin version**
8.0
7.3.1

**CLI**
PHP 8.0

FredericMartinez:
I will create a PR, but for the record, 2 methods need to have return casted with `bool` :

* Adyen\Payment\Logger\AdyenLogger => `public function addRecord($level, $message, array $context = []): bool`
* Adyen\Payment\Logger\Handler\AdyenBase  => `public function isHandling(array $record): bool`

FredericMartinez:
After thoses fixes :arrow_up:, it will lead to another [PHP 8.0 issue](https://php.watch/versions/8.0/deprecate-required-param-after-optional):

```
Deprecated Functionality: Required parameter $storeId follows optional parameter $paymentMethodCode in /vendor/adyen/module-payment/Model/Cron.php on line 2206
```

Fix:
```
-protected function _getConfigData($field, $paymentMethodCode = 'adyen_cc', $storeId)
+protected function _getConfigData($field, $paymentMethodCode = 'adyen_cc', $storeId = null)

tnaber:
Thank you for opening this issue and leaving your solutions here! Did you still want to create a PR for this?

Morerice:
Issue was resolved in v8.2.1

aligent-lturner:
@Morerice the issue mentioned in https://github.com/Adyen/adyen-magento2/issues/1264#issuecomment-983961237 has not actually been resolved. While this is just a warning, it is still incorrect code for Magento 2.4.4

PR #1509, opened 2022-05-13
Prevent optional parameter being declared before required one

aligent-lturner:
**Description**
Version 7 is not currently compatible with Magento 2.4.4 due to deprecated functionality existing.
The `$storeId` parameter of the `\Adyen\Payment\Model\Cron->_getConfigData()` function is declared after the optional `$paymentMethodCode` parameter, but does not provide a default value. This results in errors when attempting to run Magento commands, such as `bin/magento setup:upgrade`:

```
Cache types config flushed successfully
Cache cleared successfully
File system cleanup:
/app/generated/code/Composer
/app/generated/code/Fastly
/app/generated/code/Magento
/app/generated/code/PayPal
/app/generated/code/Psr
/app/generated/code/Symfony
The directory '/app/generated/metadata/' doesn't exist - skipping cleanup
Updating modules:
Deprecated Functionality: Optional parameter $paymentMethodCode declared before required parameter $storeId is implicitly treated as a required parameter in /app/vendor/adyen/module-payment/Model/Cron.php on line 2088
ERROR: 1
```

By adding `null` as the default for the `$storeId` parameter, the issue is resolved, as it is now considered an optional parameter, even if it is always provided.

An alternative would be to change the ordering of the parameters, but this would also require all calls to the function be changed.

**Fixed issue**:  #1264 (already marked as closed, but issue not completely fixed)

### DIFF
--- a/Model/Cron.php
+++ b/Model/Cron.php
@@ -2085,7 +2085,7 @@ class Cron
      * @param $storeId
      * @return mixed
      */
-    protected function _getConfigData($field, $paymentMethodCode = 'adyen_cc', $storeId)
+    protected function _getConfigData($field, $paymentMethodCode = 'adyen_cc', $storeId = null)
     {
         $path = 'payment/' . $paymentMethodCode . '/' . $field;
         return $this->_scopeConfig->getValue($path, \Magento\Store\Model\ScopeInterface::SCOPE_STORE, $storeId);


Morerice (approved):

Morerice:
Hi @aligent-lturner,

Thanks for opening this PR!

Regards,
Jean
Adyen

tnaber (approved):

PR #1518, opened 2022-05-16
[PW-6325] - Use formattedBillingAddress for AmazonPay for the virtual products

candemiralp:
**Description**

Use `formattedBillingAddress` for AmazonPay for the virtual product `addressDetails`

**Tested scenarios**

AmazonPay

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
@@ -750,6 +750,20 @@ define(
                             phoneNumber: formattedShippingAddress.telephone
                         };
                     }
+                    else if (formattedBillingAddress &&
+                        formattedBillingAddress.telephone) {
+                        configuration.addressDetails = {
+                            name: formattedBillingAddress.firstName +
+                                ' ' +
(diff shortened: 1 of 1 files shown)


michaelpaul (approved):

tnaber (changes requested):
Wouldn't it be better to set a variable address to either formattedBillingAddress or formattedShippingAddress and then configure the addressDetails based on that variable? 
Additionally, what happens when there is no address available? (I guess it just throws an error based on the empty addressDetails in the config right? That's probably fine then)

candemiralp:
> 

@tnaber As a part of this ticket, creating a follow-up ticket related to convert all the address fields to billing address instead of shipping address has already been planned. After that follow-up, we will use one address for all of the payment method. 

For additional question, billing address is coming from user details, so it won't be empty in any case.

tnaber (approved):

AlexandrosMor:
@candemiralp make sure that you test the declined flow for amazon pay

candemiralp:
> @candemiralp make sure that you test the declined flow for amazon pay

I will test declined flow before merging this branch. Thanks @AlexandrosMor!

PR #1390, opened 2022-03-14
[PW-6021] Add 'maintain state' option on payment confirmation

tnaber:
**Description**
In #1276 it was suggested to implement a 'No status change' option on payment confirmation to allow for different flows (e.g. capturing after POD, see discussion). This was already possible in 7.2.0 by not setting any status (leaving it to '-- Please Select --'), however this feature disappeared for later versions. This PR adds the option to select 'Maintain status' in the `Order status: payment confirmed` dropdown. The '-- Please Select --' option has been removed to prevent confusion to which status will be selected, the default is 'Processing'.

The state will be maintained only in the Manual Capture flow.

image
 
**Tested scenarios**
- [x] Manual capture
- [x] Immediate capture
- [x] Verified by community

**Fixed issue**: 
Closes #1276

Issue #1276, opened 2021-12-03
[PW-6021] Forced order status change for Capture notifications


sheppeybrian:
**Describe the bug**
Force change of order status on capture notifications

Due to the capture of payment being one of the last functions in our order flow (we don't charge customer card until after despatch)
we require the order status to remain in its last state/status and not be forced back into a "processing" state / status.

This was possible in 7.2.0 by not allocating a dropdown option in admin setting 
"Order status: payment confirmed" --Please Select--

Since update if a option is not selected it results in a order status update to the default "closed" state in magento

Can you allow for a option of "No Status Change" 

I don't really see the requirement to restrict the order status options to only a Processing state either.  
A manual capture request could be requested during a On Hold or Complete state.

**To Reproduce**
Updated from 7.2.0 - 7.3.2
Suspect this may be a issue related to changes made to resolve [PW-5812]

**Expected behavior**
No change of order status if one is not selected in admin dropdown options

**Magento version**
2.4.3

**Plugin version**
7.3.2

tnaber:
Hi @sheppeybrian,

Thank you for opening this issue and sorry for the slow response. So you would like to have the option of there being no status change when the manual capture webhook notification is handled? Especially since such a status change will cause the order to go back to the processing state which will ultimately interfere with your currently configured workflow. 
Can you verify that we correctly understand your problem?

Cheers,
Titus

sheppeybrian:
Yes this is correct
Reverted back to 7.2.0 and the magento order status stays unchanged on capture as we have not allocated it one in modules configuration.
The new modules results in a closed order state if one is not selected.

A closed or processing state really does not work with our our order flow of not capturing payments until the order ships.

tnaber:
Hi @sheppeybrian, 

We have been looking into this issue on v7.3.2. We could not replicate the behaviour that a flow enters the 'closed' state after processing the capture notification. 

Our states (not custom statuses) had the following flow for these steps: **Order created -> authorisation webhook -> invoice created -> capture notificaiton processed**
- Not setting the "Order status: payment confirmed" results in: _payment_review -> processing -> payment_review -> payment_review_
- Setting "Order status: payment confirmed" to processing results in: _payment_review -> processing -> payment_review -> processing_

As we understand it, you want to have the following flow: _payment_review -> processing -> payment_review -> payment_review_. With the option of having no state change after the capture notification. 

Could you provide me with:
[...]

PR #1533, opened 2022-05-23
[PW-6544] Replaced 'Processing' with 'Pending Payment' in invoicing flow

tnaber:
**Description**
As pointed out in the discussion in https://github.com/Adyen/adyen-magento2/pull/1390. There was the possibility of an order being closed where it shouldn't when the order was shipped before the invoice was captured.
Since 8.1.0 an invoice will always set and orders' status to Processing to allow for quick multiple captures (without having to wait for the first one to be captured). If the order would be shipped, Magento's flow sees the closed state as the only option due to https://github.com/magento/magento2/blob/b092dd6df013e8b1e77e8b5d639148ed76a7172a/app/code/Magento/Sales/Model/ResourceModel/Order/Handler/State.php#L33
By introducing the 'Pending Payment' state, the order will be set to Pending Payment after it has been invoiced. This allows for a quick second capture (in case of partial payment) and allows for Shipping the order (it will stay in Pending Payment). It will move to Processing if the full amount has been captured. 

**Tested scenarios**
- [x] Partial Payment
- [x] Normal Payment
- [x] Shipping before capture
- [x] Shipping after capture  

**Additional notes**
Most likely this will make the 'maintain state' option redundant (from #1390), we will leave this in for now. But might remove this later.

**Fixed issue**:  
 #1390
 #1276

### DIFF
--- a/Helper/Webhook.php
+++ b/Helper/Webhook.php
@@ -72,7 +72,7 @@ class Webhook
      * Indicative matrix for possible states to enter after given event
      */
     const STATE_TRANSITION_MATRIX = [
-        'payment_pre_authorized' => [Order::STATE_NEW, Order::STATE_PROCESSING],
+        'payment_pre_authorized' => [Order::STATE_NEW, Order::STATE_PENDING_PAYMENT],
         'payment_authorized' => [Order::STATE_PROCESSING]
     ];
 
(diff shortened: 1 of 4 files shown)


peterojo (approved):

candemiralp (approved):

igi8819:
@tnaber thanks, your changes broke our project order status flow......never see that buggy module as adyen magento 2

PR #1537, opened 2022-05-25
PW-6713: Build on multiple environments

michaelpaul:
**Description**

Run builds on a custom Docker image supporting multiple versions of PHP and Magento. The E2E tests will only run on a release PR because they take too much time. 

**Tested scenarios**

* No business logic changed


peterojo (dismissed):

.github/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+DB_PREFIX=m2_ \
+ELASTICSEARCH_SERVER="<will be defined>" \
+ELASTICSEARCH_PORT=9200 \
+ELASTICSEARCH_INDEX_PREFIX=magento2 \
+ELASTICSEARCH_TIMEOUT=15 \
+ADMIN_NAME=admin \
+ADMIN_LASTNAME=admin \
+ADMIN_EMAIL=admin@example.com \
+ADMIN_USERNAME=admin \
+ADMIN_PASSWORD=admin123 \
```
acampos1916, line 20: Should we setup these instances with stronger credentials and move them to Github secrets? They are ephemeral but good practice to do so

michaelpaul: These are default values we must set when declaring environment variables. They are actually overwritten by secrets at runtime: https://github.com/Adyen/adyen-magento2/blob/9890f0fa1060e288ece13d0097e43106ca85df66/.github/workflows/integration-test.yml#L18.

Morerice (dismissed):

acampos1916 (approved):

candemiralp (dismissed):

peterojo (approved):

PR #1541, opened 2022-05-27
[PW-6682]The toggle switch example for admin panel

AlexandrosMor:
**Description**

Example implementation of toggle button based on this [tutorial](https://getridbug.com/magento/magento2-toggle-button-yes-no-to-enable-disable-module-function-in-shipment-method/)


AlexandrosMor:
Close this one as it is an example code

PR #1550, opened 2022-05-31
[PW-6425] Enable auto-capture for Afterpay transactions

RokPopov:
**Description**
Enable auto capture for AfterPay transactions.

**Tested scenarios**
Complete the AfterPay transaction with auto-capture for invoice payment methods enabled, checked whether the transaction was successful and the order updated upon the processing of the notifications by cronjob.

Complete manual capture AfterPay transaction with auto-capture for invoice payment method disabled, captured the payment manually within Magento with invoicing it, payment was successfully settled.


Helper/Webhook.php
```diff
@@ -1348,7 +1348,7 @@ private function isAutoCapture($notificationPaymentMethod): bool
 
             // if auto capture mode for openinvoice is turned on then use auto capture
             if ($captureModeOpenInvoice &&
-                $this->adyenHelper->isPaymentMethodOpenInvoiceMethodValidForAutoCapture($notificationPaymentMethod)
+                $this->adyenHelper->isPaymentMethodOpenInvoiceMethod($notificationPaymentMethod)
```
candemiralp, line 1351: @RokPopov 
Does a payment method always support `AutoCapture` if it is an `OpenInvoice` method?

RokPopov: Double-checking this, as of now this is the case - we support `AutoCapture` for all `OpenInvoice` payment methods

candemiralp (approved):

PR #1553, opened 2022-06-02
[PW-6743] - Hide ApplePay on non Safari browsers

candemiralp:
**Description**
ApplePay was shown on Chrome however this payment method can be used with Safari only from browsers.

Related constructor updated.

**Tested scenarios**
- Check Google Chrome
- Check Safari


tnaber (approved):
;)

Morerice (approved):

PR #1557, opened 2022-06-02
[PW-6726] Enable REFUND notifications to be processed

### LINKED ISSUE #1493: [PW-6726] Refund notification stuck to processing
**Describe the bug**
Refund notification stuck to in processing

**To Reproduce**
1. Place an order
2. Ship
3. Refund
4. When Adyen Refund notification cron will run, credit memo will be created, order status updated but notification stay to processing at 1 and done to 0

**Expected behavior**
Adyen notification should end correctly and set to done = 1 and processing = 0 in adyen_notification table..

**Magento version**
2.4 3-p1

**Plugin version**
8.2.0 and 8.2.2

RokPopov:
**Description**
There was an issue with processing the REFUND notification when a credit memo (refund) was created. Magento sets the order state and status to cancelled. When we then attempt to process the webhook, there was no mapping for the order state in WEBHOOK_ORDER_STATE_MAPPING array. This solution adds the STATUS_CLOSED to the mentioned array and adds the `Notification::REFUND` case to the `handleUnchangedStates` function.

**Tested scenarios**
1. Place the order, process the AUTHORISATION notification related to it and ship it
2. Refund the order and process the REFUND notification
3. `Total_Refunded` column in `adyen_order_payment` updates with the correct value, as the REFUND notification is processed correctly

**Fixed issue**:  
Closes #1493


candemiralp (approved):

AlexandrosMor (approved):

Helper/Webhook.php
```diff
@@ -440,6 +441,8 @@ private function handleOrderTransition(Notification $notification, $transitionSt
     private function handleUnchangedStates(Order $order, Notification $notification): void
     {
         switch ($notification->getEventCode()) {
+            case Notification::REFUND:
+                $this->refundPayment($notification);
```
peterojo, line 445: `refundPayment()` is already called in `handleOrderTransition()`. If `handleUnchangedStates()` is called it means that the webhook module has determined that the order is already in the desired state

peterojo: Probably adding closed state to the mapping should be enough to fix the issue

peterojo (changes requested):
Could you test with removing the changes in `handleUnchangedStates`

peterojo (approved):

PR #1558, opened 2022-06-07
Changed addError/addInfo method usage with error/info [Magento 2.4.4 issue]

mtytula:
**Description**
Changed usage of `addError` / `addInfo` method to `error` / `info`. I've installed this module on Magento 2.4.4 that has requirement `"monolog/monolog": "^2.3"` and this version does not have those methods.

This record is showing up during placeOrder request
`[2022-06-07T05:30:30.176911+00:00] report.ERROR: Call to undefined method Adyen\Payment\Logger\AdyenLogger::addInfo()
`

**Tested scenarios**

**Fixed issue**:


Morerice (approved):

Morerice:
Hi @mtytula,

Thanks for opening this PR. We support Magento [v2.3.7+](https://github.com/magento/magento2/blob/2.3/composer.json), which requires minimum monolog version [1.17](https://github.com/Seldaek/monolog/blob/1.17.0/src/Monolog/Logger.php), which in turn contains both `addError` and `error` functions.

Hence, this change should not pose any issues since we can just directly use the `error` function on all the Magento version we support.

Thanks,
Jean
Adyen

mtytula:
> Hi @mtytula,
> 
> Thanks for opening this PR. We support Magento [v2.3.7+](https://github.com/magento/magento2/blob/2.3/composer.json), which requires minimum monolog version [1.17](https://github.com/Seldaek/monolog/blob/1.17.0/src/Monolog/Logger.php), which in turn contains both `addError` and `error` functions.
> 
> Hence, this change should not pose any issues since we can just directly use the `error` function on all the Magento version we support.
> 
> Thanks, Jean Adyen

Those methods are also available within Magento 2.3.3 version. So I think those changes should not cause any new bugs

acampos1916 (approved):

PR #1569, opened 2022-06-10
[PW-6783] - Implement handleReject() for HPP methods.

candemiralp:
**Description**

If a user is blocked from the Customer Area due to fraud risk, the payment is getting refused as expected. But, PayPal pop-up window is staying active with an empty spinner. In order to close this window `handleReject()` method was implemented from `component`.

**Tested scenarios**

- PayPal
- Ideal
- Generic Gift Card

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
@@ -313,6 +313,7 @@ define(
                         self.isPlaceOrderActionAllowed(true);
                         fullScreenLoader.stopLoader();
                         self.showErrorMessage(response);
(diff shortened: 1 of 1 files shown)


peterojo (approved):

acampos1916 (approved):

michaelpaul (approved):

kweij:
On our client's webshop, sometimes a successful MobilePay payment fails to update/continue the checkout. After some investigation we found that it coincides with the following JS error:

`Error: component.handleReject is not a function. (In 'component.handleReject(response)', 'component.handleReject' is undefined)`

This relates to the change that is made by this pull-request. It appears that it cannot be assumed that the `handleReject()` method is present on the `component`, so I propose this needs a check first.

PR #1580, opened 2022-06-22
[PW-6821] - Add check to InvoiceObserver to verify if the order was made with Adyen payment method

RokPopov:
**Description**
When an order is placed by any payment method, the invoice observer listens doesn't check to see if the Adyen payment method was selected. This PR is adding a check to identify whether the payment method is Adyen payment method.

**Tested scenarios**
- created an order with manual capture with Adyen payment method to make sure the logger creates an entry in debug.log
- created an order with non-Adyen payment method to see whether it breaks out of the execute() function of the InvoiceObserver and doesn't write in debug.log

### DIFF
--- a/Helper/PaymentMethods.php
+++ b/Helper/PaymentMethods.php
@@ -156,14 +156,23 @@ class PaymentMethods extends AbstractHelper
     }
 
     /**
+     * @return array
+     */
+    public function getAdyenPaymentCodes(): array
+    {
+        return [self::ADYEN_HPP, self::ADYEN_CC, self::ADYEN_ONE_CLICK];
+    }
+
+    /**
      * @param string $methodCode
      * @return bool
      */
     public function isAdyenPayment(string $methodCode): bool
     {
-        if($methodCode !== self::ADYEN_HPP || $methodCode !== self::ADYEN_CC || $methodCode !== self::ADYEN_ONE_CLICK) {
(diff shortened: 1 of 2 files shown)


candemiralp (approved):

dronerdk:
Thanks for creating this bug, we have the same issue with banktranfer and and it gets the status changed from Adyen plugin,
When is the new release scheduled for?

Morerice:
Hi @dronerdk,

Release should be out within this week.

Regards,
Jean
Adyen

dronerdk:
Hello

We still face the problem that non-Adyen payment methods will change order status as is was an Adyen payment.
We see this problem in the native "banktransfer" method. 
When we create an order with banktransfer the order status will be set as the Adyen configuration and not what we expected it should be (pending payment)
One think to keep in mind. We made 1 change in the banktransfer.php model. We added   protected $_canCapture = true; so we can keep invoices as pending unit the customer pays the payment by manual bank transfer.
So when we create the banktranfer invoice we do not capture the payment so the invoices will keep pending state and it is in this scenario that the order status will kep the Adyen configuration "approved". It shoud be "pending payment". When we "capture" then banktransfer the status will correctly be processing.
Skærmbillede 2022-07-04 kl  10 04 25
Skærmbillede 2022-07-04 kl  10 02 18





Skærmbillede 2022-07-04 kl  10 06 05

PR #1587, opened 2022-06-28
[PW-6724] Separator line between Credit Card and the next payment method

RokPopov:
**Description**
Add the border line between Credit Card and the next payment method in the checkout page.

**Tested scenarios**
Proceed to the checkout page and observer the separator line between the Credit Card payment method and the next payment method available in the list of payment methods.

### DIFF
--- a/view/frontend/web/css/styles.css
+++ b/view/frontend/web/css/styles.css
@@ -11,3 +11,7 @@
 .adyen-checkout__dropdown__button {
     width: auto;
 }
+
+.payment-method:nth-of-type(3) {
+    border-bottom: 1px solid #ccc;
+}


candemiralp (approved):

michaelpaul (approved):

MrGekko:
This commit causes a extra line in the wrong place in my checkout. 
It might be solved by doing this instead:
```css
.payment-method+#hpp_actionModalWrapper .payment-method-title {
    border-bottom: 1px solid #ccc;
}
```
Thoughts?

RokPopov:
Hi @MrGekko,

Thank you for pointing this out to us.

Trying to implement your proposed solution resulted in the separator line not showing in the checkout page bellow the cards, in both cases, when the stored cards are enabled/disabled:
Screen Shot 2022-07-26 at 10 04 27
Screen Shot 20
22-07-26 at 10 03 52

However, only using the `#hpp_actionModalWrapper` for selecting the element always bellow the cards, the separator line seems to be always showing:
Screen Shot 2022-07-26 at 10 06 32
Screen Shot 2022-07-26 at 10 06 52

Your solution doesn't seem to work for me, do I need to adjust something else?

Thank you in advance,

Kind regards,
Rok,
 Adyen

MrGekko:
Hi Rok,

Thanks for your research into the issue. I should have known my sollution did not suit every (theme) implementation..
Do you have a sollution? Adyen can be used side-by-side with other payment sollutions. so selecting the third payment-method and sticking a border to it is not the sollution either. I believe [this ](https://github.com/magento/magento2/blob/7c6b6365a3c099509d6f6e6c306cb1821910aab0/app/design/frontend/Magento/luma/Magento_Checkout/web/css/source/module/checkout/_payments.less#L44) is where Luma adds the border.

I did not test this, but might reversing the border work?:
```
.payment-method+#hpp_actionModalWrapper .payment-method-title {
    border-top: 1px solid #ccc;
}
```
This way it will comply with the `.payment-method+.payment-method .payment-method-title {` of luma.

RokPopov:
Hi @MrGekko,

Thank you for your suggestion!
I tested your proposal and it doesn't seem to be working. The class `.payment-method-title` doesn't seem to be found:

Screen Shot 2022-08-04 at 10 17 48

However, if I remove the above mentioned class and only target `.payment-method+#hpp_actionModalWrapper`, we do find the first element bellow the card component. Testing it, it also seems to be working in both cases, where the shopper has already stored payment method and where the shopper doesn't have any payment details stored:

Screen Shot 2022-08-04 at 10 21 51

Screen Shot 2022-08-04 at 10 22 29

Screen Shot 2022-08-04 at 10 24 41

What would be your thoughts on this?

Thank you in advance,

Kind regards,
Rok,
Adyen

MrGekko:
Hi, thanks, that looks nice!

RokPopov:
Hey @MrGekko,

Thank you for looking into this with us.
I have created a [PR](https://github.com/Adyen/adyen-magento2/pull/1653) with the fix that will be included in the next release.

Kind regards,
Rok,
Adyen

PR #1589, opened 2022-06-29
Fixed getting cc type name

mtytula:
**Description**
I've found a bug during testing payments with credit cards. It keeps showing "Unknown" no matter what kind of test credit card was used.
image

I did some debug and here is what I found. There is a method called `getAdyenCcTypes()` and it loads all cc types from adyen_pament.xml. `\Adyen\Payment\Block\Info\Cc::getCcTypeName` method uses this method and looks after valid ccType in    `$types` variable. This variable is an array with ccType keys, but in uppercase. So I think the right solution was to add change the line `\Adyen\Payment\Block\Info\Cc.php:30`

Here is the result.
image

**Tested scenarios**
1. Create an order with credit card.
2. Check your order confirmation email or check your order on admin order page.


mtytula:
I've changed my approach because not all of cc types was passing upper-cassing cc type code (e.g. visa. Array key is "VI" and logic with my previous change was still returning "Unkown" value)

Morerice:
Hi @mtytula

Apologies for the delay. As you can see from the attached picture I was un
able to replicate this issue. Could you please add the version you're on so that maybe I can attempt to replicate on the same version?

Thanks,
Jean
Adyen

Screen Shot 2022-07-25 at 15 01 00

mtytula:
> Hi @mtytula
> 
> Apologies for the delay. As you can see from the attached picture I was un able to replicate this issue. Could you please add the version you're on so that maybe I can attempt to replicate on the same version?
> 
> Thanks, Jean Adyen
> 
> Screen Shot 2022-07-25 at 15 01 00

I'm using the 8.3.1 version

mtytula:
@Morerice , did you checked this? :)

Morerice:
> @Morerice , did you checked this? :)

Hi @mtytula,

Sorry for the delay. I tried to reproduce this on the versions you mentioned, but I was still unable to do so. I haven't had time yet to look more deep in the code (to check the output of `$this->getInfo()->getCcType()` on my system).

If this was indeed an issue on these versions, I would expect quite a large number of users to reach out to us about it, since denoting the correct card type is vital for the users. However till now, we haven't had any other users who contacted us about this.

I should be able to check the output of the function this week, so that I can get back to you with more info.

Regards,
Jean
Adyen

Morerice:
Hi @mtytula,

I've checked this and on my end `$this->getInfo()->getCcType()` was returing `VI` as expected. Could you check what is saved in the `sales_order_payment.cc_type` field, in your database?

Regards,
Jean
Adyen

mtytula:
Hi @Morerice ,

image

Morerice:
> Hi @Morerice ,
> 
> image

That is definitely what is causing the issue then. On my setup, that field has `VI`. Could you maybe debug on your end what is being saved when `setCcType` is being called in our module? I would recommend checking the calls in `CheckoutResponseValidator` and `AdyenCcDataAssignObserver`.

Regards,
Jean
Adyen

mtytula:
@Morerice 
[2022-08-04T05:52:55.799152+00:00] report.ERROR: AdyenCcDataAssignObserver [] []
[2022-08-04T05:52:55.799564+00:00] report.ERROR: paymentInfo->setCcType(additionalData[self::CC_TYPE]); >>>> cc_type =  visa [] []

Morerice:
Hi @mtytula,

So just to be sure we're on the same page you're getting no `additionalData` parameter in the response in the CheckoutResponseValidator right?

Regards,
Jean
Adyen

mtytula:
@Morerice , you're correct.

Morerice:
> @Morerice , you're correct.

Got it. In that case this doesn't seem to be an issue with the magento module, but rather an issue with some configuration on Adyen side which may be causing the `additionalData` **not** to be received. Could you please create an internal Adyen ticket by sending an email to magento@adyen.com? 

This would also be beneficial so that you can share pspReferences with us to be able to identify the problem. Please include a link to this github issue so that our tech support team also reaches out to me if required.

I will close this PR for now. However if we identify that the issue is related to the module, I will re-open it and add further info.

Best regards,
Jean
Adyen

PR #1602, opened 2022-07-06
[PW-6887] - Add debug logging to webhook rate limiter, use only IP address for cache key

RokPopov:
**Description**
Refactor the creation of cache key with using only the IP address and dropping the session ID. The reason behind this is that it is easy to use a new session ID with each attempt, therefore it doesn't make sense to use it as the cache key.

For the visibility of potential fraudulent attempts of requests on the merchant's notification endpoint, add logging in the debug.log file to inform of the potential attacks.

**Tested scenarios**
- send notification to endpoint with correct username pwd, validate if request is accepted and not blocked after 6 tries
- send notification to endpoint with wrong username pwd and see if request is not rejected immediately
- send notification to endpoint with incorrect username pwd, validate if request is rejected while blocked after 6 tries
- send notification to endpoint with wrong username pwd while logging the cache key to verify the expiration date increases exponentially


michaelpaul (dismissed):

candemiralp (approved):

Helper/RateLimiter.php
```diff
@@ -52,38 +52,49 @@ class RateLimiter
...
+        return "adyen-logins-" . $this->remoteAddress->getRemoteAddress();
     }
 
     public function saveSessionIdIpAddressToCache()
     {
         $cacheValue = $this->getNumberOfAttempts();
 
+        if($cacheValue === self::NUMBER_OF_ATTEMPTS) {
+            $this->adyenLogger->addAdyenDebug(
+                "Notification has been rejected because the allowed number of attempts is exceeded."
```
acampos1916, line 94: Can we adjust the message and add the IP address by using sprintf()?
```suggestion
                "Webhook has been rejected because the allowed number of authentication attempts has been exceeded."
```

RokPopov: 🙏

AlexandrosMor (approved):

PR #1618, opened 2022-07-18
[PW-6705] - Create a separate payment method for MOTO

candemiralp:
**Description**

Creating a different payment method will be more handy in order not to customise adyen_cc method further and to solve possible bottlenecks at manual capture and refund phases.

`adyen_moto` payment method added. Now, merchants can select the moto merchant account from order the creation page. That merchant account will be used for the transaction.

New MOTO payment flow requires major changes in the dependency injections. Since, every order can be placed with different merchant accounts; requests, http clients and related data builders are now being populated according to related MOTO merchant account that is stored in the additional_information field of the payment.

**Tested scenarios**

- Customer e-mail
- Operations on adyen_cc after upgrade to adyen_moto


Helper/Data.php
```diff
@@ -1506,6 +1513,47 @@ public function initializeAdyenClient($storeId = null, $apiKey = null)
...
 
+    /**
+     * Initializes and returns Adyen Client and sets the required parameters of it
+     *
+     * @param null|int|string $storeId
+     * @param string|null $apiKey
+     * @return \Adyen\Client
+     * @throws \Adyen\AdyenException
+     */
+    public function initializeAdyenMotoClient($motoMerchantAccount, $storeId = null)
```
candemiralp, line 1524: It could be nice to add an option parameter to `initializeAdyenClient()` method as `motoMerchantAccount` and discard the new method.

Block/Form/Moto.php
```diff
@@ -0,0 +1,295 @@
...
+        $this->configHelper = $configHelper;
+        $this->customerSession = $customerSession;
+        $this->vaultHelper = $vaultHelper;
+    }
+
+    /**
+     * @return string
+     * @throws \Adyen\AdyenException
+     */
+    public function getClientKey()
```
candemiralp, line 128: Remove this method since `clientKey` is being fetched from the data attribute of the dropdown in the frontend.

Helper/Data.php
```diff
@@ -1506,6 +1513,47 @@ public function initializeAdyenClient($storeId = null, $apiKey = null)
...
+            $storeId = $this->storeManager->getStore()->getId();
+        }
+
+        $motoMerchantAccounts = $this->adyenConfigHelper->getMotoMerchantAccounts($storeId);
+
+        $motoMerchantAccountProperties = $motoMerchantAccounts[$motoMerchantAccount];
+        $apiKey = $this->_encryptor->decrypt($motoMerchantAccountProperties['apikey']);
+
+        $client = $this->createAdyenClient();
+        $client->setApplicationName("Magento 2 plugin");
```
acampos1916, line 1538: Let's move string to constant

Helper/Data.php
```diff
@@ -1506,6 +1513,47 @@ public function initializeAdyenClient($storeId = null, $apiKey = null)
...
+        $client->setXApiKey($apiKey);
+        $moduleVersion = $this->getModuleVersion();
+
+        $client->setMerchantApplication($this->getModuleName(), $moduleVersion);
+        $client->setExternalPlatform($this->productMetadata->getName(), $this->productMetadata->getVersion());
+
+        if ($motoMerchantAccountProperties['demo_mode'] === '1') {
+            $client->setEnvironment(\Adyen\Environment::TEST);
+        } else {
+            $client->setEnvironment(\Adyen\Environment::LIVE, $this->getLiveEndpointPrefix($storeId));
```
acampos1916, line 1548: Let's check if the prefix should be in the MOTO config

acampos1916: We're going to use the main config prefix as it is configured on company level

Model/Ui/AdyenMotoConfigProvider.php
```diff
@@ -0,0 +1,283 @@
...
+     */
+    public function getConfig()
+    {
+        // set to active
+        $config = [
+            'payment' => [
+                self::CODE => [
+                    'isActive' => true,
+                    'successPage' => $this->_urlBuilder->getUrl(
+                        'checkout/onepage/success',
```
acampos1916, line 113: `successPage`: I think this might not be needed and could be removed

AlexandrosMor (approved):

view/adminhtml/templates/info/adyen_moto.phtml
```diff
@@ -0,0 +1,107 @@
...
+    <div>
+        <?= $block->escapeHtml(__('Merchant Account: %1', $block->getMotoMerchantAccount())); ?><br/>
+    </div>
+<?php endif; ?>
+
+<?php if ($_pspReference = $_info->getAdditionalInformation('pspReference')): ?>
+    <div>
+        <?php if ($_isDemoMode): ?>
+            <?= $block->escapeHtml(__('Adyen PSP Reference: ')); ?>
+            <a href="https://ca-test.adyen.com/ca/ca/accounts/showTx.shtml?pspReference=<?= $block->escapeHtml(
```
acampos1916, line 39: Can we use the `if` to only wrap the `<a>` tag? That way we don't have to duplicate the label.

Maybe even setting the URL in the header of this file with a ternary: `$_isDemoMode:https://ca-test.adyen.com/ca/ca/accounts/showTx.shtml?https://ca-live.adyen.com/ca/ca/accounts/showTx.shtml`;

I think this is done this way on an old template for CC but this could be improved here.

etc/adminhtml/system/adyen_moto_settings.xml
```diff
@@ -12,25 +12,25 @@
 -->
 <include xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Config:etc/system_include.xsd">
     <group id="adyen_moto_advanced_settings" translate="label" showInDefault="1" showInWebsite="1" showInStore="1" sortOrder="600">
-        <label>MOTO settings</label>
+        <label>MOTO</label>
```
acampos1916, line 15: Can we add a `<comment>` to the config section like in PBL? Maybe with the info from the tooltip

view/adminhtml/templates/info/adyen_moto.phtml
```diff
@@ -0,0 +1,107 @@
...
+    <?= $block->escapeHtml($block->getMethod()->getTitle()); ?>
+</div>
+<?php
+$_info = $block->getInfo();
+$_isDemoMode = $block->isDemoMode();
+?>
+
+<?php if ($block->getMotoMerchantAccount() != ""): ?>
+    <div>
+        <?= $block->escapeHtml(__('Merchant Account: %1', $block->getMotoMerchantAccount())); ?><br/>
```
acampos1916, line 31:
```suggestion
        <?= $block->escapeHtml(__('Adyen Merchant Account: %1', $block->getMotoMerchantAccount())); ?><br/>
```

acampos1916:
Can we add `Customer Email` to the tested scenarios? I would like to make sure that the customer doesn't receive the Adyen-specific information, only the payment method info.

acampos1916:
Tested scenario: operations on adyen_cc orders after upgrade.

candemiralp:
>

acampos1916 (approved):

peterojo (approved):

PR #1623, opened 2022-07-21
[PW-6706] - Add capture, cancel, refund and void commands for MOTO

candemiralp:
**Description**

Since MOTO is using different merchant account, it is required to create separate commands for capture, cancel, refund and void.

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/Gateway/Http/Client/TransactionMotoCancel.php
+++ b/Gateway/Http/Client/TransactionMotoCancel.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ *
+ * Adyen Payment module (https://www.adyen.com/)
+ *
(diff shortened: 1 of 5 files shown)


michaelpaul (approved):
Looking good, an improvement would be to add unit tests, specially for `MotoRefundDataBuilder`.

candemiralp:
> Looking good, an improvement would be to add unit tests, specially for `MotoRefundDataBuilder`.

Hey @michaelpaul, thanks for the review and suggestion. I am planning to create separate ticket for MOTO tests, I can also add this unit test during that stage.

AlexandrosMor (approved):

AlexandrosMor (approved):

PR #1634, opened 2022-07-29
[PW-6922] Add PayByLink support for headless integration

RokPopov:
**Description**
This PR is adding PBL as a supported payment method for the headless integration. In this solution, a new custom event is added that has a new observer assigned to it. The observer is setting the expiry date in the payment. `AdyenOrderPaymentStatus` model is modified so it can handle adding the payment link to payment's additional information. `ExpiryDateDataBuilder` is modified so it can fetch the expiry date from the request when there is no `stateData`.

**Tested scenarios**
* Place an order by sending the request to `/V1/carts/mine/payment-information` endpoint with `adyen_pay_by_link` as a payment method
* Make a POST request to the `/V1/adyen/orders/payment-status` endpoint. In the body, use `orderId that was received in the response from the `/payment-information endpoint` in the previous step. Make sure that `paymentLink` is received the the response.

### DIFF
--- a/Gateway/Request/ExpiryDateDataBuilder.php
+++ b/Gateway/Request/ExpiryDateDataBuilder.php
@@ -40,13 +40,25 @@ class ExpiryDateDataBuilder implements BuilderInterface
     public function build(array $buildSubject)
     {
         $paymentFormFields = $this->request->getParam('payment');
-        $expiryDate = date_create_from_format(
-            AdyenPayByLinkConfigProvider::DATE_TIME_FORMAT,
-            $paymentFormFields["adyen_pbl_expires_at"] . ' 23:59:59'
-        );
+        $expiryDate = null;
+        $payment = $buildSubject['payment']->getPayment()->getAdditionalInformation()['adyen_pbl_expires_at'];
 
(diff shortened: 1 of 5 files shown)


Gateway/Request/ExpiryDateDataBuilder.php
```diff
@@ -40,13 +40,25 @@ public function __construct(
     public function build(array $buildSubject)
     {
         $paymentFormFields = $this->request->getParam('payment');
-        $expiryDate = date_create_from_format(
-            AdyenPayByLinkConfigProvider::DATE_TIME_FORMAT,
-            $paymentFormFields["adyen_pbl_expires_at"] . ' 23:59:59'
-        );
+        $expiryDate = null;
+        $payment = $buildSubject['payment']->getPayment()->getAdditionalInformation()['adyen_pbl_expires_at'];
```
AlexandrosMor, line 44: In `AdyenPayByLinkDataAssignObserver` there is a constant for this one. Should we move it in another location and used here too

RokPopov: Thank you @AlexandrosMor, I changed to the const in all the locations

PR #1635, opened 2022-08-01
Fix typing issue as requiresManualReview expects an array

bramstroker:
**Description**
In the following commit additional data on the notification model was introduced.
https://github.com/Adyen/adyen-magento2/commit/121969c342157855c4374ef1c18b1abe361d875f

However the ternary statement is not correct as it fall backs to an empty string, but the `requiresManualReview` is type hinted as `array`, which obviously will cause a fatal PHP error.
`Adyen\Payment\Helper\CaseManagement::requiresManualReview(): Argument #1 ($additionalData) must be of type array, string given`

**Tested scenarios**

**Fixed issue**:


Morerice (approved):

michaelpaul (approved):

AlexandrosMor (approved):

Morerice:
Thanks for your contribution @bramstroker!

PR #1648, opened 2022-08-04
[PW-6911] - Resolve incompatibility with Magento 2.3.x

### LINKED ISSUE #1615: [PW-6911] No longer compatible with Magento 2.3.x
**Describe the bug**
A change was introduced in two places that is not compatible with Magento 2.3.x:

- v7 line:  https://github.com/Adyen/adyen-magento2/pull/1437
- main (v8) line: https://github.com/Adyen/adyen-magento2/pull/1436

The issue is that the code is relying on a function `getInitCustomerData`, which was introduced in this commit - https://github.com/magento/magento2/commit/4d24466fb2ce20fcd769fd4b68c9e4d3e476fe35
This commit is only present in 2.4.x of Magento (as can be seen from a list of tags for the commit).

From a technical point-of-view, 7.3.7 and 8.2.0 should have been major version upgrades, as they were not backwards-compatible. This is not really practical now, since there have been other versions released since then.
The code would need to be made compatible with 2.3.x, or there should be explicit directions that it is not compatible - either in documentation or by composer requirements.

**To Reproduce**
1. Make purchase in 2.3.x environment
2. Observe error on success page.

**Expected behavior**

**Magento version**
2.3.x

**Plugin version**
>= 7.3.7, >= 8.2.0

 - OS: any
 - Browser any
 - Version any

 - Device: any
 - OS: any
 - Browser any
 - Version any

**Additional context**

candemiralp:
**Description**

`getInitCustomerData()` function called on `success` page does not exist in Magento 2.3.x versions. Remove that additional check for existence of the customer information and reload cart directly.

**Tested scenarios**

**Fixed issue**:   Closes #1615

### DIFF
--- a/view/frontend/templates/checkout/success.phtml
+++ b/view/frontend/templates/checkout/success.phtml
@@ -18,9 +18,7 @@
     require(['Magento_Customer/js/customer-data'], function (customerData) {
         'use strict';
 
-        customerData.getInitCustomerData().done(function () {
-            customerData.reload(['cart'], true);
-        });
+        customerData.reload(['cart'], true);
     });
 </script>
 <?php if ($block->renderAction()): ?>
@@ -115,4 +113,3 @@
 </script>
 <div id='donation-container'></div>
 <?php endif; ?>
-
--- a/view/frontend/templates/checkout/success.phtml
+++ b/view/frontend/templates/checkout/success.phtml
@@ -113,3 +113,4 @@
 </script>
 <div id='donation-container'></div>
 <?php endif; ?>
+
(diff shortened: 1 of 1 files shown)


michaelpaul (dismissed):

view/frontend/templates/checkout/success.phtml
```diff
@@ -115,4 +113,3 @@
 </script>
 <div id='donation-container'></div>
 <?php endif; ?>
-
```
RokPopov, line 118: @candemiralp I think the convention is to have an empty line at the end of the file?

RokPopov (changes requested):

RokPopov (approved):

michaelpaul (approved):

PR #1649, opened 2022-08-04
[PW-6911-v7] - Resolve incompatibility with Magento 2.3.x

candemiralp:
**Description**

`getInitCustomerData()` function called on `success` page does not exist in Magento 2.3.x versions. Remove that additional check for existence of the customer information and reload cart directly.

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/view/frontend/templates/checkout/success.phtml
+++ b/view/frontend/templates/checkout/success.phtml
@@ -30,9 +30,7 @@
     require(['Magento_Customer/js/customer-data'], function (customerData) {
         'use strict';
 
-        customerData.getInitCustomerData().done(function () {
-            customerData.reload(['cart'], true);
-        });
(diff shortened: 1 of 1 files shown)


michaelpaul (approved):

view/frontend/templates/checkout/success.phtml
```diff
@@ -56,4 +54,3 @@
     </script>
     <div id="ActionContainer"></div>
 <?php endif; ?>
-
```
RokPopov, line 59: @candemiralp could you add a break line to the end of the file?

RokPopov (changes requested):

RokPopov (approved):

michaelpaul (approved):

Issue #1632, opened 2022-07-28
[PW-7006] Orders Stuck in a processing status after being completely invoiced and shipped


jordanvector:
**Describe the bug**
We have noticed on the Adyen 8.3.2 that when an order is completely invoiced and shipped correctly it is being put back into a processing status rather than a complete status like it should.  This causes a huge issue with order processing processes on the backend that depend on that status being correctly in a completed state when actually complete.

**To Reproduce**
1.  Go through typical order process with the below settings 
![Screen Shot 2022-07-28 at 1 34 30 PM (2)](https://user-images.githubusercontent.com/6155117/181601745-2e4528ac-32dd-4cca-b41c-b9b2d51522e4.png)
2.  Invoice and ship all items in the order 
3. Notice that the order status is now back to processing instead of complete

**Expected behavior**
order should be marked to complete after all items are invoiced and shipped .

**Magento version**
2.4.3

**Plugin version**
8.3.2

**Desktop (please complete the following information):**
 - OS: mac os
 - Browser edge

jordanvector:
What ends up happening is the order seems to be put back into a pending payment status for some reason.  Looking at the order that should be complete we see this status, which seems to be the only option at that point.

![Screen Shot 2022-08-01 at 4 20 42 PM (2)](https://user-images.githubusercontent.com/6155117/182246275-699c1a82-c9f8-4d39-b71b-dc53041c7ae5.png)

RokPopov:
Hi @jordanvector,

Thank you for reporting this issue with us.
Indeed, I was able to replicate the issue - it seems that the order is stuck on the status that is configured for the payment authorisation. This shouldn't be the case.
I have opened an internal ticket and we will be further investigating this.

Kind regards,
Rok,
Adyen

jordanvector:
Thank you @RokPopov I did some more testing and it appears like no change of configuration in the order processing settings resolves this issue, always seems to be bounced back into pending payment

fahu:
We experience a similar issue where the order is put back to state "new". In our case it's because we receive two notifications from Adyen:
1. Notification informing that payment is successful -> order is moved to processing
2. After a few more minutes we receive a notification informing that payment has failed -> order is moved to status "new" again.

jordanvector:
@RokPopov I have had a good look at whats happening here and I have a few thoughts and opened a PR for a fix for this

[...]

PR #1650, opened 2022-08-04
fix for #1632

jordanvector:
**Description**

This is a fix for [PW-7006] Orders Stuck in a processing status after being completely invoiced and shipped #1632. Overall the logic in the Helper/Order.php file should really be revisited.  The main issue arises from updating order state in various scenarios that overall should be avoided if possible, mainly due the inability of admin users to edit order state via the admin if any issues arise.  But essentially the source of this issue is that in a happy path of an order workflow, the order state is either kept in new or pending_payment (both scenarios the order on the adyen and magento side has been successfully authorized and payment captured), changing the order state to processing actually fixes this issue in both of those scenarios and magento then properly completes the order when all items are invoiced and shipped.

**Tested scenarios**

Tested order authorized invoiced and shipped all the way through

**Fixed issue**:  #1632 

Fixes issues where orders are stuck in a processing or new status when they should be complete.

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -257,6 +257,9 @@ class Order extends AbstractHelper
             // Else add comment
             if (!empty($status) && $maintainingState) {
                 $order->addStatusHistoryComment(__($comment), $status);
+                if (strcmp($order->getState(), MagentoOrder::STATE_NEW) == 0 || strcmp($order->getState(), MagentoOrder::STATE_PENDING_PAYMENT) == 0) {
+                    $order->setState(MagentoOrder::STATE_PROCESSING);
+                }
                 $this->adyenLogger->addAdyenNotificationCronjob(
                     'Maintaining current status: ' . $status,
                     $this->adyenLogger->getOrderContext($order)


RokPopov:
Hi @jordanvector ,

We would like to thank you for the PR that you opened.
Checking the code, some concerns popped up, such as the fact that there is a possibility of creating custom statuses for the order in Magento admin panel. The solution that you proposed is hardcoding the strings of specific statuses instead of using the dynamic ones. This seems to not be fully addressing the root of the issue, but rather the symptoms of it. 
However, please know we have included the points that you made into the investigation ticket we have opened for this and we will be looking into it.

In the meantime we will be closing this PR now but feel free to let us know if you think we are missing something.

Kind regards,
Rok,
Adyen

jordanvector:
@RokPopov just to be clear there was no hardcoding of order statuses here, this updates order **STATE** which is hugely different (and also is using the magento core states) and also considers custom order statuses if any were used as well.  Adyen should really reconsider making any changes to order state as i is highly problematic and error prone.

Issue #1588, opened 2022-06-29
[PW-6860] Order cancelled in Magento but payment is correct in Adyen


ifuents:
**Describe the bug**
Order cancelled in Magento but payment is correct in Adyen.

**To Reproduce**
1. Go to checkout and pay for order using Adyen Payment Method (Using Card Payments method)
3. Proceed with Payment and finish it successfully
4. Manually close de modal window before it closes by itself

**Expected behavior:**
Order should not be cancelled and it should finish successfully

**Magento version:**
2.4.4

**Plugin version:**
8.2.2

**Desktop:**
 - OS: Ubuntu 20.04.4 LTS
 - Browser: Firefox 101.0.1 (64-bit)

**Additional context**
After order is cancelled according to the steps to reproduce, Magento cron processes the Adyen Notification successfully and adds AUTHORISATION information to Order Notes (sales_order_status_history) but order remains cancelled.

Module is configured with default values.
Payment Method used is **Card Payments**

siimm:
Having exactly the same problem.
It seems that at some point after authorisation in checkout, an ajax call is made to backend
`POST /rest/sv/V1/internal/adyen/paymentDetails`
with a payload of:
`{"payload":"{\"orderId\":\"<ORDER_ID>\",\"cancelled\":true}","form_key":"1vlmXXXXXXXXX"}`
based on this the backend cancels the order.
However since it was actually successfully authorised, a success AUTHORISED callback is received and the order is actually paid for, but it stays in canceled state and status.

In case of correct behaviour a payload like this is posted to that endpoint:
`{"payload":"{\"details\":{\"threeDSResult\":\"ab<threeDencryptedvaluehere>c=\"},\"orderId\":\"<OREDER_ID>\"}","form_key":"XXXXXXXXXXX"}`

had to revert the module to 7.x.x in order to continue receiving orders. This happens to about 20% of total adyen_cc orders
It sounds a bit like some sort of race condition perhaps.

RokPopov:
Hi @ifuents , @siimm 

Thank you for reporting this issue.
[...]

PR #1651, opened 2022-08-08
[PW-6860] Order cancelled in Magento and status:Settled in Adyen CA

RokPopov:
**Description**
This PR is addressing the issue of the way we handle the orders that make use of the popup modals in cases of slow internet connectivity on the shopper's end. As of now, we are setting the order to cancelled state in Magento, however the `/payment-details` call is going through to Adyen. This results the orders in Magento being cancelled and at the same time settled in Adyen CA. 
This fix hides the `X` button in the `handleOnAdditionalDetails` function once the `/payment-details` call goes through, so that the shopper can't cancel the order that has already been sent to Adyen. 
**Tested scenarios**
- Place an order with a payment method that initiates the popup modal, enter the password and click `OK`. Make sure that the `X` button is not showing in the top right of the modal.
- Place an order with a payment method that initiates the popup modal, enter the password and click `OK`. Try to cancel the order with pressing `Esc` and make sure that the order is not cancelled in Magento.
-  Place an order with a payment method that initiates the popup modal, enter the password and click `OK`. Try to cancel the order with clicking outside of the popup modal. Make sure that the order is not cancelled in Magento.

**Fixed issue**:  #1588

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
@@ -312,6 +312,7 @@ define(
 
                 var self = this;
 
+                $(this.modalLabel + ' .action-close').hide();
                 var request = result.data;
                 request.orderId = self.orderId;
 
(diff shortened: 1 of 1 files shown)


michaelpaul (approved):

peterojo (approved):

phpWebStudio:
Why you didn't fix adyen-hpp-method.js also? Does it make sence?

PR #1666, opened 2022-08-12
[PW-6504] Logging review

peterojo:
**Description**

- check config before every debug logging
- cleanup redundant logs and add context where possible
- use debug or warning where applicable instead of info log


Cron/WebhookProcessor.php
```diff
@@ -65,7 +65,7 @@ public function execute()
         try {
             $this->doProcessWebhook();
         } catch (Exception $e) {
-            $this->adyenLogger->addAdyenNotification($e->getMessage() . "\n" . $e->getTraceAsString());
+            $this->adyenLogger->addAdyenWarning($e->getMessage() . "\n" . $e->getTraceAsString());
```
AlexandrosMor, line 68: Why this error is not part of the notification as it is related to the webhook process ?

peterojo: The idea is that any exception that is caught at this level means that the cron processing (which runs every minute) has terminated rather than skipping to the next notification. So I thought it makes sense to log this as a warning.

AlexandrosMor: ok thank you for the explanation :)

candemiralp (approved):

Morerice (approved):

PR #1683, opened 2022-08-26
[PW-7101] Allow custom `shopperLocale` in REST `/retrieve-adyen-payment-methods` endpoint

RokPopov:
**Description**
The purpose of this PR is to add support for custom `shopperLocale` parameter in REST `retrieve-adyen-payment-methods` endpoint for guest and logged-in users.

The corresponding API models have been updated to pass `shopperLocale` value through to `paymentMethodsHelper->getPaymentMethods()` method

**Tested scenarios**
- made a call to `/V1/guest-carts/:cartId/retrieve-adyen-payment-methods` to make sure the custom `shopperLocale` value is passed for guest users
- made a call to `/V1/carts/mine/retrieve-adyen-payment-methods` to make sure the custom `shopperLocale` value is passed for logged-in users

### DIFF
--- a/Api/AdyenPaymentMethodManagementInterface.php
+++ b/Api/AdyenPaymentMethodManagementInterface.php
@@ -24,7 +24,8 @@ interface AdyenPaymentMethodManagementInterface
      *
      * @param string $cartId
      * @param null|\Magento\Quote\Api\Data\AddressInterface
+     * @param string|null $shopperLocale
      * @return \Magento\Checkout\Api\Data\PaymentDetailsInterface
      */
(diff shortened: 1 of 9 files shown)


peterojo (dismissed):
Looks good 👍🏿
Although I don't think the extra parameter is necessary for the `Internal` endpoints since those are used only from the plugin and we don't need to send a custom locale

acampos1916 (dismissed):

peterojo (approved):

acampos1916 (approved):

PR #1685, opened 2022-08-30
[PW-7150]Add null for the moto merchant account

AlexandrosMor:
**Description**

In the ManagementHelper.php class the `$this->dataHelper->initializeAdyenClient` has wrong parameters which cause the management api config to throw an exception. The `initializeAdyenClient` has `$motoMerchantAccount` and it should be `null` when we do not use moto

**Tested scenarios**
Required Settings in admin panel with automated flow

### DIFF
--- a/Helper/ManagementHelper.php
+++ b/Helper/ManagementHelper.php
@@ -124,7 +124,7 @@ class ManagementHelper
         bool $demoMode
     ): void {
         $storeId = $this->storeManager->getStore()->getId();
-        $client = $this->dataHelper->initializeAdyenClient($storeId, $apiKey, $demoMode);
+        $client = $this->dataHelper->initializeAdyenClient($storeId, $apiKey, null, $demoMode);
 
         $management = new Management($client);
         $params = [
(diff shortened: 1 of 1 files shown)


candemiralp (dismissed):

Helper/ManagementHelper.php
```diff
@@ -206,7 +206,7 @@ private function getManagementApiService($apiKey, $environment): Management
             // API key contains '******', set to the previously saved config value
             $apiKey = $this->configHelper->getApiKey($environment);
         }
-        $client = $this->dataHelper->initializeAdyenClient($storeId, $apiKey, $environment === 'test');
+        $client = $this->dataHelper->initializeAdyenClient($storeId, $apiKey,null, $environment === 'test');
```
Morerice, line 209:
```suggestion
        $client = $this->dataHelper->initializeAdyenClient($storeId, $apiKey, null, $environment === 'test');
```

Morerice (changes requested):

Morerice (approved):

peterojo (approved):

PR #1692, opened 2022-08-31
[PW-5788] Return cctype if the pm type if missing from adyen_payment.xml

AlexandrosMor:
**Description**

The apple tx variants are self described for example: `elodebit_applepay `, `visa_applepay `, `mc_applepay`  and with this pull request the type is returned in case the name is not in the payment method [list]( https://github.com/Adyen/adyen-magento2/blob/develop/etc/adyen_payment.xml.). This list is used to map the tx variants with a payment method name for orders in magento admin 

**Tested scenarios**

- Google pay
- CC
- Ideal
- Sepadirect tokenized pm

### DIFF
--- a/Block/Info/Cc.php
+++ b/Block/Info/Cc.php
@@ -31,15 +31,8 @@ class Cc extends AbstractInfo
 
         if (isset($types[$ccType])) {
             return $types[$ccType]['name'];
-        }
-        // TODO::Refactor this block after tokenization of the alternative payment methods.
-        // This elseif block should be removed after the tokenization of the alternative payment methods (In progress: PW-6764). More general approach is required.
-        // Also remove `sepadirectdebit` from translation files.
-        elseif ($ccType == 'sepadirectdebit') {
-            return __('sepadirectdebit');
-        }
-        else {
-            return __('Unknown');
+        } else {
+            return $ccType;
         }
     }
 }
--- a/etc/adyen_payment.xml
+++ b/etc/adyen_payment.xml
@@ -77,5 +77,53 @@
             <label>Korean Local Card</label>
             <code_alt>korean_local_card</code_alt>
         </type>
+        <type id="AMEXAPPLEPAY" order="160">
+            <label>Applepay American express</label>
+            <code_alt>amex_applepay</code_alt>
+        </type>
+        <type id="DISCOVERAPPLEPAY" order="170">
+            <label>Applepay Discover</label>
+            <code_alt>discover_applepay</code_alt>
+        </type>
+        <type id="ELECTRONAPPLEPAY" order="180">
+            <label>Applepay Electron</label>
+            <code_alt>electron_applepay</code_alt>
+        </type>
+        <type id="ELOAPPLEPAY" order="190">
+            <label>Applepay Elo</label>
+            <code_alt>elo_applepay</code_alt>
+        </type>
+        <type id="ELODEBITAPPLEPAY" order="200">
+            <label>Applepay Elodebit</label>
+            <code_alt>elodebit_applepay</code_alt>
+        </type>
+        <type id="InteracAPPLEPAY" order="210">
+            <label>Applepay Interac</label>
+            <code_alt>interac_applepay</code_alt>
+        </type>
+        <type id="JCBAPPLEPAY" order="220">
+            <label>Applepay Jcb</label>
+            <code_alt>jcb_applepay</code_alt>
+        </type>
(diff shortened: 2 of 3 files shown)


RokPopov (approved):

Block/Info/Cc.php
```diff
@@ -31,15 +31,8 @@ public function getCcTypeName()
...
-        // TODO::Refactor this block after tokenization of the alternative payment methods.
-        // This elseif block should be removed after the tokenization of the alternative payment methods (In progress: PW-6764). More general approach is required.
-        // Also remove `sepadirectdebit` from translation files.
-        elseif ($ccType == 'sepadirectdebit') {
-            return __('sepadirectdebit');
-        }
-        else {
-            return __('Unknown');
+        } else {
+            return $ccType;
```
acampos1916, line 35: As the TODO says, can we remove sepadirectdebit from the translation files now? If it isn't being used anywhere else as a string shown to the user I'd say let's remove it in this PR

candemiralp, line 35: How about adding them to [this](https://github.com/Adyen/adyen-magento2/blob/develop/etc/adyen_payment.xml) instead of showing directly to the shopper?

AlexandrosMor: After talking with @candemiralp we decided that it is a solution that it is useful in case that there is not mapping. Furthermore for now it would be nice to keep the list up to date with apple pay

AlexandrosMor: @acampos1916 Could you review again ?

AlexandrosMor: yes I added all the variants :) @candemiralp

candemiralp (approved):

RokPopov (approved):

etc/adyen_payment.xml
```diff
@@ -77,5 +77,49 @@
...
+            <code_alt>mc_applepay</code_alt>
+        </type>
+        <type id="VIAPPLEPAY" order="250">
+            <label>Apple Pay Visa</label>
+            <code_alt>visa_applepay</code_alt>
+        </type>
+        <type id="GIROCARDAPPLEPAY" order="260">
+            <label>Apple Pay Girocard</label>
+            <code_alt>girocard_applepay</code_alt>
+        </type>
```
acampos1916, line 123: Would these options show in the config panel Card types? if so this might require an upgrade script so that they are enabled.

But if it does show then I don't think it would be great. As a merchant I would like to enable Visa once, for Cards, Apple Pay, Google Pay, etc. What do you think?

AlexandrosMor: The list for the config Credit Card Types is [here](https://github.com/Adyen/adyen-magento2/blob/develop/Model/Config/Source/CcType.php#L22) and we check if the allowed type is in the list. In that the case the cc type will be shown in the config.

etc/adyen_payment.xml
```diff
@@ -77,5 +77,49 @@
             <label>Korean Local Card</label>
             <code_alt>korean_local_card</code_alt>
         </type>
+        <type id="AMEXAPPLEPAY" order="160">
+            <label>Apple Pay American express</label>
```
acampos1916, line 81:
```suggestion
            <label>Apple Pay American Express</label>
```

etc/adyen_payment.xml
```diff
@@ -77,5 +77,49 @@
...
+        <type id="ELECTRONAPPLEPAY" order="180">
+            <label>Apple Pay Electron</label>
+            <code_alt>electron_applepay</code_alt>
+        </type>
+        <type id="ELOAPPLEPAY" order="190">
+            <label>Apple Pay Elo</label>
+            <code_alt>elo_applepay</code_alt>
+        </type>
+        <type id="ELODEBITAPPLEPAY" order="200">
+            <label>Apple Pay Elodebit</label>
```
acampos1916, line 97:
```suggestion
            <label>Apple Pay Elo Debit</label>
```

etc/adyen_payment.xml
```diff
@@ -77,5 +77,49 @@
...
+            <code_alt>elodebit_applepay</code_alt>
+        </type>
+        <type id="InteracAPPLEPAY" order="210">
```
acampos1916, line 100:
```suggestion
        <type id="INTERACAPPLEPAY" order="210">
```

acampos1916 (approved):

candemiralp (approved):

Issue #1521, opened 2022-05-18
[PW-6744] Multishipping + Offline payment method error


matmarchant:
**Describe the bug**
When using multishipping checkout and choosing to pay with an offline payment method e.g. check, cash on delivery, bank transfer, free. An error is thrown when you click place order because the resultCode is missing from the additional payment information and is required by adyen/module-payment/Block/Checkout/Multishipping/Success.php

**To Reproduce**
1. Go to Multishipping checkout
2. Check out using multiple addresses
3. Select any offline payment method
4. Click on Place Order
6. See error Notice: Undefined index: resultCode in adyen/module-payment/Block/Checkout/Multishipping/Success.php on line 168

**Expected behavior**
The checkout success page should be displayed

**Magento version**
2.4.3

**Plugin version**
8.1.1

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - Browser Chrome
 - Version 101.0.4951.64

**Additional context**

tnaber:
Hi @matmarchant,

Thank you for opening this issue. As the offline payments methods will not have a resultCode we should exempt them from our checker. I have created a ticket for this.

Cheers,
Titus

Morerice:
Hi @matmarchant,

This issue has been fixed in the linked PR. It will be live in the next release, later this week.

Thanks,
Jean
Adyen

PR #1697, opened 2022-09-02
[PW-6744] Add a check to Multishipping success page to verify whether it is Adyen payment method

RokPopov:
**Description**
As of now, when using multishipping checkout and choosing to pay with a non-Adyen payment method, an error is thrown when you click place order because the `resultCode` is missing from the additional payment information, which is required.

This PR introduces a solution that only requires the `resultCode` is it is Adyen's payment method.

**Tested scenarios**
- with multishipping configured in the admin panel, selected 2 shipping addresses and placed an order with non-Adyen payment method. Made sure I was redirected to the success page and the order details were displayed.

**Fixed issue**:  #1521

### DIFF
--- a/Block/Checkout/Multishipping/Success.php
+++ b/Block/Checkout/Multishipping/Success.php
@@ -13,6 +13,7 @@
 namespace Adyen\Payment\Block\Checkout\Multishipping;
 
 use Adyen\Payment\Helper\Data;
+use Adyen\Payment\Helper\PaymentMethods;
 use Adyen\Payment\Helper\PaymentResponseHandler;
 use Adyen\Payment\Model\PaymentResponse;
 use Adyen\Payment\Model\ResourceModel\PaymentResponse\Collection;
@@ -32,6 +33,11 @@ class Success extends \Magento\Multishipping\Block\Checkout\Success
         PaymentResponseHandler::REFUSED,
         PaymentResponseHandler::PRESENT_TO_SHOPPER
     );
+
+    /**
+     * @var bool
+     */
+    private $isAdyenPayment;
 
     /**
      * @var PaymentResponse[]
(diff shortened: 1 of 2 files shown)


Block/Checkout/Multishipping/Success.php
```diff
@@ -192,4 +206,14 @@ public function getPaymentFailedLabel()
     {
         return __('Payment Failed');
     }
+
+    public function isAdyenPayment()
```
Morerice, line 210:
```suggestion
    public function isAdyenPayment(): bool
```

Block/Checkout/Multishipping/Success.php
```diff
@@ -192,4 +206,14 @@ public function getPaymentFailedLabel()
...
+    {
+        return $this->isAdyenPayment;
+    }
+
+    public function setIsAdyenPayment()
```
Morerice, line 215:
```suggestion
    public function setIsAdyenPayment(bool $isAdyenPayment)
```

Block/Checkout/Multishipping/Success.php
```diff
@@ -192,4 +206,14 @@ public function getPaymentFailedLabel()
...
+    {
+        $this->isAdyenPayment = true;
```
Morerice, line 217:
```suggestion
        $this->isAdyenPayment = $isAdyenPayment;
```

Morerice (changes requested):

Block/Checkout/Multishipping/Success.php
RokPopov: 🙏🏽

Block/Checkout/Multishipping/Success.php
RokPopov: 🙏🏽

Morerice (dismissed):

Block/Checkout/Multishipping/Success.php
```diff
@@ -95,6 +103,8 @@ public function __construct(
 
         $orderIds = $this->getOrderIds();
 
+        $this->isAdyenPayment = false;
```
acampos1916, line 106: Maybe this isn't necessary? The initial value of the property would be null, which I think would be returned as false by `isAdyenPayment(): bool`

RokPopov: True that, thank you! But then a type error was thrown for `null` being returned, so I also changed the `isAdyenPayment(): bool` to `isAdyenPayment(): ?bool`

acampos1916 (approved):

peterojo (approved):

Issue #1682, opened 2022-08-26
[PW-7147] - Adyen Webhook is not updating order status in magento if we manually capture from Adyen customer area 


ashar01:
It looks like if capture set to manual or capture with delay then adyen plugin doesn't work properly. Below is the functionality of Adyen plugin after upgrade and before upgrade.
Adyen version 8
1- When we place order from magento it shows status 'Payment Review'.
2- After few minutes order status changes to 'Pending'.
3- As the capture delay is set to manual, We capture payments from Adyen customer area.
But the problem is it doesn't inform magento by changing status or adding comment that payment is being captured. We need to manually check each transaction one by one and match with magento and then create invoice which is hassle. However this works differently in Adyen version 6 plugin.

Adyen Version 6
1- When we place order from magento it shows status 'Payment Review'.
2- After few minutes order status changes to 'Pending'.
3- As the capture delay is set to manual, We capture payments from Adyen customer area.
4- Once payment is settled Adyen changes order status in magento to 'Complete' using webhook. (This no longer working in Adyen version 8 plugin)

Conclusion:
it looks like capture delay no longer supported.
It also looks like with latest version of plugin we can't capture payments from Adyen customer area

Morerice:
Hi @ashar01,

Thanks for opening this issue. I'm able to reproduce the issue where the capture request was triggered via the Adyen Customer Area. I will create an internal ticket for it.

In the meantime, you can create a capture request from the Magento admin order page, by creating an invoice. This will automatically trigger a capture request to Adyen, which will then be handled by the plugin.

Regards,
Jean
Adyen

lano-vargas:
I'm having similar issues on latest version Extension version 8.3.1. Where not matter what status we set in the configuration it's always in "Payment Review" . The setting "Order status: payment authorisation" is set to "Pending Payment" when an order is authorised via the webhook notification the status is still as  "Payment Review"  it should change to "Pending Payment" or "Pending" depending on each of your configuration mine is set to "Payment Review" unfortunately not happening.

Has anyone had to downgraded and is back working properly and is there any doc for this module on process for downgrading using composer ? Thanks in advance.

Morerice:
Hi @lano-vargas,

Are the webhooks that are being received triggered from functionality within the Adyen Magento module, or are they triggered from the Adyen customer area?

I'm asking because this issue is specifically for webhooks triggered from the customer area.

Thanks,
Jean
Adyen

lano-vargas:
[...]

PR #1714, opened 2022-09-15
[PW-7147] Handle Capture Requests triggered from Customer Area

RokPopov:
**Description**
Currently in our plugin, if a capture request is triggered from the CA, the plugin is not handling the corresponding webhook and the order is not updated on Magento.
This PR introduces a solution where Adyen and Magento invoice are created during the processing of the CAPTURE webhook. This process creates entries in both `sales_invoice` and `adyen_invoice` tables. 
With this PR, only full capture flow is supported. In cases where partial capture is initiated from the Adyen Customer Area, the webhook gets processed but no invoice is created on Magento's end. The merchant has to create the offline invoice from the admin panel.

**Tested scenarios**
- process CAPTURE notification with full order amount and see whether both invoices are created and linked
- process CAPTURE notification with partial amount of the order and make sure no invoice is created and the order can still be invoiced offline

**Fixes**:  #1682

### DIFF
--- a/Helper/Invoice.php
+++ b/Helper/Invoice.php
@@ -287,7 +287,7 @@ class Invoice extends AbstractHelper
 
         if (is_null($adyenInvoice) && $order->canInvoice()) {
                 if ($isFullAmountCaptured) {
-                   $adyenInvoice = $this->createInvoiceFromWebhook($order, $notification);
+                   $adyenInvoiceObject = $this->createInvoiceFromWebhook($order, $notification);
                 } else {
                     $order->addStatusHistoryComment(__(sprintf(
                         'Partial %s webhook notification w/amount %s %s was processed, no invoice created.
@@ -317,7 +317,7 @@ class Invoice extends AbstractHelper
         }
 
         /** @var AdyenInvoice $adyenInvoiceObject */
-        $adyenInvoiceObject = $invoiceFactory->load($adyenInvoice[InvoiceInterface::ENTITY_ID], InvoiceInterface::ENTITY_ID);
+        $adyenInvoiceObject = $adyenInvoiceObject ?? $invoiceFactory->load($adyenInvoice[InvoiceInterface::ENTITY_ID], InvoiceInterface::ENTITY_ID);
 
         $additionalData = $notification->getAdditionalData();
         $acquirerReference = $additionalData[Notification::ADDITIONAL_DATA] ?? null;
--- a/Helper/Invoice.php
+++ b/Helper/Invoice.php
@@ -263,8 +263,7 @@ class Invoice extends AbstractHelper
         $adyenInvoice = $this->adyenInvoiceResourceModel->getAdyenInvoiceByCaptureWebhook($order, $notification);
         $fullAmountCaptured = $this->adyenDataHelper->originalAmount($notification->getAmountValue(), $notification->getAmountCurrency()) >= $order->getBaseGrandTotal();
 
-        if (is_null($adyenInvoice)) {
-            if ($order->canInvoice()) {
+        if (is_null($adyenInvoice) && $order->canInvoice()) {
                 if($fullAmountCaptured) {
                     $this->createInvoiceFromWebhook($order, $notification);
                 } else {
@@ -274,7 +273,6 @@ class Invoice extends AbstractHelper
                         $notification->getAmountCurrency(),
                         $this->adyenDataHelper->originalAmount($notification->getAmountValue(), $notification->getAmountCurrency())
                     )), false);
-                }
             }
         }
 
@@ -404,7 +402,7 @@ class Invoice extends AbstractHelper
                 __('Created invoice #%1.', $invoice->getId())
             )
                 ->setIsCustomerNotified(false)
-                ->save();;
+                ->save();
         }
 
         //Create entry in adyen_invoice table
--- a/Helper/Invoice.php
+++ b/Helper/Invoice.php
@@ -239,7 +239,7 @@ class Invoice extends AbstractHelper
         $adyenInvoice->setAmount($this->adyenDataHelper->originalAmount($captureAmountCents, $order->getBaseCurrencyCode()));
         $adyenInvoice->setStatus(InvoiceInterface::STATUS_PENDING_WEBHOOK);
 
-        if(isset($invoiceId)) {
+        if (isset($invoiceId)) {
             $adyenInvoice->setInvoiceId($invoiceId);
         }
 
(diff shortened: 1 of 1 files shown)


Helper/Invoice.php
```diff
@@ -338,4 +355,69 @@ public function isFullInvoiceAmountManuallyCaptured(InvoiceModel $invoice): bool
 
         return $invoiceAmountCents === $invoiceCapturedAmountCents;
     }
+
+    /**
+     * Create both Adyen and Magento invoice from webhook if full amount is manually captured from Adyen CA
+     *
+     * @param Order $order
+     * @param Notification $notification
+     * @return AdyenInvoice
```
michaelpaul, line 364: Looks like this method is not returning an `AdyenInvoice` so that the caller `handleCaptureWebhook` can update `$adyenInvoice`.

Helper/Invoice.php
```diff
@@ -249,14 +261,19 @@ public function handleCaptureWebhook(Order $order, Notification $notification):
...
-                $notification->getOriginalReference(),
-                $notification->getPspreference(),
-                $order->getIncrementId()
-            ));
+        if (is_null($adyenInvoice) && $order->canInvoice()) {
+                if($fullAmountCaptured) {
+                    $this->createInvoiceFromWebhook($order, $notification);
+                } else {
+                    $order->addStatusHistoryComment(__(sprintf(
+                        'Partial %s webhook notification w/amount %s %s was processed, no invoice created. Please create offline invoice.',
```
michaelpaul, line 271: Since we don't have an `$adyenInvoice` in this `else` branch, should it `return` after adding this comment?

Helper/Invoice.php
```diff
@@ -249,14 +261,19 @@ public function handleCaptureWebhook(Order $order, Notification $notification):
...
+        $fullAmountCaptured = $this->adyenDataHelper->originalAmount($notification->getAmountValue(), $notification->getAmountCurrency()) >= $order->getBaseGrandTotal();
 
-        if (is_null($adyenInvoice)) {
-            throw new \Exception(sprintf(
-                'Unable to find adyen_invoice linked to original reference %s, psp reference %s, and order %s',
-                $notification->getOriginalReference(),
-                $notification->getPspreference(),
-                $order->getIncrementId()
-            ));
+        if (is_null($adyenInvoice) && $order->canInvoice()) {
```
michaelpaul, line 266: If `$adyenInvoice` is `null` and we cannot invoice (`$order->canInvoice()` is `false`), can we still `throw` the exception?

michaelpaul (changes requested):

Helper/Invoice.php
RokPopov: 👌

Helper/Invoice.php
RokPopov: Throwing an exception here as well, as the return type of the `handleCaptureWebhook()` should be `AdyenInvoice`

Helper/Invoice.php
```diff
@@ -249,14 +262,32 @@ public function handleCaptureWebhook(Order $order, Notification $notification):
     {
         $invoiceFactory = $this->adyenInvoiceFactory->create();
         $adyenInvoice = $this->adyenInvoiceResourceModel->getAdyenInvoiceByCaptureWebhook($order, $notification);
+        $fullAmountCaptured = $this->adyenDataHelper->originalAmount($notification->getAmountValue(), $notification->getAmountCurrency()) >= $order->getBaseGrandTotal();
```
candemiralp, line 265:
```suggestion
        $isFullAmountCaptured = $this->adyenDataHelper->originalAmount($notification->getAmountValue(), $notification->getAmountCurrency()) >= $order->getBaseGrandTotal();
```

Helper/Invoice.php
```diff
@@ -249,14 +262,32 @@ public function handleCaptureWebhook(Order $order, Notification $notification):
...
 
-        if (is_null($adyenInvoice)) {
-            throw new \Exception(sprintf(
-                'Unable to find adyen_invoice linked to original reference %s, psp reference %s, and order %s',
+        if (is_null($adyenInvoice) && $order->canInvoice()) {
+                if($fullAmountCaptured) {
```
candemiralp, line 268:
```suggestion
                if($isFullAmountCaptured) {
```

candemiralp (approved):

peterojo (approved):

Helper/Invoice.php
```diff
@@ -249,14 +262,32 @@ public function handleCaptureWebhook(Order $order, Notification $notification):
...
         $invoiceFactory = $this->adyenInvoiceFactory->create();
         $adyenInvoice = $this->adyenInvoiceResourceModel->getAdyenInvoiceByCaptureWebhook($order, $notification);
+        $isFullAmountCaptured = $this->adyenDataHelper->originalAmount($notification->getAmountValue(), $notification->getAmountCurrency()) >= $order->getBaseGrandTotal();
 
-        if (is_null($adyenInvoice)) {
-            throw new \Exception(sprintf(
-                'Unable to find adyen_invoice linked to original reference %s, psp reference %s, and order %s',
+        if (is_null($adyenInvoice) && $order->canInvoice()) {
+                if($isFullAmountCaptured) {
+                    $this->createInvoiceFromWebhook($order, $notification);
```
michaelpaul, line 269: I think we have to update `$adyenInvoice`, otherwise it will remain `null` and fail later (e.g. line 294):
```suggestion
                    $adyenInvoice = $this->createInvoiceFromWebhook($order, $notification);
```

peterojo (dismissed):

AlexandrosMor (dismissed):

Helper/Invoice.php
```diff
@@ -267,14 +280,40 @@ public function handleCaptureWebhook(Order $order, Notification $notification):
...
-            throw new \Exception(sprintf(
-                'Unable to find adyen_invoice linked to original reference %s, psp reference %s, and order %s',
+        $isFullAmountCaptured = $this->adyenDataHelper->originalAmount(
+            $notification->getAmountValue(),
+            $notification->getAmountCurrency()
+            ) === $order->getBaseGrandTotal();
+
+        if (is_null($adyenInvoice) && $order->canInvoice()) {
+                if ($isFullAmountCaptured) {
+                   $adyenInvoice = $this->createInvoiceFromWebhook($order, $notification);
```
michaelpaul, line 290: I've just noticed that `$adyenInvoice` is an `array` instead of `AdyenInvoice`. In this case we should set `$adyenInvoiceObject`:
```suggestion
                   $adyenInvoiceObject = $this->createInvoiceFromWebhook($order, $notification);
```
Then later on line 320, use the invoice we've just created or load an existing one from DB:

```php
        $adyenInvoiceObject = $adyenInvoiceObject ?? $invoiceFactory->load($adyenInvoice[InvoiceInterface::ENTITY_ID], InvoiceInterface::ENTITY_ID);
```

michaelpaul (approved):

AlexandrosMor (approved):

peterojo (approved):

PR #1716, opened 2022-09-19
[PW-7208] Update adyen method call for virtual products

AlexandrosMor:
**Description**

The payment method call was not triggered for virtual products as the shipping information was missing.
It should be called for products without shipping information as well
**Tested scenarios**
- with virtual product
- with normal product and shipping address 
- changing the shipping address 

Co-authored-by: @candemiralp

### DIFF
--- a/view/frontend/web/js/view/payment/adyen-methods.js
+++ b/view/frontend/web/js/view/payment/adyen-methods.js
@@ -75,6 +75,9 @@ define(
                         console.log('Fetching the payment methods failed!');
                     });
                 };
+                if (quote.shippingAddress() === null) {
+                    retrievePaymentMethods();
+                }
                 //Retrieve payment methods to ensure the amount is updated, when applying the discount code
                 setCouponCodeAction.registerSuccessCallback(function () {
                     retrievePaymentMethods();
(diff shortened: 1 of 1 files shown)


Morerice (dismissed):

michaelpaul (dismissed):

RokPopov (dismissed):

candemiralp (approved):

view/frontend/web/js/view/payment/adyen-methods.js
```diff
@@ -75,6 +75,9 @@ define(
                         console.log('Fetching the payment methods failed!');
                     });
                 };
+                if (quote.shippingAddress() === null) {
+                    retrievePaymentMethods();
```
acampos1916, line 79: Should retrievePaymentMethods be changed then to use billingAddress as well? How does retrievePaymentMethods get the shippingAddress if it is null?

AlexandrosMor: At this point for the virtual product there are not shipping information so we could retrieve the pm if it is null. The billing address is for a later step

candemiralp (approved):

peterojo (approved):

PR #1724, opened 2022-09-22
[PW-7095] Radio button config poc

AlexandrosMor:
**Description**

The buttons are displayed vertical with labels and tool tip
Screenshot 2022-09-22 at 16 44 12

**Tested scenarios**

Tested if the values are saved into the database and also if the changes from the database reflect on the UI


peterojo:
This PR is only as a POC, we will open a new PR if we choose to move forward with this for all yes/no fields

PR #1749, opened 2022-10-03
[PW-7214] - Implement combo card funding source selection for POS payments in the frontend

candemiralp:
**Description**

Debit/Credit Card can be selected on the checkout page if the card has multiple funding source. This feature is only available for combo cards (Brazil and Mexico).


Gateway/Http/Client/TransactionPosCloudSync.php
```diff
@@ -29,6 +29,9 @@
 
 class TransactionPosCloudSync implements ClientInterface
 {
+    const FUNDING_SOURCE_DEBIT = 'debit';
```
Morerice, line 32: Can we move these constants to the PointOfSale helper?

view/frontend/web/js/view/payment/method-renderer/adyen-pos-cloud-method.js
```diff
@@ -106,6 +108,34 @@ define(
 
                 return connectedTerminals;
             },
+            isFundingSourceAvailable: function () {
+                if (quote.billingAddress() === null) {
+                    return false;
+                }
+                var countryId = quote.billingAddress().countryId;
```
Morerice, line 115: Can we use `let` or `const` instead of `var` in this file?

Gateway/Http/Client/TransactionPosCloudSync.php
```diff
@@ -108,7 +111,7 @@ public function placeRequest(TransferInterface $transferObject): array
         $terminalId = $request['terminalID'];
 
         if (array_key_exists('chainCalls', $request)) {
-            $quote = $this->initiatePosPayment($terminalId, $request['numberOfInstallments']);
+            $quote = $this->initiatePosPayment($terminalId, $request['fundingSource'], $request['numberOfInstallments']);
```
AlexandrosMor, line 114: Should we check if the `fundingSource` exists ?

candemiralp: Hey @AlexandrosMor, thanks for the review! We don't need to check, because it is set by the knockout observer to `credit` by default.

AlexandrosMor: Cool, thanks!

AlexandrosMor (dismissed):

PR #1754, opened 2022-10-05
[PW-6998] Review refundOrder function

RokPopov:
**Description**
The purpose behind this PR is to update what we do in `Helper\Order::refundOrder`. 
Instead of using `$notification->getOriginalReference()` to check if it's a partial payment, we are switching to using the entries in the `adyen_order_payment` table and making sure that `adyen_order_payment` is found before calling `registerRefundNotification()` in the other if block.

**Tested scenarios**
- successful full and partial refund scenarios


peterojo (approved):

Test/Unit/Helper/OrderTest.php
```diff
@@ -218,6 +220,7 @@ public function testRefundOrderAdyenOrderPaymentNotFound()
         );
 
         $order = $this->createOrder();
+        $order->method('getData')->willReturn(['entity_id' => '55']);
```
AlexandrosMor, line 223: What is the `55` ?

AlexandrosMor (approved):

Issue #1711, opened 2022-09-15
Failed to reset reservedOrderId Deprecated Functionality: preg_match(


Nuranto:
**Describe the bug**

Found in logs : 

```
AdyenLoggerTest.ERROR: Failed to reset reservedOrderId Deprecated Functionality: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in /var/www/html/vendor/adyen/module-payment/Plugin/PaymentInformationResetOrderId.php on line 57 [] []
```

**To Reproduce**

I don't know how to reproduce.
But I guess a simple `(string)` could fix the issue

**Expected behavior**

**Magento version**
2.4.5

**Plugin version**
8.5

Morerice:
Hi @Nuranto,

Thanks for opening this issue. Indeed the solution seems to be trivial (an is_null check could also work). Could you please also specify the PHP version you're on?

Thanks,
Jean
Adyen

Nuranto:
```
PHP 8.1.10 (cli) (built: Sep 13 2022 10:17:59) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.10, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.10, Copyright (c), by Zend Technologies
```

PR #1761, opened 2022-10-07
[PW-7285][PW-7149] - Align Magento api call timeout with POS device

Morerice:
**Description**
This PR will increase the timeout of Adyen's cloud terminal api to 120s to align with the [default adyen timeout](https://docs.adyen.com/point-of-sale/pos-timeouts#device-time-out).

**Tested scenarios**
* POS Timeout
  1. Create an order from the admin backend
  2. Let POS timeout
  3. Both magento and POS stop at the same time

* Transaction after POS Timeout
  1. Repeat previous scenario
  2. Re-attempt the payment
  3. Payment successful

* Intersection between frontend and backend carts:
  1. Start creating order from admin backend for existing customer
  2. Login as the customer on the frontend and create a cart
  3. Cart created by admin and cart created by user do not intersect

**Fixes**:  
#1711


AlexandrosMor (approved):

AlexandrosMor (approved):

Model/AdyenInitiateTerminalApi.php
```diff
@@ -88,9 +91,9 @@ public function __construct(
         $client = $this->adyenHelper->initializeAdyenClient($this->storeId, $apiKey);
 
         //Set configurable option in M2
-        $posTimeout = $this->adyenHelper->getAdyenPosCloudConfigData('pos_timeout', $this->storeId);
-        if (!empty($posTimeout)) {
-            $client->setTimeout($posTimeout);
+        $this->timeout = $this->adyenHelper->getAdyenPosCloudConfigData('total_timeout', $this->storeId);
```
candemiralp, line 94: Could we use `Config` helper's new method here to get the configuration value instead of `Data` helper?

candemiralp (approved):

PR #1766, opened 2022-10-13
[PW-4938] Add 3-digit padding helper method for shopperReference

RokPopov:
**Description**
Since checkout v66 Shopper reference should be 3 digits and above which means in cases that is less the api returns error. Add a method that pads all the shopper reference before adding it to any request.

**Tested scenarios**
- made sure all the requests where the $shopperReference is relevant, have that variable padded

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1793,8 +1793,8 @@ class Data extends AbstractHelper
         }, ARRAY_FILTER_USE_KEY);
     }
 
-    public function padShopperReference($customerId)
+    public function padShopperReference($shopperReference)
     {
-        return str_pad($customerId, 3, '0', STR_PAD_LEFT);
+        return str_pad($shopperReference, 3, '0', STR_PAD_LEFT);
     }
 }
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1793,7 +1793,11 @@ class Data extends AbstractHelper
         }, ARRAY_FILTER_USE_KEY);
     }
 
-    public function padShopperReference($shopperReference)
+    /**
+     * @param $shopperReference
+     * @return string
+     */
+    public function padShopperReference($shopperReference): string
     {
         return str_pad($shopperReference, 3, '0', STR_PAD_LEFT);
     }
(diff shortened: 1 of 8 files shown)


Model/AdyenInitiateTerminalApi.php
```diff
@@ -262,7 +262,8 @@ public function addSaleToAcquirerData($request, $quote)
 
             if (!empty($recurringContract) && !empty($shopperEmail)) {
                 $saleToAcquirerData['shopperEmail'] = $shopperEmail;
-                $saleToAcquirerData['shopperReference'] = str_pad((string)$customerId, 3, '0', STR_PAD_LEFT);
+                $saleToAcquirerData['shopperReference'] = $this->adyenHelper->padShopperReference($customerId);
+//                                    str_pad((string)$customerId, 3, '0', STR_PAD_LEFT);
```
candemiralp, line 266: Is this comment relevant?

Model/Api/PaymentRequest.php
```diff
@@ -230,12 +238,12 @@ public function disableRecurringContract($recurringDetailReference, $shopperRefe
         }
     }
 
-    /**
-     * @param $shopperReference
-     * @return string
-     */
-    private function getShopperReferencePadding($shopperReference)
```
candemiralp, line 237: Shall we keep this or delete?

Helper/Data.php
```diff
@@ -1792,4 +1792,9 @@ public function parseTransactionId($transactionId)
             return is_string($index);
         }, ARRAY_FILTER_USE_KEY);
     }
+
+    public function padShopperReference($customerId)
```
candemiralp, line 1796: Let's use the variable name `shopperReference` instead of `customerId`.

Helper/Data.php
```diff
@@ -1793,7 +1793,11 @@ public function parseTransactionId($transactionId)
         }, ARRAY_FILTER_USE_KEY);
     }
 
-    public function padShopperReference($shopperReference)
+    /**
+     * @param $shopperReference
+     * @return string
+     */
+    public function padShopperReference($shopperReference): string
```
Morerice, line 1800: Can we also add the argument type?
```suggestion
    public function padShopperReference(string $shopperReference): string
```

Morerice (changes requested):

Morerice (approved):

AlexandrosMor (approved):

PR #1768, opened 2022-10-14
[PW-7265] - Implement installments and funding source selection for backend POS payments

candemiralp:
**Description**

- Installments for backend orders is now possible. Note that, a merchant should enable the installments for POS payments from configuration.
- Funding source selection for Brazil and Mexico is implemented (combo card). Shipping address and currency should match for these countries (BR => BRL, MX => MXN) in order to enable funding source dropdown.


michaelpaul (commented):
Looking good! Just a small improvement to add.

candemiralp:
> 

Hello @michaelpaul! Thanks for the review.

michaelpaul (approved):

RokPopov (approved):

PR #1770, opened 2022-10-18
[PW-7341] - MOTO payment method can't fetch MOTO accounts according to store ID

candemiralp:
**Description**

MOTO merchant account list in the backend order creation page always returns default store's MOTO accounts. 
`storeId` parameter added to account list getter and http client builders.

**Tested scenarios**

- MOTO payments across different store configuration

### DIFF
--- a/Gateway/Request/MotoMerchantAccountDataBuilder.php
+++ b/Gateway/Request/MotoMerchantAccountDataBuilder.php
@@ -50,6 +50,7 @@ class MotoMerchantAccountDataBuilder implements BuilderInterface
         }
 
         $request['body'] = $this->adyenRequestsHelper->buildMotoMerchantAccountData($motoMerchantAccount);
+        $request['clientConfig']['storeId'] = $payment->getMethodInstance()->getStore();
 
         return $request;
     }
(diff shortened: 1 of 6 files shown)


Gateway/Request/MotoMerchantAccountDataBuilder.php
```diff
@@ -50,6 +50,7 @@ public function build(array $buildSubject)
         }
 
         $request['body'] = $this->adyenRequestsHelper->buildMotoMerchantAccountData($motoMerchantAccount);
+        $request['clientConfig']['storeId'] = $payment->getMethodInstance()->getStore();
```
acampos1916, line 53: Does this mean that the Magento store ID will be sent in the request?

candemiralp: Hey @acampos1916, thanks for the review! No, that doesn't mean it is a part of the request. Actual request is being built around `$request['body']`. However, `$request['clientConfig']` this array is for helper parameters and we are not sending it in the request.

RokPopov (approved):

Issue #1736, opened 2022-09-28
[PW-7274] - Wrong brand_code on sales_order_payment table


jonathanribas:
**Describe the bug**
After a first failed payment attempt with an HPP payment method (Apple Pay / Klarna), `brand_code` is set on `sales_order_payment` table /  `additional_information` field.

On a second attempt with Credit Card payment method, `brand_code` is not updated.

There are maybe other data that is not wanted from first attempt. 

If you have auto capture enabled for Open Invoice payment methods (Klarna ...), second attempt with Credit Card will go with auto-capture too and it's maybe not wanted on your setup: https://github.com/Adyen/adyen-magento2/blob/develop/Gateway/Request/CheckoutDataBuilder.php#L120

**To Reproduce**
1. Place an order with Apple Pay but it has to be refused
2. Do a second attempt with Credit Card which will be successful
3. Check `sales_order_payment` table for this order, on `additional_information` field, `brand_code` will have `apple_pay` value

**Expected behavior**
`brand_code` should be emptied for Credit Card .

**Solution**

Reinit Additional Information only with Credit Card data on https://github.com/Adyen/adyen-magento2/blob/develop/Observer/AdyenCcDataAssignObserver.php#L82

**Magento version**
2.4.5

**Plugin version**
8.3.5

Morerice:
Hi @jonathanribas,

Thanks for opening this issue. We have created an internal ticket to investigate this further.

Thanks,
Jean
Adyen

PR #1784, opened 2022-10-24
[PW-7274] - Clear additional information before setting in the observer

candemiralp:
**Description**

Additional information in the observer collects all the information during the payment attempts. It causes unrelated information to be saved in the `additionalInformation` column of `sales_order_payment` table. That information comes from the previous payment attempts and observer doesn't clean it up. 

Before setting an additional information in the observer, now plugin clears it first.

**Fixes**:  #1736


Morerice (approved):

AlexandrosMor (approved):

jonathanribas:
Thank you guys!

Issue #1787, opened 2022-10-26
[PW-7392] - Magento webhooks are marked as processed if an exception occurs during the order save


alex-james:
**Describe the bug**
Whilst the Magento cron job is processing an AUTH webhook for an order, it encountered a MySQL deadlock exception as some other process had locked the table. The meant that no updates (status / comments etc were made to the Magento order). Instead of marking the webhook as failed and re-trying at the next cron job run, it was instead marked it as processed and despite the order being paid for it was left within Magento in a Payment Review status with no indication that an error had occurred.

The issue occurs in Adyen\Payment\Helper\Webhook::processNotification() where the following code traps any Magento Exception and instead of allowing the Exception to bubble up to the next catch statement at line 213, it instead adds a warning to the log file and continues execution as though the save had completed successfully. Removing this try / catch or re-throwing the error actually leads to the expected behaviour.

`try {
    $webhookHandler = self::$webhookHandlerFactory::create($notification->getEventCode());
    $order = $webhookHandler->handleWebhook($order, $notification, $transitionState);
    $this->orderRepository->save($order);
} catch (Exception $e) {
    $this->logger->addAdyenWarning($e->getMessage());
}`

**To Reproduce**
1. Create an order in Magentp
2. Update the Magento\Sales\Model\Order\ItemRepository::save() method to throw a \Magento\Framework\DB\Adapter\DeadlockException
3. Invoke the Magneto cron to process the webhook

**Expected behavior**
- The webhook should not be set to processed
- The error count for the webhook to be incremented by 1
- An error should be added to the webhook
- The next time the cron job runs, it should re-try the webhook

**Magento version**
2.4.2-p2

**Plugin version**
8.8.0

alex-james:
I've created https://github.com/Adyen/adyen-magento2/pull/1792 with a proposed fix for this issue, or for those who would prefer not to wait, the patch below can be used for 8.8.0 with cweagans/composer-patches

[ADYEN_8_8_0_WEBHOOK_ORDER_SAVE_EXCEPTION_HANDLER_ISSUE_1791.patch.zip](https://github.com/Adyen/adyen-magento2/files/9878254/ADYEN_8_8_0_WEBHOOK_ORDER_SAVE_EXCEPTION_HANDLER_ISSUE_1791.patch.zip)

alex-james:
Hi Jean,

[...]

PR #1792, opened 2022-10-27
Issue #1787 - Improved handling of transient exceptions when saving an order during webhook processing

alex-james:
**Description**
These changes address concerns outlines in issue #1787  regarding generic exception catching during the order save operation when processing a webhook and the current approach of ignoring transient exceptions and assuming all exceptions are not recoverable.

**Tested scenarios**
This has been tested by throwing a Deadlock exception in Magento\Sales\Model\Order\ItemRepository::save() and checking the webhook logs the error, does not mark the webhook as procesed, and increments the error count.

**Fixes**:  #1787


Morerice:
Hi @alex-james,

I'm closing this PR as mentioned in #1787. Until this issue is fixed this PR may be implemented as a partial fix.

Thanks,
Jean
Adyen

PR #1809, opened 2022-11-03
[PW-7440] - Fix broken MFTF tests

candemiralp:
**Description**

Update Magento 2 baseurl on the Github Actions pipeline.

### DIFF
--- a/.github/workflows/functional-test.yml
+++ b/.github/workflows/functional-test.yml
@@ -2,7 +2,7 @@ name: Functional Tests
 on:
   workflow_dispatch:
   pull_request:
-    branches: [ main ]
+    branches: [ develop ]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


AlexandrosMor (dismissed):

.github/workflows/functional-test.yml
```diff
@@ -2,7 +2,7 @@ name: Functional Tests
 on:
   workflow_dispatch:
   pull_request:
-    branches: [ main ]
+    branches: [ develop ]
```
Morerice, line 5: I think we can leave the MFTF tests to only be executed on a release PR, instead of having them run on every PR

candemiralp: Thank you for you review @Morerice, I converted against develop for only testing. I will revert back after completing the PR.

AlexandrosMor (approved):

AlexandrosMor (approved):

PR #1815, opened 2022-11-07
PW-7447 - Return empty array if additionalData is empty

Morerice:
**Description**
Currently, if a webhook has no additionalData, an empty string [is returned](https://github.com/Adyen/adyen-magento2/blob/a72ed17dbd0d1bfe58fa9ab101f0910c68a90887/Helper/Webhook.php#L407). This will then cause an issue when the `updateOrderPaymentWithAdyenAttributes` function, which requires an array as its third parameter is then called.

**Tested scenarios**
* Auth webhook
* Webhook w/o additional data

### DIFF
--- a/Helper/Webhook.php
+++ b/Helper/Webhook.php
@@ -279,7 +279,7 @@ class Webhook
     {
         $additionalData = !empty($notification->getAdditionalData()) ? $this->serializer->unserialize(
             $notification->getAdditionalData()
-        ) : "";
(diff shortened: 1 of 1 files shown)


candemiralp (approved):

RokPopov (approved):

Morerice:
Integration tests are failing due to github environment limitation, not actual issues with these functions

PR #1817, opened 2022-11-07
[PW-7409] Remove credit card type selection from configuration

RokPopov:
**Description**
This PR removes the available credit card type selector from the plugin configuration. 
Instead of this multi-select, we are fetching the card types from Adyen CA, utilising the `paymentMethods` call. Besides removing multi-select from configuration, the rest of the source models and related classes are kept in the source. Therefore, they can be used by instalments etc.

**Tested scenarios**
- removed the row in `core_config_data` table related to `payment/adyen_cc/cctypes`
- upgraded the database
- made sure credit cards are being fetched from Adyen Customer Area and are displayed correctly on the checkout page


view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -134,7 +134,7 @@ define(
 
                 let componentConfig = {
                     enableStoreDetails: self.getEnableStoreDetails(),
-                    brands: self.getAvailableCardTypeAltCodes(),
+                    brands: self.paymentMethodsResponse.getPaymentMethods(),
```
candemiralp, line 137: This is probably wrong. `brands` should be filled with `paymentMethods`'s `brands` array. Could you please dump the outputs of this `self.getAvailableCardTypeAltCodes()` and this `self.paymentMethodsResponse.getPaymentMethods()` and compare the results?

candemiralp (changes requested):

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -366,16 +366,22 @@ define(
...
-            getAvailableCardTypeAltCodes: function() {
-                var ccTypes = window.checkoutConfig.payment.ccform.availableTypesByAlt[this.getCode()];
-                return Object.keys(ccTypes);
+            getBrands: function() {
+                let paymentMethods =
+                    adyenPaymentService.getPaymentMethods()._latestValue.paymentMethodsResponse.paymentMethods;
+
+                for (let i = 0; i < paymentMethods.length; i++) {
+                    let paymentMethod = paymentMethods[i];
+                    if (Object.values(paymentMethod).includes("Credit Card")) {
```
candemiralp, line 381: What do you think about relying on `"type": "scheme"` instead of `"name": "Credit Card"`. If a localisation is applied in the future to name parameter, `getBrands()` function may fail.

candemiralp (approved):

AlexandrosMor (dismissed):

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -366,16 +366,22 @@ define(
...
+            getBrands: function() {
+                let paymentMethods =
+                    adyenPaymentService.getPaymentMethods()._latestValue.paymentMethodsResponse.paymentMethods;
+
+                for (let i = 0; i < paymentMethods.length; i++) {
+                    let paymentMethod = paymentMethods[i];
+                    if (Object.values(paymentMethod).includes("scheme")) {
+                        return paymentMethod.brands;
+                    }
+                }
```
michaelpaul, line 384: Here it's how you can get the latest value from an observable: 

```suggestion
                const methods = adyenPaymentService.getPaymentMethods()();
                if (!methods.paymentMethodsResponse) {
                    return [];
                }
                // search brands
                for (const method of methods.paymentMethodsResponse.paymentMethods) {
                    if (method.type === "scheme" && method.brands) {
                        return method.brands;
                    }
                }
                // brands not found
                return [];
```

michaelpaul (changes requested):

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
RokPopov: Thank you @michaelpaul!

candemiralp (approved):

PR #1824, opened 2022-11-09
[PW-7349] - Make notification IP address check optional

candemiralp:
**Description**

IP address validation is now optional. It can be configured under the `Advanced settings / Security / Check webhook's IP address` field.

**Tested scenarios**

- Webhooks


Helper/Config.php
```diff
@@ -251,6 +252,22 @@ public function getNotificationsHmacKey($storeId = null): ?string
         return $this->encryptor->decrypt(trim($key));
     }
 
+    /**
+     * Retrieve flag for notifications_ip_check
+     *
+     * @param int $storeId
+     * @return bool
+     */
+    public function getNotificationsIpCheck($storeId = null): bool
```
Morerice, line 261:
```suggestion
    public function getNotificationsIpCheck(int $storeId = null): bool
```

Helper/IpAddress.php
```diff
@@ -74,6 +82,12 @@ public function __construct(
      */
     public function isIpAddressValid($ipAddresses)
```
Morerice, line 83:
```suggestion
    public function isIpAddressValid($ipAddresses): bool
```

etc/adminhtml/system/adyen_security.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0"?>
```
Morerice, line 1: Are we going to re-create the whole section for this or can we put it in an already existing one?

Morerice (changes requested):

etc/adminhtml/system/adyen_security.xml
```diff
@@ -0,0 +1,36 @@
...
+ * Adyen Payment module (https://www.adyen.com/)
+ *
+ * Copyright (c) 2022 Adyen N.V. (https://www.adyen.com/)
+ * See LICENSE.txt for license details.
+ *
+ * Author: Adyen <magento@adyen.com>
+ */
+-->
+<include xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Config:etc/system_include.xsd">
+    <group id="adyen_pwa" translate="label" type="text" sortOrder="100" showInDefault="1" showInWebsite="1" showInStore="1">
```
michaelpaul, line 14: This setting is not related to PWA. I agree with @Morerice, this should go into another existing section.

candemiralp: Hello @michaelpaul, thanks for the review. `adyen_pwa` is definitely a typo, I am going to fix it. However, I am currently discussing the suitable position with @aziya-im. Then, this PR will be updated.

.github/Makefile
```diff
@@ -39,6 +39,7 @@ configure: n98-magerun2.phar
 	bin/magento config:set payment/adyen_giving/background_image ''
 	bin/magento config:set payment/adyen_abstract/merchant_account "${ADYEN_MERCHANT}"
 	bin/magento config:set payment/adyen_moto/active 1
+	bin/magento config:set payment/adyen_abstract/notifications_ip_check 1
```
michaelpaul, line 42: We can disable this check in CI: 

```suggestion
	bin/magento config:set payment/adyen_abstract/notifications_ip_check 0
```

etc/adminhtml/system/adyen_security.xml
candemiralp: Hello @michaelpaul, thanks for the review. Default value is false for this configuration. Moving it to the `Required Settings` can lead the merchants to a wrong direction. Since, it is an optional security check, what do you think about keeping the current position?

Morerice (approved):

michaelpaul (approved):

etc/adminhtml/system/adyen_security.xml
michaelpaul: It's fine by me. We might need to update documentation pages as well.

PR #1827, opened 2022-11-10
[PW-7472] Fix Vault token generation for 3DS2 payments

RokPopov:
**Description**
This PR addresses a bug in the recurring flow. Currently, when a shopper tries to save the card details for future payments using a 3DS2 card, the token is not created and we are not populating vault_payment_token table. The reason for this is that within the 3DS2 payment flow, in the response to /payment/details endpoint we are checking whether the payment is done with a stored alternative payment method. As this is not the case with credit card payments, this function returns false. Therefore, we execute the part of the code that creates a billing agreement.
This is not correct, hence this PR. The implemented solution checks whether the payment is done with a credit card and also recurring mode (Adyen Tokenization or Vault). If Vault, it creates a new entry in the vault_payment_token table, otherwise it proceeds with creation of billing agreement.

**Tested scenarios**
- complete 3DS2 redirect and make sure the `vault_payment_token` is populated with the correct record
- make sure the shopper can use the stored 3DS2 payment details


Helper/PaymentResponseHandler.php
```diff
@@ -164,7 +167,7 @@ public function handlePaymentResponse($paymentsResponse, $payment, $order = null
         }
 
         if (!empty($paymentsResponse['resultCode']))
```
candemiralp, line 169: Can we add curly braces here?

Helper/PaymentResponseHandler.php
```diff
@@ -247,11 +250,19 @@ public function handlePaymentResponse($paymentsResponse, $payment, $order = null
                         }
                     } else {
```
candemiralp, line 251: What do you think for adding a `elseif` block instead of just `else` with the condition of `$paymentInstanceCode === AdyenCcConfigProvider::CODE` and remove this condition from line 263?

Helper/PaymentResponseHandler.php
```diff
@@ -247,11 +250,19 @@ public function handlePaymentResponse($paymentsResponse, $payment, $order = null
...
+
+                        // if adyen tokenization set up, create entry in paypal_billing_agreement table
+                        if ($recurringMode === self::ADYEN_TOKENIZATION) {
+                            $this->recurringHelper->createAdyenBillingAgreement(
+                                $order,
+                                $paymentsResponse['additionalData'],
+                                $payment->getAdditionalInformation()
+                            );
+                            // if vault set up, create entry in vault_payment_token table
+                        } elseif ($recurringMode === self::VAULT && $paymentInstanceCode === AdyenCcConfigProvider::CODE) {
```
candemiralp, line 263: If we add `$paymentInstanceCode === AdyenCcConfigProvider::CODE` condition above then this one will be redundant.

Helper/PaymentResponseHandler.php
```diff
@@ -247,11 +250,19 @@ public function handlePaymentResponse($paymentsResponse, $payment, $order = null
...
                     } else {
                         $order = $payment->getOrder();
-                        $this->recurringHelper->createAdyenBillingAgreement(
-                            $order,
-                            $paymentsResponse['additionalData'],
-                            $payment->getAdditionalInformation()
-                        );
+                        $recurringMode = $this->configHelper->getCardRecurringMode($storeId);
+
+                        // if adyen tokenization set up, create entry in paypal_billing_agreement table
```
candemiralp, line 255:
```suggestion
                        // if Adyen tokenization set up, create entry in paypal_billing_agreement table
```

Helper/PaymentResponseHandler.php
RokPopov: I like it, way cleaner

candemiralp (approved):

AlexandrosMor (approved):

PR #1851, opened 2022-11-25
[PW-7500] - Fix installment configuration for the same minimum amounts

### LINKED ISSUE #1840: [PW-7500] - Can`t save multiple installments for one threshold
**Describe the bug**
Adyen offers to configure multiple installment rules, for a certain threshold, card type and number of installments.
If there is more than one rule that is matching for a certain credit card, the user can pick the desired number of installments in the checkout with an dropdown.

Unfortunately you cannot save multiple installment rules that have the same card type and min threshold.
This is because this line would always overwrite multiple rules, when the module configuration is saved in the backend.
https://github.com/Adyen/adyen-magento2/blob/develop/Model/Config/Backend/Installments.php#L80 

**To Reproduce**
1. Go to Backend -> Store -> Configuration -> Sales -> Payment Methods -> Adyen Payments -> Configure Payment Methods -> Card Payments -> Installments Setup
2. Create 2 new installment rules like
- 10 (Minimum Amount) 2 (Number of Installments) "Visa"
- 10 (Minimum Amount) 3 (Number of Installments) "Visa"
3. Save Configuration
4. Open the same configuration to see that only one of aboves rules is saved.

**Expected behavior**
All rules should be saved (even if they are using the same card type and minimum amount.

**Magento version**
2.4.5 but not limited to it...

**Plugin version**
8.9.0 but same issue in older versions...

**Additional context**
I've checked the Adyen Documentation and it should be possible to offer the user a choice of the number of installments.
It also works when I manually push the settings to the database.
It's just that this code prevents this config from being saved.
https://github.com/Adyen/adyen-magento2/blob/develop/Model/Config/Backend/Installments.php#L80
Looks like a defect and no purpose.

candemiralp:
**Description**

Fix installment configuration for the same minimum amounts

Add second dimension to installment sub values array.

**Tested scenarios**

- Credit card payments with installments
- Admin initiated orders with installments with MOTO

Fixes  #1840


Model/Config/Backend/Installments.php
```diff
@@ -77,7 +79,7 @@ public function beforeSave()
             $ccTypes = $data['cc_types'];
 
             foreach ($ccTypes as $ccType) {
-                $result[$ccType][$amount] = $installments;
+                $result[$ccType][$amount][$counter++] = $installments;
```
Morerice, line 82: Do we need the counter variable here? Or can we just do this instead, to automatically increment the index?

```suggestion
                $result[$ccType][$amount][] = $installments;
```

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -157,13 +157,13 @@ define(
 
                                 if (creditCardType in allInstallments) {
                                     // get for the creditcard the installments
-                                    var installmentCreditcard = allInstallments[creditCardType];
+                                    var cardInstallments = allInstallments[creditCardType];
```
Morerice, line 160: I know these variables aren't all related to this PR, but while we're at it, can we move away from `var` and have `let` or `const` instead?

Morerice (changes requested):

Model/Config/Backend/Installments.php
candemiralp: Hey @Morerice, thanks for the review. Indeed, we are not using `counter` variable in anywhere else. I removed it.

Morerice (approved):

AlexandrosMor (approved):

peterojo (approved):

PR #1855, opened 2022-11-29
[PW-7534] Use more resources for e2e tests

cenkiravul:
**Description**

- Use more resources for e2e tests to ensure more success rate in pipelines
- Seperate functional tests and e2e test workflow

**Tested scenarios**

Tested with all e2e scenarios from integration-tools-tests repository

Fixes  
PW-7534


.github/workflows/e2e-test.yml
```diff
@@ -1,19 +1,16 @@
-name: Integration Tests
```
candemiralp, line 1: Since this flow configuration file contains GraphQL tests, is the name of the flow correct? Can we count GraphQL tests as E2E also?

Morerice (dismissed):

.github/workflows/e2e-test.yml
cenkiravul: It was a mistake, thanks for noticing, fixed.

Morerice (approved):

candemiralp (approved):

PR #1876, opened 2022-12-13
[PW-7618] - Fix Adyen Pay by Link for virtual products

### LINKED ISSUE #1873: [PW-7618] - Virtual orders and payment links
Hello

We have an issue when a admin creates an order with only a virtuel product and selects Adyen payment link.
This error will appear when trying to submit order:
Error: Call to a member function getStreet() on null in /var/www/droner.dk/public_html/vendor/adyen/module-payment/Gateway/Request/CheckoutDataBuilder.php:241
Stack trace:

M2: 2.4.3-p3
Adyen: 8.11.0

candemiralp:
**Description**

Change the condition for other delivery information in the `CheckoutDataBuilder` for Adyen Pay by Link payments.

**Tested scenarios**

Fixes  #1873

### DIFF
--- a/Gateway/Request/CheckoutDataBuilder.php
+++ b/Gateway/Request/CheckoutDataBuilder.php
@@ -22,6 +22,7 @@ use Adyen\Payment\Observer\AdyenHppDataAssignObserver;
 use Magento\Catalog\Helper\Image;
 use Magento\Payment\Gateway\Request\BuilderInterface;
 use Magento\Quote\Api\CartRepositoryInterface;
+use Magento\Sales\Model\Order;
 
 class CheckoutDataBuilder implements BuilderInterface
 {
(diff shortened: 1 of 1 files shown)


RokPopov (dismissed):

Gateway/Request/CheckoutDataBuilder.php
```diff
@@ -227,22 +219,30 @@ public function build(array $buildSubject)
     }
 
     /**
-     * @param \Magento\Sales\Model\Order $order
+     * @param array $requestBody
+     * @param Order $order
      * @return array
      */
-    private function getOtherDeliveryInformation(\Magento\Sales\Model\Order $order): array
+    private function addOtherDeliveryInformation(array $requestBody, Order $order): array
```
Morerice, line 226: Can we keep this function to only require the order and return the array as before? I think it should be the responsibility of wherever this function is from to add this array to the body, not the responsibility of the function itself.

I think with this implementation the function is way too tightly bound to the location where we're calling it from.

candemiralp: Thank you for your review @Morerice, I committed that suggestion.

raoulritter (approved):

AlexandrosMor (approved):

PR #1903, opened 2023-01-05
[PW-7711] Remove brand_code unsetter for alternative payment methods

### LINKED ISSUE #1904: Open Invoice methods broken since 8.9.0
**Describe the bug**
Since version 8.9.0 Open Invoice Methods like Afterpay are not working anymore, this is because the following change in the file Observer/AdyenHppDataAssignObserver.php:

```php
$paymentInfo->unsAdditionalInformation('brand_code');
```

This has effect on Gateway/Request/CheckoutDataBuilder.php#L121 where brandCode is expected to check if this is a open invoice method, but due to the change in the observer, brandCode will be empty.

**To Reproduce**
- Place a Magento order with for example Afterpay. 
- The error message you will see is: Invalid open invoice request.
- Inspect the adyen/info.log and notice that the lineItems element is missing in the request.

**Expected behavior**
Brand code will not be unset and is available for checking in the CheckoutDataBuilder class.

**Magento version**
2.4.4

**Plugin version**
8.9.x, 8.10.x 8.11.x, 8.12.x

candemiralp:
**Description**

This part of the code was introduced on the early versions to prevent leftover parameter `brand_code` in the additional information field from the alternative payment methods if the second payment attempt was completed with adyen_cc or adyen_onclick. However, it created another issue on the headless payments, since the observer is hit twice. On the second hit, it always removes the `brand_code` field, which causes failing Klarna payments due to lack of line items.

This part was removed from `AdyenHppDataAssignObserver` class.

**Tested scenarios**

- Klarna

Fixes #1904

### DIFF
--- a/Observer/AdyenHppDataAssignObserver.php
+++ b/Observer/AdyenHppDataAssignObserver.php
@@ -89,9 +89,6 @@ class AdyenHppDataAssignObserver extends AbstractDataAssignObserver
         $data = $this->readDataArgument($observer);
         $paymentInfo = $this->readPaymentModelArgument($observer);
 
-        // Remove remaining brand_code information from the previous payment
-        $paymentInfo->unsAdditionalInformation('brand_code');
-
         // Get additional data array
         $additionalData = $data->getData(PaymentInterface::KEY_ADDITIONAL_DATA);
         if (!is_array($additionalData)) {


Morerice (approved):

peterojo (approved):

xsarus-academy:
@Morerice When will you release this fix?

PR #1922, opened 2023-01-24
[PW-6531v2] Add Test webhook configuration from management api

RokPopov:
**Description**
As a second iteration of the management API functionality within the plugin, this PR introduces a button to test the webhook configuration by sending a dummy notification with the event code AUTHORISATION.  
For the time being, only the webhooks that were configured through management API can be tested. The button is only showing if the `webhook_id` is existing in the db, meaning that the button to test the webhook configuration won't show if the webhook hasn't been configured through the management API.

**Tested scenarios**
- make sure the button is not showing if the `webhook_id` hasn't been stored in `core_config_data` db table
- make sure the button shows if the `webhook_id` is in the above mentioned db table
- make a successful call to the [management api endpoint](https://docs.adyen.com/api-explorer/Management/1/post/merchants/_merchantId_/webhooks/_webhookId_/test) for testing the webhooks

### DIFF
--- a/view/adminhtml/templates/config/webhook_test.phtml
+++ b/view/adminhtml/templates/config/webhook_test.phtml
@@ -51,15 +51,15 @@
 </script>
 <?php if ($block->isWebhookIdConfigured()) : ?>
     <?php echo $block->getButtonHtml() ?>
+    <span class="adyen_webhook_test_config" id="webhook_progress">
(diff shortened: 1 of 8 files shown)


view/adminhtml/templates/config/webhook_test.phtml
```diff
@@ -1,7 +1,7 @@
+<?php /** @var Adyen\Payment\Model\Config\Adminhtml\WebhookTest $block */ ?>
 <script>
     require([
             'jquery',
```
candemiralp, line 4:
```suggestion
            'jquery'
```

candemiralp (approved):

PR #1924, opened 2023-01-24
GraphQl - Add saveStateData mutation

dimitriBouteille:
Add `saveAdyenStateData` mutation.


lewisvoncken:
@dimitriBouteille 

I have another solution which doesn't require an additional GraphQL to save the data.

https://github.com/Adyen/adyen-magento2/compare/develop...lewisvoncken:adyen-magento2:patch-1

dimitriBouteille:
I think this mutation is required. 
In my case, the [setPaymentMethodOnCart](https://developer.adobe.com/commerce/webapi/graphql/schema/cart/mutations/set-payment-method/) and [placeOrder](https://developer.adobe.com/commerce/webapi/graphql/schema/cart/mutations/place-order/) mutation are not executed in same time ...

lewisvoncken:
Hi @dimitriBouteille 

I had the same issue that the setPaymentMethodOnCart is separate from the placeOrder and with the change it saves it to the database when it selects the payment method in the setPaymentMethodOnCart.

This means that I only execute the setPaymentMethodOnCart mutation and later the placeOrder which reduces the GraphQLs from 3 to 2.

dimitriBouteille:
@lewisvoncken I didn’t really understand how `setStateData` works, but now it’s clearer :)

PR #1929, opened 2023-01-25
[PW-7778] - Implement GooglePay as an individual payment method

Morerice:
**Description**
This PR will add GooglePay as an individual payment method. Order will be have payment method set to `adyen_googlepay`.

**Tested scenarios**
* Successful payment
* Webhook processed
* GooglePay window closed during checkout

### DIFF
--- a/Model/Methods/GooglePay.php
+++ b/Model/Methods/GooglePay.php
@@ -1,0 +1,52 @@
+<?php
+/**
+ *
+ * Adyen Payment Module
+ *
+ * Copyright (c) 2023 Adyen N.V.
+ * This file is open source and available under the MIT license.
+ * See the LICENSE file for more info.
+ *
+ * Author: Adyen <magento@adyen.com>
+ */
+namespace Adyen\Payment\Model\Methods;
+
+use Adyen\Payment\Helper\PaymentMethods\PaymentMethodInterface;
+use Adyen\Payment\Model\AdyenPaymentMethod;
+
+class GooglePay extends AdyenPaymentMethod implements PaymentMethodInterface
+{
+    const CODE = 'adyen_googlepay';
+    const TX_VARIANT = 'googlepay';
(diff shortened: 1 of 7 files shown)


view/frontend/web/js/view/payment/method-renderer/adyen-paypal-method.js
```diff
@@ -76,6 +76,10 @@ define(
                         street: formattedShippingAddress.street
                     };
                 }
+
+                // Paypal requires extra configuration
```
candemiralp, line 80: Is this related to the GooglePay?

Model/Methods/GooglePay.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+namespace Adyen\Payment\Model\Methods;
+
+use Adyen\Payment\Helper\PaymentMethods\PaymentMethodInterface;
+use Adyen\Payment\Model\AdyenPaymentMethod;
+
+class GooglePay extends AdyenPaymentMethod implements PaymentMethodInterface
+{
+    const CODE = 'adyen_googlepay';
+    const TX_VARIANT = 'googlepay';
```
candemiralp, line 20: Is there a way to support `paywithgoogle`?

view/frontend/web/js/view/payment/method-renderer/adyen-paypal-method.js
Morerice: This was a fix required for PayPal that I also included in this PR

Model/Methods/GooglePay.php
Morerice: We could somehow make it work, but since we are already asking merchants to use `googlepay` from v8 upwards, I don't think it's necessary

candemiralp (approved):

RokPopov (approved):

PR #1939, opened 2023-02-01
[PW-7769][PW-7975] - Create and use vault tokens using individual payment methods

Morerice:
**Description**
Since payment methods will now be represented by individual classes, there will be changes required to create and use tokens in vault. Since these classes will be used, we will not need the payment method classes in `Helper\PaymentMethods` so this directory can be deleted. This PR will implement this functionality for the following three payment methods:
* PayPal
* Klarna
* GooglePay

**Tested scenarios**
1. PayPal
    * Token creation
    * Payment w/token
2. Klarna
    * Token creation
    * Payment w/token
3. GooglePay
    * Token creation
    * Payment w/token


Plugin/PaymentVaultDeleteToken.php
```diff
@@ -60,14 +49,11 @@ public function __construct(
      * @return PaymentTokenInterface[]|void
      * @throws NoSuchEntityException
      */
-    public function beforeDelete(
-        PaymentTokenRepositoryInterface $subject,
-        PaymentTokenInterface $paymentToken
-    ) {
+    public function beforeDelete(PaymentTokenRepositoryInterface $subject, PaymentTokenInterface $paymentToken) {
```
candemiralp, line 52: `$subject` parameter is not being used, can we delete it?

Helper/Vault.php
```diff
@@ -198,76 +205,45 @@ public function saveRecurringPaymentMethodDetails($payment, array $additionalDat
 
     /**
      * Build the recurring data when payment is done through a payment method (not card)
+     * TODO: Still to add, check configuration to see if pm is actually enabled for tokenization
```
candemiralp, line 208: We have some checks on line 216 and 224. Is this `TODO` still valid?

Plugin/PaymentVaultDeleteToken.php
Morerice: We can't, because this is in a plugin so it has to match the magento function definition

Helper/Vault.php
candemiralp, line 208:
```suggestion
```

candemiralp (approved):

peterojo (approved):

PR #1940, opened 2023-02-02
[BUGFIX] In the setStateData also save the stateData so the placeOrde…

lewisvoncken:
…r GraphQL can be a separate query

**Description**

**Tested scenarios**

Fixes


Morerice:
Hi @lewisvoncken,

Thanks for opening this PR. Unfortunately this is not something that we want to implement since we want to avoid as much as possible to store the stateData in the database, even though this is encrypted.

Regards,
Jean
Adyen

PR #1945, opened 2023-02-06
[PW-7786] Move from shipping address to billing address

RokPopov:
**Description**
In this PR the functionality which utilizes addresses was updated to use the billing address instead of the shipping address.
The following hierarchy is implemented in `Helper/PaymentMethods::getCurrentCountryCode()` to obtain the country to be passed for the paymentMethods call:
1. Fixed countryCode config set in the AdyenHppConfigData (label: ISO Country Code (optional))
2. Country passed in request from frontend to backend (changed to billing address)
3. If customer is logged in, get country from default billing address
4. Magento Default tax country config (CONFIG_XML_PATH_DEFAULT_COUNTRY)
5. Empty string ""

**Tested scenarios**
1. **Normal Cart Logged-in User** -> Use the quote billing address (passed from the frontend) to get the country. If billing address is changed in payment method page, make sure the payment methods are re-fetched.
2. **Virtual Cart Logged-in User** -> (Quote billing address isn't set). Use the user's default billing address to get the country. If billing address is changed in payment method page, make sure the payment methods are re-fetched.
3. **Normal Cart Guest User** -> Billing address is shipping address when they first enter it. Use the quote billing address (passed from the frontend) to get the country. If billing address is changed in payment method page, make sure the payment methods are re-fetched.
4. **Virtual Cart Guest User** -> Magento Default tax country config used. If billing address is changed in payment method page, make sure the payment methods are re-fetched.


Helper/PaymentMethods.php
```diff
@@ -316,7 +323,7 @@ protected function getCurrentCountryCode($store, $country = null)
             return $country;
         }
 
-        if ($country = $this->getQuote()->getBillingAddress()->getCountry()) {
+        if ($isCustomerLoggedIn && $country = $this->getQuote()->getBillingAddress()->getCountry()) {
```
Morerice, line 326: Instead of using the session, can we check if customer is a guest customer using the quote?
Using the session here might cause some issues related to headless merchants

RokPopov: good one, thanks!

Helper/PaymentMethods.php
```diff
@@ -316,7 +318,7 @@ protected function getCurrentCountryCode($store, $country = null)
             return $country;
         }
 
-        if ($country = $this->getQuote()->getBillingAddress()->getCountry()) {
+        if (!is_null($isCustomerLoggedIn) && $country = $this->getQuote()->getBillingAddress()->getCountry()) {
```
RokPopov, line 321: `$quote->getCustomerId()` returns an int if customer is logged in and `null` if not. Therefore I am checking whether the customer ID is not null here

Helper/PaymentMethods.php
```diff
@@ -316,7 +318,7 @@ protected function getCurrentCountryCode($store, $country = null)
             return $country;
         }
 
-        if ($country = $this->getQuote()->getBillingAddress()->getCountry()) {
+        if (!is_null($customerId) && $country = $this->getQuote()->getBillingAddress()->getCountry()) {
```
RokPopov, line 321: If the customer is logged in, `$quote->getCustomerId()` will return an int. Otherwise, it returns `null`. That is why I am checking that `$customerId` is not `null`.

Morerice, line 321:
```suggestion
        if (isset($customerId) && $country = $quote->getBillingAddress()->getCountry()) {
```

Morerice (changes requested):

Morerice (dismissed):

Morerice (approved):

Helper/PaymentMethods.php
```diff
@@ -307,8 +307,10 @@ protected function getCurrentCountryCode($store, $country = null)
             return $country;
         }
 
-        if ($country = $this->getQuote()->getBillingAddress()->getCountry()) {
-            return $country;
+        $quote = $this->getQuote();
+        $customerId = $quote->getCustomerId();
+        if (isset($customerId)) {
```
candemiralp, line 312:
```suggestion
        if (!is_null($quote->getCustomer()) && !is_null($quote->getCustomer()->getDefaultBillingAddress())) {
```

Default billing address may be null for a customer if she/he didn't fill it before.
In that case, `$quote->getCustomer()->getDefaultBillingAddress()` will throw an exception.

PR #1967, opened 2023-02-21
[PW-8009] - Mount component for vault card tokens

Morerice:
**Description**
Currently we are not offering the functionality to mount the component on card tokens stored in vault. Hence, all the payments using the already created token will be cvc-less payments.

This ticket will be used to investigate how we can mount the adyen component in the checkout page on these tokens stored in the Magento vault. It will also add functionality to process a payment for a tokenized card (w/mounted component) when 3ds challenge is not triggered.

**Tested scenarios**

* Mounting the component on multiple vault tokens
* Tokenized payment w/o 3ds challenge


Gateway/Request/RecurringVaultDataBuilder.php
```diff
@@ -3,37 +3,45 @@
  *
  * Adyen Payment module (https://www.adyen.com/)
  *
- * Copyright (c) 2022 Adyen BV (https://www.adyen.com/)
+ * Copyright (c) 2023 Adyen BV (https://www.adyen.com/)
```
candemiralp, line 6:
```suggestion
 * Copyright (c) 2023 Adyen N.V. (https://www.adyen.com/)
```

view/frontend/web/js/view/payment/method-renderer/vault.js
```diff
@@ -2,7 +2,7 @@
  *
  * Adyen Payment module (https://www.adyen.com/)
  *
- * Copyright (c) 2019 Adyen BV (https://www.adyen.com/)
+ * Copyright (c) 2023 Adyen BV (https://www.adyen.com/)
```
candemiralp, line 5:
```suggestion
 * Copyright (c) 2023 Adyen N.V. https://www.adyen.com/)
```

view/frontend/web/js/view/payment/method-renderer/vault.js
```diff
@@ -11,50 +11,244 @@
...
+                paymentMethodsResponse,
+                this.handleOnAdditionalDetails.bind(this)
+            );
+
+            if (this.checkoutComponent) {
+                this.checkoutComponentBuilt(true)
+            }
+        },
+
+        // TODO: implement handleOnAdditionalDetails
```
candemiralp, line 84: Since `handleOnAdditionalDetails` wasn't implemented, what do you think about merging it to a feature branch first?

view/frontend/web/template/payment/card-vault-form.html
```diff
@@ -0,0 +1,91 @@
...
+            <div class="primary">
+                <button class="action primary checkout"
+                        type="submit"
+                        data-bind="
+                            click: placeOrder,
+                            attr: {title: $t('Place Order')},
+                            enable: isButtonActive()
+                        "
+                        disabled>
+                    <span translate="'Place Orderrrrrr'"></span>
```
candemiralp, line 85: Should we change the span's translate attribute?

view/frontend/web/js/view/payment/method-renderer/vault.js
```diff
@@ -11,50 +11,244 @@
...
+        handleOnChange: function (state, component) {
+            validTokens[this.getId()] = !!state.isValid;
+            isValidObserver(validTokens)
+        },
+
+        isButtonActive: function () {
+            return this.isActive() && this.isPlaceOrderActionAllowed() && isValidObserver()[this.getId()];
+        },
+
+        // TODO check: installments, add modal and handleOnAdditionalDetails for 3ds2
```
candemiralp, line 128: We have another TODO related to the action modal above, shall we delete this one? Also, we can create a separate ticket to investigate the installments for recurring payments.

view/frontend/web/js/view/payment/method-renderer/vault.js
Morerice: I don't think we need to cause this won't interfere with any other features/tasks that are not vault-related. We should add this in an upcoming PR.

view/frontend/web/template/payment/card-vault-form.html
Morerice: I left it as is for now to create a small difference from the normal template. Eventually this will def be fixed.

view/frontend/web/js/view/payment/method-renderer/vault.js
Morerice: Yea we can do this

view/frontend/web/template/payment/card-vault-form.html
candemiralp: Ah cool, thanks for the explanation.

peterojo (approved):

PR #1983, opened 2023-03-07
#1982: Make sure flow is stopped in case response is null

### LINKED ISSUE #1982: Warning: Undefined variable $result
> Warning: Undefined variable $result in /app/nxtk5rtoi53wm/vendor/adyen/module-payment/Controller/Process/Result.php on line 225

Under some circumstances I am getting this exception. Apparently the redirect is not happening on line 222 and flow is continuing with `$result` variable being undefined. According to Adobe the best practice is to `return $this->_redirect(..)`. In this particular case it will ensure the flow is broken and line 255 is not reached.

**Magento version**
[e.g. 2.4.5-p1]

**Plugin version**
[e.g. 8.0.14]

tim-bezhashvyly:
**Description**

Fixes #1982

**Tested scenarios**

n/a

### DIFF
--- a/Controller/Process/Result.php
+++ b/Controller/Process/Result.php
@@ -203,25 +203,25 @@ class Result extends Action
         // Receive all params as this could be a GET or POST request
         $response = $this->getRequest()->getParams();
 
-        if ($response) {
-            $result = $this->validateResponse($response);
-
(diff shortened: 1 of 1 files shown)


Morerice:
PR closed due to reason mentioned in #1982

PR #2024, opened 2023-04-11
[PW-8250] Create Orders helper

candemiralp:
**Description**

This helper will be used to extend Adyen checkout service in order to make `/orders` call to Checkout API.

**Tested scenarios**

- Order creation

### DIFF
--- a/Helper/Orders.php
+++ b/Helper/Orders.php
@@ -1,0 +1,84 @@
+<?php
+/**
+ *
+ * Adyen Payment module (https://www.adyen.com/)
+ *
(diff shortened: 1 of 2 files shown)


RokPopov (approved):

Morerice (commented):
Should we maybe rename this helper to OrdersApi or something similar to avoid confusion with the already existing Order helper?

candemiralp:
> Should we maybe rename this helper to OrdersApi or something similar to avoid confusion with the already existing Order helper?

Good catch @Morerice, thank you for the review. I will push it.

Morerice (approved):

PR #2025, opened 2023-04-11
[PW-8288] Show merchant account on googlepay component in checkout

peterojo:
**Description**

Google pay component shows "Pay Unverified Merchant" at the bottom of the popup in the checkout page. 
This PR fixes it by adding the merchant account to the component configuration.

**Tested scenarios**

Pay with Googlepay button in Checkout.

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
@@ -716,7 +716,7 @@ define(
                 // Show Merchant name for googlepay component
                 if (paymentMethod.methodIdentifier.includes('googlepay')) {
                     if ('configuration' in configuration && !('merchantName' in configuration.configuration)) {
-                        configuration.configuration.merchantName = adyenConfiguration.getMerchantAccount();
+                        configuration.configuration.merchantName = configuration.configuration.gatewayMerchantId;
                     }
                 }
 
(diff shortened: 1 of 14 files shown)


Morerice (approved):

candemiralp (approved):

RokPopov (dismissed):

view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js
```diff
@@ -712,6 +712,14 @@ define(
                         configuration.totalPriceLabel = configuration.configuration.merchantName;
                     }
                 }
+
+                // Show Merchant name for googlepay component
+                if (paymentMethod.methodIdentifier.includes('googlepay')) {
+                    if ('configuration' in configuration && !('merchantName' in configuration.configuration)) {
+                        configuration.configuration.merchantName = configuration.configuration.gatewayMerchantId;
```
peterojo, line 719: gatewayMerchantId will contain the merchants configured merchant name in a LIVE environment. So we can use this and not expose the merchant account code to the shopper

raoulritter (approved):

RokPopov (approved):

carlosperales95:
Hey guys, I double checked again and for TEST, `gatewayMerchantId` will show their merchant account name (as adyen is the gateway, and their ID is their Adyen merchant account). 

`merchantName` however, is the name that is rendered in the payment sheet. In TEST environment, or if a merchant isn't recognized, a “Pay Unverified Merchant” message is displayed in the payment sheet.

So maybe it would be worth to review this again. For TEST, we do not want to show “Pay Unverified Merchant”, so we could simply use gatewayMerchantId. 

For LIVE, we do not just want to expose Adyen merchant account names to shoppers, so `merchantName` could be used.

PR #2032, opened 2023-04-14
[ECP-9073] FIX: Prevent deadlock exception caused by concurrent execution of adyen_payment_process_notification and aggregate_sales_report_order_data

### LINKED ISSUE #1933: Order is moved to status "closed" after shipment if notification processing failed before
**Bug Description**
If a payment notification could not get processed at the first try (e.g. due to a deadlock exception) the order is moved to state `closed` after an shipment is created even though it has not been refunded.

image

With #1787 the Adyen module should be able to correctly handle failures that occur while processing notifications. However, it seems that this bug was overlooked during testing.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a new order.
2. "Fake" an exception during notification processing (e.g. a Deadlock exception).
3. Make sure that the notification gets processed after it failed for the first time (and therefore the order status is "processing").
4. Ship the order.
5. Take not that the order is in status "closed" instead of the desired status "complete".

**Expected behaviour**
Order is moved to status "complete".

**Magento version**
2.4.3-p3

**Plugin version**
8.11.1

**Additional context**
This issue is independent of browser or OS.

fahu:
**Issue**
### Issue
Occasionally, we encounter a deadlock exception when the `aggregate_sales_report_order_data` cron job runs. This results in orders being left in an inconsistent state.

**Cause**
The `adyen_payment_process_notification` and `aggregate_sales_report_order_data` cron jobs can run concurrently and may attempt to access/modify the same `sales_order_item` row, leading to a deadlock.

As the Adyen `adyen_payment_process_notification` cron job does not have a proper rollback strategy (as requested here: https://github.com/Adyen/adyen-magento2/issues/1933), the affected order is left in an invalid state, such as missing the invoiced amounts.

**Solution**
To address this issue, we have made the `adyen_payment_process_notification` cron job dependent on the completion of the `aggregate_sales_report_order_data` job. This ensures that the `adyen_payment_process_notification` job will not run until the `aggregate_sales_report_order_data` job has completed.

**Tested scenarios**
- [x] General processing of Adyen notifications still works
- [x] The `adyen_payment_process_notification` is not executed while the cronjob `aggregate_sales_report_order_data` runs.

Fixes  #1933 

### Remark
I strongly believe that the `WebhookProcessor` should still implement a proper rollback strategy in case of a failure are described with this comment: https://github.com/Adyen/adyen-magento2/issues/1933#issuecomment-1432738663

As you can see this specific issue can happen with any Magento instance that uses the Adyen Magento plugin and is not caused by other third-party extensions.


candemiralp:
Hello @fahu,

Thank you for your contribution. However, this fix can not be applied due to following reasons.

1. [Magento_Cron:etc/crontab.xsd](https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Cron/etc/crontab.xsd) doesn't contain a node definition for `depends` keyword.
2. After fix applied on the test environment, following error message was observed on the cron logs and webhooks could not be processed.
```
In Filesystem.php line 168:
                                                                               
  Invalid Document                                                             
  Element 'depends': This element is not expected. Expected is ( config_path   
  ).                                                                           
  Line: 17
```

Since this fix can't be applied, I am closing the PR. But, we will continue investigating potential fixes to prevent deadlock exception during webhook processing.

Best Regards,
Can

PR #2033, opened 2023-04-14
Fixed invalid method call

barbazul:
The addWarning method is not defined in AdyenLogger

**Description**
When the order matching fails, the call to log the error is using a method that is not defined

**Tested scenarios**

* Chack payment status with wrong order number


peterojo (approved):

peterojo:
@barbazul the method is actually defined in `\Monolog\Logger` which `AdyenLogger` extends. So this PR is not necessary, but I'm approving this PR for uniformity sake, since this is the only instance in the plugin where we call `\Monolog\Logger::addWarning` directly

candemiralp (approved):

Issue #1852, opened 2022-11-25
Webhooks failing with Maximum Line Length Limit Exceeded


josh-carter:
**Describe the bug**
Webhooks were failing with an error "Maximum line length limit exceeded" within the Adyen dashboard under developers > webhooks > troubleshoot. 

The issue turned out to be due to the length of the content-security-policy/content-security-policy-report-only response header body from the `adyen/process/json` request

Unclear as to whether the change that caused the problem was on the Adyen side or whether it's just come to light as the merchant in question had a large set of Content security policies.

This also seemed to only be affecting Klarna/PayPal orders.

**To Reproduce**
1 - Setup a Magento/Adobe Commerce instance with a large set of content security policies
2 - Test Webhook from Adyen dasbhoard
3 - Get Maximum Line Length Limit Exceeded error message 

**Expected behavior**
No errors handling notification webhooks no matter on the size of the content-security-policy response headers

**Magento version**
v2.4.5.-p1

**Plugin version**
v8.3.2

Adyen Support Ticket  - #3448958

xgerhard:
Any update on this?

Morerice:
Hi @josh-carter

Apologies for the lack of response on this issue. The problem is not specific to the magento2 plugin, but it's more of an issue related to internal webhook functionality.

As @xgerhard the most common solution we see to this issue is to clear as much as possible the CSP list. Other than that there's no know solution for now. @josh-carter could you look into this solution?

Regards,
Jean
Adyen

aligent-lturner:
I would suggest that the Adyen module could remove the CSP header if the request is to the notification controller. The header is meaningless outside of a browser context anyway.

oezerenc:
Is there any update on this. We are using CSP and every time we change our config about CSP, we have to log into Adyen backend and test the webhooks, if they are still working. When will we see here a solution for this?

I agree with @aligent-lturner. The csp header can be removed by the module, because it is not relevant anyway outside of a browser context.

aligent-lturner:
Ultimately, the problem is at the Adyen end - the presence of a large CSP header shouldn't result in this error anyway, if the response is being handled appropriately.

aligent-lturner:
> Hi @josh-carter,
> 
[...]

PR #2039, opened 2023-04-26
Fix: Remove CSP headers from JSON endpoint

SamJUK:
**Description**

Removed the CSP headers from the Json controller, to fix the "Maximum line length limit exceeded" webhook error for  stores with large CSP headers. 

Fixes  
#1852

### DIFF
--- a/Controller/Process/Json.php
+++ b/Controller/Process/Json.php
@@ -33,11 +33,12 @@ use Magento\Framework\Exception\LocalizedException;
 use Magento\Framework\Serialize\SerializerInterface;
 use Symfony\Component\Config\Definition\Exception\Exception;
 use Magento\Framework\HTTP\PhpEnvironment\RemoteAddress;
+use Magento\Csp\Api\CspAwareActionInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


candemiralp:
Hello @SamJUK,

Thank you for your contribution. We are going to test this fix and approve the PR after the check.

Best Regards,
Can

SamJUK:
Hey,

Is there any update on this?

Thanks
Sam

RokPopov:
Hey @SamJUK,

Could you please check my comment in the GitHub issue [here](https://github.com/Adyen/adyen-magento2/issues/1852#issuecomment-1547908291)? Testing the solution, it doesn't seem to work for the reason of Magento setting the CSP headers from the csp_whitelist.xml file after the execution of our Json controller, so whatever we do in `modifyCsp` will be overwritten by Magento at a later stage of request processing. 

May I ask, for you the modifyCsp solves the issue and the headers don't get re-set from the csp_whitelist.xml after the `modifyCsp` is executed within the Json controller?

Cheers,
Rok

PR #2037, opened 2023-04-24
Added missing payment methods to PaymentMethodUtil

carlosperales95:
**Description**
ACH variant was missing in the MANUAL_CAPTURE_SUPPORTED_PAYMENT_METHODS array
Walley and Walley_b2b variants were missing in OPEN_INVOICE_PAYMENT_METHODS array

PR #2042, opened 2023-05-02
Release 8.18.1

RokPopov:
Fixes:
- #2036 
- #2037 
- #2041


candemiralp (approved):

Helper/Invoice.php
```diff
@@ -258,7 +258,10 @@ public function createAdyenInvoice(
         $adyenInvoice = $this->adyenInvoiceFactory->create();
         $adyenInvoice->setPspreference($pspReference);
         $adyenInvoice->setAdyenPaymentOrderId($adyenOrderPayment[OrderPaymentInterface::ENTITY_ID]);
-        $adyenInvoice->setAmount($this->adyenDataHelper->originalAmount($captureAmountCents, $order->getBaseCurrencyCode()));
+        $adyenInvoice->setAmount($this->adyenDataHelper->originalAmount(
```
acampos1916, line 261: Would it be safer to use the ChargedCurrency helper here? https://github.com/Adyen/adyen-magento2/blob/main/Helper/ChargedCurrency.php#L279-L293

acampos1916 (approved):
LGTM besides comment above, which can be a separate ticket

Helper/Invoice.php
RokPopov: @acampos1916 you are right about this one, but we have an investigation ticket open for the bulk solution, we plan to go through all the fields that contain amount and currency and review whether we are including the charged currency logic to differentiate between the base amount/order amount and fix the places where we aren't

PR #2059, opened 2023-05-19
#1852 Add Observer to remove CSP headers from Adyen requests

josh-carter:
**Description**

- Adds observer on core event ``
- Checks action name, if its an Adyen request, remove the CSP headers

**Tested scenarios**

- Tested all payment flows, currently working in production for a client
- Tested notifications no longer fail when using a large amount of CSP rules

Fixes  
#1852


josh-carter:
Hi @RokPopov 

Yep as i mentioned in the issue #1852 I think a better approach would be to make the routes that it removes the CSP headers from configurable via DI, i could update this PR to include a better more maintainable approach asap

peterojo:
Hi @josh-carter!
Sorry for the long delay, and thanks for your solution. 
Would you be able to update your PR to only remove the headers for the `/adyen/process/json` endpoint?

peterojo (approved):

RokPopov (approved):

PR #1991, opened 2023-03-10
[PW-8242] Mount action component on card vault

candemiralp:
**Description**

Action component mounted for the card vault payment methods. CVC is now required to complete vault payments. 3DS challenge will be triggered for the required cases.

**Tested scenarios**

- Card tokenization with Magento Vault and Adyen One Click
- 3DS card vault payments with CardOnFile recurring processing model
- 3DS one click payments with CardOnFile recurring processing model

Fixes #1980

PR #2062, opened 2023-05-22
[PW-8378] Make CVC field configurable for card vault payments

RokPopov:
**Description**
This PR introduces a new configuration field in the admin panel for `Magento Vault` recurring mode. It enables the merchants that have the ADP `skipCVCForOneClick` enabled in Adyen Customer Area to control whether the CVC field is required or not for vault payments. 

Newly added admin config field and correlated config path in `core_config_data` db table solve the issue introduced by [this PR](https://github.com/Adyen/adyen-magento2/pull/1991) where we added the checkout component for card vault payments under the assumption that all the CardOnFile payments require CVC field, which is not the case. 

**Tested scenarios**
- configure recurring mode to `Magento Vault` in admin panel
- set `Require CVC` field to `Yes` and confirm the CVC field is required in the vault form when the shopper tries to pay with the stored card, confirm payment is successful
- set `Require CVC` field to `No` and confirm the CVC field is not required in the vault form when the shopper tries to pay with the stored card, confirm payment is successful

### DIFF
--- a/Model/Ui/AdyenCcConfigProvider.php
+++ b/Model/Ui/AdyenCcConfigProvider.php
@@ -169,6 +169,15 @@ class AdyenCcConfigProvider implements ConfigProviderInterface
             $config['payment']['adyenCc']['installments'] = [];
         }
 
+        // fetch the config value of require_cvc
+        $cvcReq = $this->configHelper->getConfigData(
+            'require_cvc',
+            Config::XML_ADYEN_CC_VAULT,
+            $storeId,
+            true);
+
+        $config['payment']['adyenCc']['cvcRequired'] = $cvcReq;
+
         return $config;
     }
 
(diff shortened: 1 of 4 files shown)


Model/Ui/AdyenCcConfigProvider.php
```diff
@@ -169,6 +169,14 @@ public function getConfig()
             $config['payment']['adyenCc']['installments'] = [];
         }
 
+        // check if cvc is required
+        $config['payment']['adyenCc']['cvcRequired'] = $this->configHelper->getConfigData(
```
peterojo, line 173: @RokPopov would you mind using either (`cvc_required` & `cvcRequired`) or (`require_cvc` & `requireCvc`) throughout? So that we keep the naming consistent with the config path

candemiralp (approved):

candemiralp (approved):

PR #2070, opened 2023-05-26
PW-8374 - Add support for vault on RC-V9 for wallet payment methods

raoulritter:
Due to the new separate payment methods on RC-V9. In order to have vault on the wallet payment methods. 

**Description**

Remove validation for cards and Tx variants as this shouldn't be gotten from there with the changes on RC-V9. 

**Tested scenarios**

Tested vault payment methods on express checkout. 

Fixes


candemiralp:
This issue will be resolved as a part of token handling refactoring in #2122. No need to merge this PR anymore.

PR #2073, opened 2023-05-30
[PW-8252] Implement multi giftcard flow

candemiralp:
**Description**

This pull request introduces the new implementation for giftcard payments. Recently, `adyen_hpp` was deprecated completely. Previous implementation of giftcard payments required `hpp_fallback` page to be active. However, with the recent deprecation, it is not possible anymore.

This new implementation uses `/order` endpoint of Adyen Checkout API, which creates an order on Adyen's side and allow multiple payments for this order. In this case, the plugin allows one or more giftcard payments which can be topped-off with another payment method (Cards, Klarna, PayPal etc.).

All of the payment is now completed on the checkout instead of redirecting shopper to Adyen hosted payment page for giftcard payments.


Gateway/Request/GiftcardDataBuilder.php
```diff
@@ -0,0 +1,65 @@
...
+     */
+    private function validateGiftcardStateData(array $stateDataArray): array
+    {
+        foreach ($stateDataArray as $key => $item) {
+            $stateData = json_decode($item['state_data'], true);
+            if (!isset($stateData['paymentMethod']['type']) ||
+                !isset($stateData['paymentMethod']['brand']) ||
+                $stateData['paymentMethod']['type'] !== 'giftcard') {
+                unset($stateDataArray[$key]);
+            }
```
hossam-adyen, line 60: How about using property accessors for those cases?

```suggestion
            $paymentMethodType = $propertyAccessor->getValue($stateData, '[paymentMethod] [type]' )
            if ($paymentMethodType !== 'giftcard') {
                unset($stateDataArray[$key]);
            }
```

Model/Api/AdyenGiftcard.php
```diff
@@ -0,0 +1,101 @@
...
+{
+    private StateDataCollection $adyenStateData;
+    private CartRepositoryInterface $quoteRepository;
+    private Data $adyenHelper;
+
+    /**
+     * @param StateDataCollection $adyenStateData
+     * @param CartRepositoryInterface $quoteRepository
+     * @param Data $adyenHelper
+     */
```
hossam-adyen, line 32: any reason to have those attributes? 
I think since we have the function input parameters are typed, maybe they are no longer needed

Gateway/Request/GiftcardDataBuilder.php
candemiralp: Hello @hossam-adyen,

Thank you for the review and suggestion. The component `PropertyAccess` is not one of requirements for Magento 2 and for our plugin and I consider adding this dependency just for this case would be an overkill. What is your opinion about this one?

hossam-adyen (approved):

peterojo (approved):

Issue #1942, opened 2023-02-02
 Deprecated Functionality: number_format(): Passing null to parameter #1 ($num) of type float is deprecated


ryanpalmerweb:
Hi Adyen,

We are seeing the following message appearing in our order history for MobilePay orders

Deprecated Functionality: number_format(): Passing null to parameter #1 ($num) of type float is deprecated in ... vendor/adyen/module-payment/Helper/Data.php on line 372

We are using version 8.13 of the Adyen module and PHP 8.1

Thanks,
Ryan

Morerice:
Hi @ryanpalmerweb,

Thanks for opening this issue and apologies for the delayed response. Are you seeing this error message for all payments using MobilePay or only a select few?

Also, line 372 of `Data.php` on 8.13.0 points to a [doc block](https://github.com/Adyen/adyen-magento2/blob/58d86ce58be1f3f974f541a584435e7fc2c2a0fe/Helper/Data.php#L372). Can you check if you have modified the `Data.php?

Regards,
Jean
Adyen

Morerice:
Hi @ryanpalmerweb,

We will be closing this issue due to inactivity. However if the issue persists, we still encourage you to open a ticket with our support team by sending a mail to support@adyen.com.

Regards,
Jean
Adyen

maaarghk:
@Morerice  please check out @thomas-kl1 pull request. this bug causes a crash when handling notifications:

```
[2023-07-04T10:50:04.053170+00:00] AdyenLoggerTest.ADYEN_NOTIFICATION: Critical error occurred. Notification 94406 had an error: Deprecated Functionality: number_format(): Passing null to parameter #1 ($num) of type float is deprecated in vendor/adyen/module-payment/Helper/Data.php on line 376 
 #0 [internal function]: Magento\Framework\App\ErrorHandler->handler()
#1 vendor/adyen/module-payment/Helper/Data.php(376): number_format()
#2 vendor/adyen/module-payment/Helper/AdyenOrderPayment.php(128): Adyen\Payment\Helper\Data->formatAmount()
#3 vendor/adyen/module-payment/Helper/Webhook/CaptureWebhookHandler.php(101): Adyen\Payment\Helper\AdyenOrderPayment->refreshPaymentCaptureStatus()
#4 vendor/adyen/module-payment/Helper/Webhook.php(191): Adyen\Payment\Helper\Webhook\CaptureWebhookHandler->handleWebhook()
[...]

PR #1975, opened 2023-02-27
php8 fix

skeutmeier:
**Description**

fix for PHP 8. If Products without a price are in cart, it will result in a NULL parameter on number_format, which is not allowed anymore in PHP 8

**Tested scenarios**

added bundle products to cart and be able to checkout without php error

Fixes  
fixes #1942

PR #2076, opened 2023-06-01
Fix passing null to float

thomas-kl1:
**Description**
For BC reasons I don't update the method signature so I cast the value passed to number_format as float. 

Fixes [Deprecated Functionality: number_format(): Passing null to parameter #1 ($num) of type float is deprecated](https://github.com/Adyen/adyen-magento2/issues/1942)


peterojo:
Hi @thomas-kl1 ! thanks for your contribution, but I'm closing this PR in favor of #1975

thomas-kl1:
Why the preference for a new conditional structure over a very simple cast?

PR #2077, opened 2023-06-01
Fix case where order does not have a quote ID

thomas-kl1:
**Description**
Fix the following error and improve code quality:

```
report.CRITICAL: Error when running a cron job {"exception":"[object] (RuntimeException(code: 0): Error when running a cron job at /app/x/vendor/magento/module-cron/Observer/ProcessCronQueueObserver.php:362, TypeError(code: 0): Argument 1 passed to Adyen\\Payment\\Helper\\StateData::getStateData() must be of the type int, null given, called in /app/x/vendor/adyen/module-payment/Gateway/Request/ChannelDataBuilder.php on line 38 at /app/x/vendor/adyen/module-payment/Helper/StateData.php:86)"} []
```

### DIFF
--- a/Gateway/Request/ChannelDataBuilder.php
+++ b/Gateway/Request/ChannelDataBuilder.php
@@ -19,8 +19,7 @@ use Magento\Sales\Model\Order;
 
 class ChannelDataBuilder implements BuilderInterface
 {
-    /** @var StateData */
-    private $stateDataHelper;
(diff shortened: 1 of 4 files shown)


thomas-kl1:
@candemiralp Please could you at least update with a comment and a reason why you close this MR?
If it has been fixed internally, could you refer to the commit ref-ID please?

PR #2085, opened 2023-06-08
[PW-6340] Use payment method svg logos instead of png

hossam-adyen:
**Description**
It's recommended that the SVG version of the logo should be used instead of the PNG. 

**Tested scenarios**

Manuel test shows in checkout\payment step that if you have payment method icon in your assets in SVG or PNG then it will show it, if not it will get the SVG version from Adyen payment method logos in SVG format

See: https://docs.adyen.com/online-payments/build-your-integration?platform=iOS&integration=API+Only&version=70#payment-method-logos

### DIFF
--- a/view/frontend/web/template/payment/hpp-form.html
+++ b/view/frontend/web/template/payment/hpp-form.html
@@ -30,7 +30,6 @@
             <label data-bind="attr: {'for': 'adyen_' + paymentMethod.methodIdentifier}" class="label">
 
                 <!-- ko if: icon -->
-
                 <img data-bind="attr: {
                             'src': icon.url,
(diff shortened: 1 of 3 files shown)


view/frontend/web/template/payment/hpp-form.html
```diff
@@ -30,9 +30,12 @@
             <label data-bind="attr: {'for': 'adyen_' + paymentMethod.methodIdentifier}" class="label">
 
                 <!-- ko if: icon -->
+
```
hossam-adyen, line 33:
```suggestion
```

RokPopov (approved):

candemiralp (approved):

PR #2101, opened 2023-06-30
Switch logic between X-Forwarded-For and standard RemoteIP

florisschreuder:
**Description**
As I understand it:

- Adyen wants to get the _shopper IP_
- The X-Forwarded-For header is used to store the _original Client IP_

So I think the logic should be as follows:
If X-Forwarded-For header is present, this should be the client IP. If no X-Forwarded-For header is present, the standard 'remote IP' should be the client IP.

**Tested scenarios**

Fixes


peterojo:
Hi @florisschreuder! Thanks for the suggestion, we'll investigate this solution and get back to you ASAP

candemiralp (approved):

peterojo (approved):

PR #2103, opened 2023-07-04
[PW-8225] Enable PBL for Admin API

RokPopov:
**Description**
This PR makes PBL accessible for admin api. This involves updating the can_use_checkout flag for `adyen_pay_by_link` to 1, which results in `adyen_pay_by_link` payment method being returned in the `/payment-methods` call. However, since no entry is added in the `adyen-methods.js` file, PayByLink is still not shown in the frontend.

**Tested scenarios**
- make a /payment-methods call with REST and make sure that PBL is returned in the response
- make sure that on the checkout page of the storefront, PBL is not visible as a payment option
- make sure that you can make a payment using PBL via admin orders


peterojo (dismissed):

peterojo (approved):

raoulritter (approved):

khushboo-singhvi (approved):

etc/config.xml
```diff
@@ -407,7 +407,7 @@
         <days_to_expire>1</days_to_expire>
         <payment_action>order</payment_action>
         <is_gateway>1</is_gateway>
-        <can_use_checkout>0</can_use_checkout>
+        <can_use_checkout>1</can_use_checkout>
```
candemiralp, line 410: Did you check the response of Magento's internal payment-methods call? Pay-by-link shouldn't be in the available payment methods list for headless payments.

candemiralp (changes requested):

candemiralp (approved):

PR #2136, opened 2023-08-01
Added notification about creditmemo creation

Lemongrabb:
**Description**
Normally when we create a creditmemo in Magento it is followed by an email notification about the refund.
This email should be also sent when we refund a payment via the Adyen dashboard.

**Tested scenarios**
Email is sent whenever we refund an order via the Adyen dashboard.


peterojo:
Hi @lemongrab!
Thanks so much for your contribution! When creating a credit memo from the admin area, a merchant has the option whether or not to send a copy of the credit memo to the shopper (see attached screenshot.
Screenshot 2023-08-09 at 11 30 13
Always sending the credit memo email when the merchant did not expressly create a credit memo in Magento would be in contrast to this approach. 
Therefore, for the moment we have decided not to go forward with your change.
Best regards

PR #2144, opened 2023-08-03
[ECP-8394] Use dependency injection on webhook handler factory

### LINKED ISSUE #2048: [PW-8394] Avoid using "new" to instantiate classes
Yes, it is related to a problem. I'm trying to override the class. But Adyen module is instantiating the class using "new". This is not in accordance with the Magento best practices and thus making it not open for customisation. The affected class can be found here https://github.com/Adyen/adyen-magento2/blob/develop/Helper/Webhook/WebhookHandlerFactory.php#L166

The solution is to avoid using the "new" to instantiate a class. Instead, Magento's constructor dependency injection can be used.

There is no alternative for this issue.

**Additional context**
N/A

khushboo-singhvi:
**Description**
Use dependency injection on webhook handler factory. Implementation and refactoring the WebhookHandlerFactory and Webhook Helper. Also, defining the arguments for the serializer for all the webhook handlers in di.xml

**Tested scenarios**
- Testing the payment with processing notifications and handling scenarios

Fixes #2048

### DIFF
--- a/Helper/Webhook.php
+++ b/Helper/Webhook.php
@@ -3,7 +3,7 @@
  *
  * Adyen Payment Module
  *
- * Copyright (c) 2022 Adyen N.V.
+ * Copyright (c) 2023 Adyen N.V.
  * This file is open source and available under the MIT license.
  * See the LICENSE file for more info.
  *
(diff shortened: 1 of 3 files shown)


Helper/Webhook.php
```diff
@@ -109,7 +81,7 @@ public function __construct(
         $this->logger = $logger;
         $this->orderHelper = $orderHelper;
         $this->orderRepository = $orderRepository;
-        self::$webhookHandlerFactory = $webhookHandlerFactory;
+        $this->webhookHandlerFactory = $webhookHandlerFactory;
```
hossam-adyen, line 84: Good change, I don't see any valid reason why this variable was static at the first place :)

Helper/Webhook/WebhookHandlerFactory.php
```diff
@@ -12,252 +12,94 @@
...
-                    self::$orderHelper,
-                    self::$caseManagementHelper,
-                    self::$serializer,
-                    self::$adyenLogger,
-                    self::$chargedCurrency,
-                    self::$configHelper,
-                    self::$invoiceHelper,
-                    self::$paymentMethodsHelper
-                );
+                return $this->authorisationWebhookHandler;
```
hossam-adyen, line 73: I think there is a way to remove this switch statement to be in the config and DI but this is Okay also

hossam-adyen (approved):

candemiralp (approved):

hossam-adyen (approved):

PR #2150, opened 2023-08-08
[ECP-8550] Refactor dynamic table dropdown source models

candemiralp:
**Description**

Refactor dynamic table dropdown source models according to Magento 2 documents defined as [here](https://developer.adobe.com/commerce/php/tutorials/admin/create-dynamic-row-configuration/#step-3-create-the-block-class-to-describe-a-column-with-drop-down-input).

**Tested scenarios**

- MOTO configuration
- Card installments configuration
- Recurring payment configuration


raoulritter (approved):

Block/Adminhtml/System/Config/Field/Installment.php
```diff
@@ -11,48 +11,55 @@
...
-     *
-     * @param string $value
-     * @return $this
-     */
-    public function setInputName($value)
+    public function _toHtml(): string
+    {
+        if (!$this->getOptions()) {
+            $this->setOptions($this->getSourceOptions());
+        }
```
hossam-adyen, line 52: Can we move that to the class constructor?

Block/Adminhtml/System/Config/Field/RecurringProcessingModel.php
```diff
@@ -16,25 +16,30 @@
...
+            ['label' => 'Unscheduled Card on File', 'value' => Vault::UNSCHEDULED_CARD_ON_FILE],
+            ['label' => 'Subscription', 'value' => Vault::SUBSCRIPTION]
+        ];
+    }
+
+    public function _toHtml(): string
+    {
+        if (!$this->getOptions()) {
+            $this->setOptions($this->getSourceOptions());
+        }
```
hossam-adyen, line 32: Can we move that to the class constructor?

Block/Adminhtml/System/Config/Field/SelectYesNo.php
```diff
@@ -15,24 +15,29 @@
...
+            ['label' => 'No', 'value' => '0'],
+            ['label' => 'Yes', 'value' => '1'],
+        ];
+    }
+
+    public function _toHtml(): string
+    {
+        if (!$this->getOptions()) {
+            $this->setOptions($this->getSourceOptions());
+        }
```
hossam-adyen, line 30: same :)

view/adminhtml/templates/config/token_type_table_array.phtml
```diff
@@ -42,7 +42,7 @@ $_colspan = $block->isAddAfter() ? 2 : 1;
 
                 // define row prototypeJS template
                 template: mageTemplate(
-                    '<tr id="can">'
+                    '<tr id="<%- _id %>">'
```
hossam-adyen, line 45: I liked the old naming LoL

Block/Adminhtml/System/Config/Field/SelectYesNo.php
candemiralp: I would rather keep it here because that class is an extension of `Magento\Framework\View\Element\Html` and this method will be called specifically. What do you think?

Block/Adminhtml/System/Config/Field/RecurringProcessingModel.php
candemiralp: I would rather keep it here because that class is an extension of `Magento\Framework\View\Element\Html` and this method will be called specifically. What do you think?

Block/Adminhtml/System/Config/Field/Installment.php
candemiralp: I would rather keep it here because that class is an extension of `Magento\Framework\View\Element\Html` and this method will be called specifically. What do you think?

peterojo: Magento instantiates this block internally, and this function is called at the point of rendering the template so having the options set here give us a bit more control over how and when this is run

khushboo-singhvi (approved):

Block/Adminhtml/System/Config/Field/RecurringProcessingModel.php
hossam-adyen: I would favor move this to constructor for two reasons:

1. this method seems to be called many times, we can avoid the if condition
2. the options seems not changing during the execution, it's more sense to set it once in the constructor that we know it will be called once

candemiralp: After a small discussion with @hossam-adyen, we decided to leave it as it is in order to follow Magento's extensibility best practices. An example implementation can be found in their [documentation](https://developer.adobe.com/commerce/php/tutorials/admin/create-dynamic-row-configuration).

PR #2154, opened 2023-08-09
Adding missing Exception type and small improvements to configuration wizard text

acampos1916:
**Description**
If the server fails to connect to Adyen during the automatic configuration then `ConnectionException` is thrown, and it wasn't being caught before.

Small adjustments to the configuration wizard text.

**Tested scenarios**
Automatic configuration


khushboo-singhvi (dismissed):

Controller/Adminhtml/Configuration/MerchantAccounts.php
```diff
@@ -71,7 +72,7 @@ public function execute()
 
             $resultJson->setData($response);
             return $resultJson;
-        } catch (AdyenException $e) {
+        } catch (AdyenException | ConnectionException $e) {
```
hossam-adyen, line 75: `ConnectionException` is thrown when something is wrong on the connection to Adyen so in this case it should not return 4XX response status code as it's not a client error, it should return a 5XX in this case

It would be more convenient to catch it in another catch statment

```php
     } catch (ConnectionException $e) {
            $resultJson->setHttpResponseCode(400);
            ...
```

hossam-adyen (changes requested):
thanks for your contribution 👍 , small change needed

hossam-adyen (approved):

candemiralp (approved):

PR #2155, opened 2023-08-09
[ECP-8521] Move from PHP classes to facades for alternative payment methods

candemiralp:
**Description**

In order to implement individual payment methods, we created PHP classes for the given payment method. This PHP class extends Adapter and can be used a payment method object. However, having PHP classes instead of XML entries in di.xml creates an extra complexity for payment method related issues. In addition, we need to create lots of PHP files to support that payment method. If that payment method supports recurring, another file needs to be created for vault payments.

With this PR, we migrated from PHP classes to payment method facades as a Magento best practice.

**Tested scenarios**

- Payment flow
- Payment modifications
- Recurring payments
- Webhook handling

### DIFF
--- a/Helper/Config.php
+++ b/Helper/Config.php
@@ -50,6 +50,7 @@ class Config
     const XML_CONFIGURATION_MODE = 'configuration_mode';
     const XML_ADYEN_POS_CLOUD = 'adyen_pos_cloud';
     const XML_WEBHOOK_NOTIFICATION_PROCESSOR = 'webhook_notification_processor';
+    const AUTO_CAPTURE_OPENINVOICE = 'auto';
     const XML_RECURRING_CONFIGURATION = 'recurring_configuration';
 
     protected ScopeConfigInterface $scopeConfig;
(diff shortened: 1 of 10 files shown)


Model/Ui/AdyenGenericConfigProvider.php
```diff
@@ -23,6 +23,32 @@ class AdyenGenericConfigProvider implements ConfigProviderInterface
...
+    const APPLEPAY_TX_VARIANT = 'applepay';
+    const BCMC_MOBILE_TX_VARIANT = 'bcmc_mobile';
+    const DOTPAY_TX_VARIANT = 'dotpay';
+    const FACILYPAY_3X_TX_VARIANT = 'facilypay_3x';
+    const MULTIBANCO_TX_VARIANT = 'multibanco';
+    const GOOGLEPAY_TX_VARIANT = 'googlepay';
+    const KLARNA_TX_VARIANT = 'klarna';
+    const PAYPAL_TX_VARIANT = 'paypal';
+    const SEPADIRECTDEBIT_TX_VARIANT = 'sepadirectdebit';
+
```
hossam-adyen, line 51: can we got those constant from Adyen php lib?

hossam-adyen (approved):

Model/Ui/AdyenGenericConfigProvider.php
candemiralp: Hello @hossam-adyen, thank you for your review. Currently, it is not possible to obtain this information from PHP API Library. `adyen_XXXX` notation is specific to Magento. Obtaining tx_variants from `paymentMethods` call is possible, but they need to be statically defined in `di.xml` file.

view/frontend/web/js/view/payment/method-renderer/adyen-paypal-method.js
```diff
@@ -32,6 +32,8 @@ define(
                 return paypalConfiguration
             },
             renderActionComponent: function(resultCode, action, component) {
+                debugger;
```
hossam-adyen, line 35: to be removed!

hossam-adyen (dismissed):

khushboo-singhvi (approved):

view/frontend/web/js/view/payment/method-renderer/adyen-paypal-method.js
candemiralp: Thanks for the heads-up!

hossam-adyen (approved):

khushboo-singhvi (approved):

PR #2212, opened 2023-09-12
[ECP-8601] Refund for partial payments fail

khushboo-singhvi:
**Description**
Refund for giftcard partial payments fail. The fix was made to have Merchant Account on all the partial refund array items

**Tested scenarios**
- Refund tested for partial payments(Giftcard + CC)

Fixes


hossam-adyen (commented):
lets add some unit tests

hossam-adyen (approved):

PR #2243, opened 2023-09-27
[ECP-8625] Fix Apple Pay download certificate button URL

khushboo-singhvi:
**Description**
Apple Pay certificate download button now redirects to an existing URL. URL is being fetched from the code base, instead of getting it from the patch

**Tested scenarios**

Fixes


peterojo (approved):

Setup/Patch/Data/ApplePayCertificateUrlPath.php
candemiralp, line 1: This file needs to be removed. Config table is not being used anymore to fetch the URL.

candemiralp (changes requested):

candemiralp (approved):

PR #2266, opened 2023-10-10
[ECP-8591] Anomaly in Payment Method Transition: Persistence of cc_type Value Despite Payment Method Change

khushboo-singhvi:
**Description**
After canceling a 3DS CC payment and choosing any other giftcard, the cc_type field has retained the value. This PR unsets it for all other payment methods.

Unset the cc_type value for the related observers
**Tested scenarios**

Tried changing the payment method and checking the respective fields in DB, it holds correct value as expected.

Fixes

### DIFF
--- a/Observer/AdyenBoletoDataAssignObserver.php
+++ b/Observer/AdyenBoletoDataAssignObserver.php
@@ -11,6 +11,7 @@
 
 namespace Adyen\Payment\Observer;
 
+use Adyen\Payment\Helper\Data;
 use Magento\Framework\Event\Observer;
 use Magento\Payment\Observer\AbstractDataAssignObserver;
 use Magento\Quote\Api\Data\PaymentInterface;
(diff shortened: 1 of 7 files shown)


Observer/AdyenBoletoDataAssignObserver.php
```diff
@@ -66,6 +67,9 @@ public function execute(Observer $observer)
         // Remove remaining brand_code information from the previous payment
         $paymentInfo->unsAdditionalInformation('brand_code');
 
+        // Remove cc_type information from the previous payment
+        $paymentInfo->unsAdditionalInformation('cc_type');
```
candemiralp, line 71: `cc_type` is not an additional information. It's field in the BD table. It can't be unset with the given command.

candemiralp (approved):

hossam-adyen (approved):

peterojo (approved):

PR #2268, opened 2023-10-10
[ECP- 8591] Anomaly in Payment Method Transition: Persistence of cc_type Value Despite Payment Method Change

khushboo-singhvi:
**Description**
After canceling a 3DS CC payment and choosing any other giftcard, the cc_type field has retained the value. This PR unsets it for all other payment methods.

Unset the cc_type value for the related observers

**Tested scenarios**
Tried changing the payment method and checking the respective fields in DB, it holds correct value as expected.

Fixes

### DIFF
--- a/Observer/AdyenBoletoDataAssignObserver.php
+++ b/Observer/AdyenBoletoDataAssignObserver.php
@@ -3,7 +3,7 @@
  *
  * Adyen Payment module (https://www.adyen.com/)
  *
- * Copyright (c) 2015 Adyen BV (https://www.adyen.com/)
+ * Copyright (c) 2023 Adyen BV (https://www.adyen.com/)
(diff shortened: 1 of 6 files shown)


candemiralp (dismissed):

Observer/AdyenBoletoDataAssignObserver.php
```diff
@@ -33,17 +34,17 @@ class AdyenBoletoDataAssignObserver extends AbstractDataAssignObserver
     ];
 
     /**
-     * @var \Adyen\Payment\Helper\Config
+     * @var Config
      */
     private $configHelper;
```
hossam-adyen, line 39:
```suggestion
    private Config $configHelper;
```

candemiralp (approved):

PR #2269, opened 2023-10-10
Harden webhook processing against exceptions during invoice mail sending

### LINKED ISSUE #2022: Order status stuck in "Payment review" when it actually has completed successfully
### **Describe the bug**
We are experiencing similar issue to the ones reported in #744, #782, #887, #963 - unfortunately, there was no solution or fix posted there.

Since mid March when we upgraded to **Magento 2.4.3-p3** and **Adyen 7.3.13** we have had a number of cases where orders with successful credit card transactions were not transitioned from _Payment review_ status to _Processing_ status. This issue still randomly occurs - there is no one scenario for this issue, it happens for Mastercard and Visa cards, different countries etc.

In the Adyen admin panel payment (of the affected order) is listed as successful. Magento order details page (see screenshot below) shows that payment was successfully authorized and captured and status was changed to processing - but in reality status was not changed. Status of the order in the orders grid or the status displayed at the top of the order detail page is still _Payment review_. The order is then not possible to process correctly and affects other automations that depend on the orders status.

### **To Reproduce**
No exact scenario on how to reproduce this issue, occurs randomly.

### **Expected behavior**
Order status should be changed from **Payment review** to **Processing** when payment was successfully captured.

### **Screenshots**
image

### **Additional context**
Started to occur after migrating to Magento 2.4.3-p3 and Adyen 7.3.13 - did not happens on previous (older) Adyen module versions.

mfickers:
**Description**
If an exception is thrown while the invoice mail is sent when an AUTHORISATION webhook notification is being processed, the order status will never be updated.

The cron will first process the webhook, then send the invoice mail and lastly save the updated order. Exceptions during sending of the invoice mail will interrupt this flow. The webhook will be marked as successfully processed, but the order stays in status "Payment review".

This change applies the same try-catch logic to the invoice sender that is already present in the order confirmation sender:
\Adyen\Payment\Helper\Order::sendOrderMail

**Tested scenarios**
\Adyen\Payment\Test\Unit\Helper\InvoiceTest::testSendInvoiceMailCatchesException makes sure that exceptions thrown inside of \Magento\Sales\Model\Order\Email\Sender\InvoiceSender::send are caught inside \Adyen\Payment\Helper\Invoice::sendInvoiceMail.

Fixes  #2022

### DIFF
--- a/Helper/Invoice.php
+++ b/Helper/Invoice.php
@@ -213,7 +213,7 @@ class Invoice extends AbstractHelper
             );
 
             if ($invoiceAutoMail) {
-                $this->invoiceSender->send($invoice);
+                $this->sendInvoiceMail($invoice);
             }
 
             return $invoice;
@@ -437,7 +437,7 @@ class Invoice extends AbstractHelper
         );
         $transactionSave->save();
 
-        $this->invoiceSender->send($invoice);
+        $this->sendInvoiceMail($invoice);
 
         //Send Invoice mail to customer
         $invoiceAutoMail = (bool)$this->scopeConfig->isSetFlag(
@@ -447,7 +447,7 @@ class Invoice extends AbstractHelper
         );
 
         if ($invoiceAutoMail) {
-            $this->invoiceSender->send($invoice);
+            $this->sendInvoiceMail($invoice);
             $order->addStatusHistoryComment(
                 __('Notified customer about invoice creation #%1.', $invoice->getId())
             );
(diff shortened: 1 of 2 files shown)


hossam-adyen (approved):
Hi @mfickers , thanks for your PR

candemiralp (approved):

candemiralp:
Hello @mfickers,

Thank you for your contribution. This PR has been merged!

Best Regards,
Can

PR #2302, opened 2023-10-24
[ECP-8675v8] Undefined array key "pspreference" in TransactionCapture 

RokPopov:
**Description**
If capture response doesn't contain pspreference, catch block fails [here](https://github.com/Adyen/adyen-magento2/blob/3e9dc10be04b31fdb97d8661b29edc33396c5562/Gateway/Http/Client/TransactionCapture.php#L131C30-L131C51). This PR checks if the PSP reference is set and handles the exception message accordingly if it isn't.

**Tested scenarios**
- process the capture request without the PSP reference

PR #2304, opened 2023-10-24
[ECP-8675v9] Undefined array key "pspreference" in TransactionCapture

RokPopov:
**Description**
If capture response doesn't contain pspreference, catch block fails [here](https://github.com/Adyen/adyen-magento2/blob/3e9dc10be04b31fdb97d8661b29edc33396c5562/Gateway/Http/Client/TransactionCapture.php#L131C30-L131C51). This PR checks if the PSP reference is set and handles the exception message accordingly if it isn't.

**Tested scenarios**
- process the capture request without the PSP reference

### DIFF
--- a/Gateway/Http/Client/TransactionCapture.php
+++ b/Gateway/Http/Client/TransactionCapture.php
@@ -105,10 +105,14 @@ class TransactionCapture implements ClientInterface
                 $singleResponse = $this->copyParamsToResponse($singleResponse, $request);
                 $response[self::MULTIPLE_AUTHORIZATIONS][] = $singleResponse;
             } catch (AdyenException $e) {
+                $pspReference = isset($request[OrderPaymentInterface::PSPREFRENCE]) ?
+                    $request[OrderPaymentInterface::PSPREFRENCE] :
+                    'pspReference not set';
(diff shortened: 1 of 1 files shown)


candemiralp (approved):

hossam-adyen (approved):

dufrygrant:
Can this please be added into the 'stable' 8.x branch, as this is where it was originally reported and where there are PROD installations being unable to capture as a result?

@aziya-im From our perspective it's more important to have this in the versions typically installed in PROD *today* than a branch that isn't widely deployed yet (9.x).  This was word-for-word the exact same problem (prioritisation of merge into a new branch over the widely deployed one) we reported via Sven in October 2021 and despite receiving assurances at the time that such reported issues would be treated a certain way, this is apparently not happening. 8.22.2 is now two weeks old. C'mon!

RokPopov:
Hi @dufrygrant,

Thank you for your feedback, this is much appreciated. However, please know that we also created a PR for v8 which will be released this week. You can check it out [here](https://github.com/Adyen/adyen-magento2/pull/2302).

Cheers,
Rok

PR #2316, opened 2023-10-30
[ECP-8712] Fix broken partial captures and implement adyen_authorized status

### LINKED ISSUE #2097: [PW-8506] Wrong Order status when configuration field 'payment_pre_authorized' is set
**Describe the bug**

Why is the method \Adyen\Payment\Helper\Order::setPrePaymentAuthorized called only when we do have a successful authorization there \Adyen\Payment\Helper\Webhook\AuthorisationWebhookHandler::handleSuccessfulAuthorisation ?
Shouldn't we consider authorized payment as processed?

[See Magento Doc Order Flow](https://docs.magento.com/user-guide/sales/order-workflow.html)

> Step 3: Receive Payment
> The status of the order changes to Processing when payment is received or authorized. Depending on the payment method, you might receive notification when the transaction is authorized or processed.

Today there's the following configuration fields that allows us to set the status by event:
- payment_pre_authorized
- payment_authorized

You may have something like this: 

```xml
<?xml version="1.0"?>
<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Store:etc/config.xsd">
    <default>
        <payment>
            <adyen_abstract>
                <payment_pre_authorized>pending_payment</payment_pre_authorized>
                <payment_authorized>processing</payment_authorized>
                <payment_cancelled>canceled</payment_cancelled>
            </adyen_abstract>
        </payment>
    </default>
</config>
```

So when a payment is pending authorization it is in pending_payment (so it can be cancelled automatically by Magento after a delay). When the payment is authorized we set the processing status.
However, we can see that the field 'payment_pre_authorized' is used when a payment is authorized and not before we got the notification. 'payment_authorized' is used when there's the auto-capture or the capture. So the fields are misleading and not clear. Obviously I tend to think there's an error and that \Adyen\Payment\Helper\Webhook\AuthorisationWebhookHandler::handleSuccessfulAuthorisation should at its beginning set the status from this field: payment_authorized (when $isFullAmountAuthorized is true)

**To Reproduce**
1. Enable manual capture
2. Set the configuration as above
3. Place an order
4. Order is in payment review, then pending payment
5. Order receive the successful AUTHORIZATION event
6. Don't capture the payment, the order is going to be cancelled by Magento (default value is 480 minutes)

**Expected behavior**
After the successful AUTHORIZATION event, the order is in processing state, so it can't be canceled by the Magento cron.

**Magento version**
Any

**Plugin version**
8.x (8.18.2)

**Additional context**

candemiralp:
**Description**

Invoice creation after processing `AUTHORISATION` webhook was disabled and order state/statuses was revisited. All the invoices now will be created after capture command.

To distinguish the orders authorised but not captured from the orders have no payment, a new status `Authorized` was created and was binded to state `new`. This status can be selected from the `Order management / Order status / Payment authorisation `.

**Tested scenarios**

- Manual capture state checks

Fixes #2097


hossam-adyen (dismissed):

RokPopov (approved):

khushboo-singhvi (approved):

etc/config.xml
```diff
@@ -29,7 +29,7 @@
                 <house_number_street_line>0</house_number_street_line>
                 <webhook_notification_processor>cron</webhook_notification_processor>
                 <adyen_support_email_address>support@adyen.com</adyen_support_email_address>
-                <payment_pre_authorized>processing</payment_pre_authorized>
+                <payment_pre_authorized>adyen_authorized</payment_pre_authorized>
```
hossam-adyen, line 32: Do we need migration for that change?

hossam-adyen (approved):

etc/config.xml
candemiralp: We have a data patch in the PR for the merchants who have already upgraded to 9.0.2 to revert the usage of `processing` state. If the merchant doesn't have `processing` state set in the configuration, we don't need to patch the data assuming that they have the correct configuration for `payment_pre_authorized` field.

PR #2318, opened 2023-10-31
[V8][ECP 8711] Optimize cancel expired pay by link orders cron 

hossam-adyen:
**Description**

Optimize cancel expired pay by link orders cron 

Fixes  #2202

PR #2313, opened 2023-10-27
[ECP-8626] Fix cron paybylink cancle memory issue

hossam-adyen:
**Description**
Fix https://github.com/Adyen/adyen-magento2/issues/2202 Fix cron paybylink cancel memory issue

**Tested scenarios**

Fixes

PR #2325, opened 2023-11-06
[ECP-8727] Fix the condition to fetch expired PBL orders

candemiralp:
**Description**

A bug was found on order cancellation cron designed for expired Adyen Pay by Link payment method after implementing #2313 and #2318. This bug causes cancellation of all orders in `new` state.

This PR fixes the condition and provides only expired Adyen Pay by Link orders to `CancelOrders` cron job.

**Tested scenarios**

- Cancelling only expired Adyen Pay by Link payments

### DIFF
--- a/Cron/Providers/PayByLinkExpiredPaymentOrdersProvider.php
+++ b/Cron/Providers/PayByLinkExpiredPaymentOrdersProvider.php
@@ -125,15 +125,16 @@ class PayByLinkExpiredPaymentOrdersProvider implements OrdersProviderInterface
             ->setValue(Order::STATE_NEW)
             ->create();
 
-        $orderIdFilter = $this->filterBuilder->setField('state')
+        $orderIdFilter = $this->filterBuilder->setField('entity_id')
             ->setConditionType('in')
             ->setValue($expiredOrderIds)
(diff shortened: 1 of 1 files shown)


raoulritter (approved):

Cron/Providers/PayByLinkExpiredPaymentOrdersProvider.php
```diff
@@ -125,15 +125,16 @@ protected function getExpiredOrders($expiredOrderIds): array
             ->setValue(Order::STATE_NEW)
             ->create();
 
-        $orderIdFilter = $this->filterBuilder->setField('state')
+        $orderIdFilter = $this->filterBuilder->setField('entity_id')
             ->setConditionType('in')
             ->setValue($expiredOrderIds)
             ->create();
 
-        $filterGroup = $this->filterGroupBuilder->setFilters([$stateFilter, $orderIdFilter])->create();
```
hossam-adyen, line 133: Why this line were causing the issue?

`In Magento 2 all the filters in same FilterGroup will be added using OR operator. But all the FilterGroup will be added using AND operator.`

PR #2382, opened 2023-12-05
Do not launch customer/section/load after adyen payment retrieve action

Marwaa92:
**Description**

**Tested scenarios**

Fixes

### DIFF
--- a/view/frontend/web/js/model/adyen-payment-service.js
+++ b/view/frontend/web/js/model/adyen-payment-service.js
@@ -52,7 +52,8 @@ define(
 
                 return storage.post(
                     serviceUrl,
-                    JSON.stringify(payload)
+                    JSON.stringify(payload),
+                    false
                 );
             },
 


RokPopov:
Hi @Marwaa92 

Thank you for your contribution! Would you mind giving a bit more details on this in the description? 
For example what is the issue this PR is addressing, why do you see this change to be neccessary?

Thank you in advance,
Rok

Marwaa92:
hello @RokPopov ,

This is an optimisation to prevent customer/section/load? Sections=messages to be launched as time as we launch adyen payment retrieve action.
In some contexts we are facing some problem with redis lock, when a big number of HTTPS request is launched on the checkout page.

So I propose to disable the option on the Ajax adyen payment retrieve action to reduce the number of request server on the checkout page.

best regards,
Marwa

candemiralp:
Hello @Marwaa92,

Thank you for the clarification on your PR. However, on V9 of the plugin after implementing #2219 and #2180, usage of `retrievePaymentMethods` JS function was deprecated and this will be removed soon. Your PR might not create a difference since this endpoint is not being called anymore.

Could you please check the validity of the changes once again?

Best Regards,
Can

candemiralp:
Hello @Marwaa92,

This fix seems irrelevant and I am closing this PR since we didn't get a response. Please inform us if you think this PR is valid after the previous explanation.

Thank you for your contribution.

Best Regards,
Can

PR #2385, opened 2023-12-05
[ECP-8799] [V9] Fix refund openinvoice after cart expired

### LINKED ISSUE #2367: ECP-8799 [V9] Cannot refund (via Magento) with Alternative methods
**Describe the bug**

We cannot refund online via creditmemo creation page (only button `Refund offline` is present) for alternative payments (tested on alma and google pay.).

It works fine for adyen_cc.

**To Reproduce**
1. Try to refund an order paid with googlepay

**Expected behavior**

`Refund online` button is there.

**Magento version**
2.4.6-p3

**Plugin version**
9.0.2

hossam-adyen:
**Description**
Fix refund & capture an openInvoice order when the quote expired

**Tested scenarios**
Place openInvoice order and refund it successfully when the related quote expired

Fixes  #2367

### DIFF
--- a/Gateway/Request/RefundDataBuilder.php
+++ b/Gateway/Request/RefundDataBuilder.php
@@ -18,11 +18,8 @@ use Adyen\Payment\Helper\OpenInvoice;
 use Adyen\Payment\Model\ResourceModel\Invoice\CollectionFactory;
 use Adyen\Payment\Model\ResourceModel\Order\Payment\CollectionFactory as PaymentCollectionFactory;
 use Adyen\Payment\Observer\AdyenPaymentMethodDataAssignObserver;
-use Magento\Payment\Gateway\Data\PaymentDataObject;
 use Magento\Payment\Gateway\Helper\SubjectReader;
 use Magento\Payment\Gateway\Request\BuilderInterface;
-use Magento\Payment\Model\InfoInterface;
-use Magento\Sales\Model\Order\Creditmemo;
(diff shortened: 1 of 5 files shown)


Gateway/Request/RefundDataBuilder.php
```diff
@@ -168,7 +168,7 @@ public function build(array $buildSubject): array
             );
 
             if ($this->adyenHelper->isPaymentMethodOpenInvoiceMethod($brandCode)) {
-                $openInvoiceFieldsCreditMemo = $this->openInvoiceHelper->getOpenInvoiceData($creditMemo->getOrder());
+                $openInvoiceFieldsCreditMemo = $this->openInvoiceHelper->getOpenInvoiceDataFromPayment($payment);
```
candemiralp, line 171: What would you think about using `getCreditmemosCollection()` for refunds? That allows us to get correct items. Last refunded items might be different than last invoiced items.

hossam-adyen: I've inject the credit memo invoice, to be use instead of the last invoice

```
$openInvoiceFieldsCreditMemo = $this->openInvoiceHelper->getOpenInvoiceDataFromPayment($payment, $creditMemo->getInvoice());
```

RokPopov (approved):

candemiralp (dismissed):

candemiralp (approved):

khushboo-singhvi (approved):

PR #2400, opened 2023-12-18
Remove cart dependency when creating open invoice

FrankStappers:
**Description**
Remove dependencies on product and quote when creating an open invoice. The shared functionality is used when converting a quote to an onder, but also when creating a creditmemo from an invoice. When either a quote or a product has been deleted of an invoice that needs a refund, the original code fails. This fix resolves the failures when these dependencies are no longer met.

**Tested scenarios**
Create an invoice for an open invoice payment method (e.g.Klarna). Remove the quote corresponding to the invoice. Creditmemo the invoice.

### DIFF
--- a/Helper/ChargedCurrency.php
+++ b/Helper/ChargedCurrency.php
@@ -109,6 +109,32 @@ class ChargedCurrency
         );
     }
 
+    public function getOrderItemAmountCurrency(\Magento\Sales\Model\Order\Item $item)
+    {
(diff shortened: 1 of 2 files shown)


hossam-adyen:
Hi Frank, thanks for opening this PR, but we already fixed this issue with this pr https://github.com/Adyen/adyen-magento2/pull/2385

As this one no longer needed, I will close it

PR #2432, opened 2024-01-10
Make `Payment methods enabled` a global setting

aschrammel:
**Description**
The configuration setting `Payment methods enabled` is shown across all scopes. No matter what scope is selected, the underlying backend model `Adyen\Payment\Model\Config\Backend\PaymentMethodsStatus` is always updating the default scope for the single payment methods.
This leads to inconsistencies between `payment/adyen_abstract/payment_methods_active` and all the single payment method configurations.

* Set `Payment methods enabled` to `Yes` in `Default` scope
* Having a `Website A` and a `Website B` in place
* Set `Payment methods enabled`to `Yes` in `Website A` scope

**Expected Result**
Payment methods are still active in `Website B`
**Actual Result**
Payment methods are deactived in `Website B` as well

Showing the configuration setting only in `Default` scope won't lead to such inconsistencies.

**Tested scenarios**
`Stores > Configuration > Sales > Adyen > Accepting Payments > Payment methods > Payment methods enabled` is a global setting

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
   "name": "adyen/module-payment",
   "description": "Official Magento2 Plugin to connect to Payment Service Provider Adyen.",
   "type": "magento2-module",
-  "version": "9.0.5",
+  "version": "9.0.6",
   "license": "MIT",
   "repositories": [
     {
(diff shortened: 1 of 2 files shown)


RokPopov:
Hi @aschrammel,

Thank you for your efforts and contribution to improving our plugin. We appreciate the time and thought you put into your PR regarding making the configuration global.

After careful consideration, we've decided to proceed with a different approach that implements scopable configurations for payment methods. You can take a look at the PR we're moving forward with [here](https://github.com/Adyen/adyen-magento2/pull/2587).

We value your contributions and encourage you to continue sharing your ideas and improvements with us.

Kind regards,
Rok

PR #2458, opened 2024-01-23
Remove post install commands for phpcompatibility/php-compatibility

candemiralp:
**Description**

- Remove `post-install-cmd` and `post-update-cmd` commands from composer.json as `phpcompatibility/php-compatibility` is not a dependency of Magento anymore.
- Fix codesniffing errors.

### DIFF
--- a/view/adminhtml/templates/config/applepay_button.phtml
+++ b/view/adminhtml/templates/config/applepay_button.phtml
@@ -1,3 +1,3 @@
 <?php /** @var \Adyen\Payment\Model\Config\Adminhtml\ApplepayCertificateButton $block */ ?>
 
-<input type="button" onclick="location.href='<?= $block->getActionUrl() ?>';" value="Download certificate" />
+<input type="button" onclick="location.href='<?= $block->escapeJs($block->getActionUrl()) ?>';" value="Download certificate" />
--- a/composer.json
+++ b/composer.json
@@ -41,12 +41,6 @@
     "test": [
       "Composer\\Config::disableProcessTimeout",
       "vendor/bin/phpunit -c Test/phpunit.xml"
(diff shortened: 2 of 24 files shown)


RokPopov (approved):

view/adminhtml/templates/config/applepay_button.phtml
```diff
@@ -1,3 +1,3 @@
 <?php /** @var \Adyen\Payment\Model\Config\Adminhtml\ApplepayCertificateButton $block */ ?>
 
-<input type="button" onclick="location.href='<?= $block->getActionUrl() ?>';" value="Download certificate" />
+<input type="button" onclick="location.href='<?= $block->escapeJs($block->getActionUrl()) ?>';" value="Download certificate" />
```
hossam-adyen, line 3:
```suggestion
<input type="button" onclick="location.href='<?= $block-> escapeUrl($block->getActionUrl()) ?>';" value="Download certificate" />
```

candemiralp: `onclick` attribute is a part of the JS code although it's a URL. According to the documentation, `escapeUrl()` can be safely used on `href` and `src` html attributes. But, any code which is part of a JS script needs to be escaped with `escapeJs()` function. Below example was taken from the documentation.

```
<script>
  let phrase = "Hi, my name is <?= $escaper->escapeJs($myName) ?>";
  //Do not use HTMl context methods like escapeUrl
  let redirectUrl = "<?= $escaper->escapeJs($myUrl) ?>";
  location.href = redirectUrl;
</script>
```

[Reference](https://developer.adobe.com/commerce/php/development/security/cross-site-scripting/)

PR #2479, opened 2024-02-05
[ECP-8861] Fix cancel open offer

### LINKED ISSUE #2404: Orders with failed authorisation can't be canceled
**Describe the bug**
Since we upgraded the module to the latest 8 version, orders with failed authorisation have remained stuck in payment review and are not being canceled when processing the Adyen notification. There is a generic message "The order failed to update: Transaction has been declined. Please try again later.", that I think I've been able to track to the changes introduced in this PR: https://github.com/Adyen/adyen-magento2/pull/2227/files

**To Reproduce**
1. Place an order and fail the 3d secure authentication
2. Order is present in Magento backend, with corresponding Magento payment info
3. Error shows up in the Adyen error log: "Error with cancellation"
4. A notification is received from Adyen
```
Adyen HTTP Notification(s):
eventCode: AUTHORISATION
pspReference: %PSP reference goes here%
paymentMethod: mc
success: false
reason:3D Not Authenticated
```
but the Magento order is not canceled, it's just updated with a comment saying: "The order failed to update: Transaction has been declined. Please try again later."

**Expected behavior**
Magento order should get canceled.

**Magento version**
5.4.6-p3

**Plugin version**
8.22.5

**Additional context**
I've tracked this issue and it seems to be due to the changes introduced in this PR: https://github.com/Adyen/adyen-magento2/pull/2227/files
After that was added, the PSP references used to build the cancel request have started to be determined based on `adyen_order_payment` table entries:
https://github.com/Adyen/adyen-magento2/pull/2227/files#diff-51b8cf2ebb2e0050bf1d10b97464bac8a567c6ac9cf148bb0a7a24b89432a3d9R56
```
Gateway/Request/CancelDataBuilder.php

        $pspReferences = $this->adyenPaymentResourceModel->getLinkedAdyenOrderPayments(
            $payment->getEntityId()
        );
```
But the issue is that with orders failing 3ds authentication, there is no entry in the `adyen_order_payment` table and this used to work because the PSP reference was fetched from the Magento payment:
https://github.com/Adyen/adyen-magento2/pull/2227/files#diff-51b8cf2ebb2e0050bf1d10b97464bac8a567c6ac9cf148bb0a7a24b89432a3d9L35
```
Gateway/Request/CancelDataBuilder.php

$pspReference = $payment->getCcTransId();
```

hossam-adyen:
**Description**

Handle the case when we build cancellation request for an open offer that doesn't have linked payments

Fixes  #2404

### DIFF
--- a/Gateway/Request/CancelDataBuilder.php
+++ b/Gateway/Request/CancelDataBuilder.php
@@ -63,6 +63,13 @@ class CancelDataBuilder implements BuilderInterface
             $payment->getEntityId()
         );
 
+        //Handle the case when we cancel an open offer
+        if (empty($pspReferences)) {
+            $pspReferences = [
+                ['pspreference' => $payment->getCcTransId()]
+            ];
+        }
+
         $requests['body'] = [];
         foreach ($pspReferences as $pspReference) {
             $request = [
@@ -70,7 +77,6 @@ class CancelDataBuilder implements BuilderInterface
                 "reference" => $order->getOrderIncrementId(),
                 "merchantAccount" => $merchantAccount
             ];
-
             $requests['body'][] = $request;
         }
         $requests['clientConfig'] = ["storeId" => $order->getStoreId()];


Gateway/Request/CancelDataBuilder.php
```diff
@@ -63,14 +63,20 @@ public function build(array $buildSubject): array
             $payment->getEntityId()
         );
 
+        //Handle the case when we cancel an open offer
+        if (empty($pspReferences)) {
```
candemiralp, line 67: This implementation doesn't cover partial payments. Let's discuss the approach again.

hossam-adyen:
After consideration, this is not the optimal way to solve this issue as the failed authorized payments should not call the cancellation command, we are closing this pr and refining more rubout solution for the issue.

Issue #2307, opened 2023-10-25
[ECP-8717][ECP-8955] Call to undefined method getQuoteId() in "Adyen\Payment\Gateway\Request\GiftcardDataBuilder"


AlexGorelikov:
Hey

**Describe the bug** 
Runtime exception when you want to place an order. The PayPal module is a dependent module for the Magento. We have a conflict of references

**Adyen\Payment\Gateway\Request\GiftcardDataBuilder (38line)**
$paymentDataObject->getOrder(); // PayPal\Braintree\Gateway\Data\Order\OrderAdapter

vendor/paypal/module-braintree-core/etc/di.xml 
`<preference for="Magento\Payment\Gateway\Data\Order\OrderAdapter" type="PayPal\Braintree\Gateway\Data\Order\OrderAdapter"/>`

vendor/adyen/module-payment/etc/di.xml
`<preference for="Magento\Payment\Gateway\Data\Order\OrderAdapter" type="Adyen\Payment\Gateway\Data\Order\OrderAdapter" />`

**To Reproduce:** 
Place an order

**Expected behavior**
Success order

Plugin version
9.0.1

candemiralp:
Hello @AlexGorelikov,

Thank you for creating this issue. We created an internal ticket to investigate an solve this issue.

Best Regards,
Can

aschrammel:
Hi @candemiralp 

are there any news on that issue? We are currently in process of upgrading from `8.20.1` to `9.0.5` and running into the exact same issue.

Would be great to get some info on the progress there 😄

Thx a lot in advance.

Best regards
Andreas

candemiralp:
Hello @aschrammel,

This issue is currently on our sprint board and planning to handle it on the current sprint (within 2 weeks).

Best Regards,
Can

aschrammel:
@candemiralp I've updated to 9.0.7 now and still face the same issue. Now it's working on the corrected place, but in the end I'm still running into the error 

`Call to undefined method PayPal\Braintree\Gateway\Data\Order\OrderAdapter::getQuoteId()`

Stack trace is the following:
```#0 /srv/vendor/magento/module-payment/Gateway/Response/HandlerChain.php(51): Adyen\Payment\Gateway\Response\StateDataHandler->handle()
#1 /srv/vendor/magento/module-payment/Gateway/Command/GatewayCommand.php(118): Magento\Payment\Gateway\Response\HandlerChain->handle()
[...]

PR #2481, opened 2024-02-05
Fix undefined method getQuoteId() exception when trying to check out

mikimpe:
**Description**

Change OrderAdapter preference target from Magento\Payment\Gateway\Data\Order\OrderAdapter to PayPal\Braintree\Gateway\Data\Order\OrderAdapter

`Adyen_Payment` creates a preference on `Magento\Payment\Gateway\Data\Order\OrderAdapter`. However, module `PayPal_Braintree` also creates a preference on the same class. 
Since `Adyen_Payment` does not define any dependency on `PayPal_Braintree`, it may happen that in the `config.php` file `PayPal_Braintree` is listed after `Adyen_Payment`. This causes the preference declared by `Adyen_Payment` to have no effect, and instead, the preference of `PayPal_Braintree` is taken into account. 
If this occurs, an exception is raised during the checkout:
`Call to undefined method PayPal\Braintree\Gateway\Data\Order\OrderAdapter::getQuoteId()`
To fix this, I ensured that the `Adyen_Payment` preference targets the rewrite class of `PayPal_Braintree`. Additionally, I have added the dependency in the `module.xml` file. 
By doing this, we ensure that the `Adyen_Payment` preference takes effect and avoids the exception during checkout. 
Moreover, by targeting `PayPal_Braintree` rewrite class in the preference and extending it, we also carry along the changes made by `PayPal_Braintree` to the original class. 
Since `PayPal_Braintree` is included in the core from version 2.4 onwards, this is a necessary and secure fix.

I haven't set a minimum version in the dependency on `paypal/module-braintree-core` specified in the composer.json. Feel free to set the minimum version that you find most appropriate or leave it as it is.

Fixes [#2307](https://github.com/Adyen/adyen-magento2/issues/2307)

### DIFF
--- a/Gateway/Data/Order/OrderAdapter.php
+++ b/Gateway/Data/Order/OrderAdapter.php
@@ -11,10 +11,11 @@
 
 namespace Adyen\Payment\Gateway\Data\Order;
 
-use Magento\Payment\Gateway\Data\Order\AddressAdapterFactory;
+use Magento\Quote\Api\CartRepositoryInterface;
 use Magento\Sales\Model\Order;
+use PayPal\Braintree\Gateway\Data\Order\AddressAdapterFactory;
 
-class OrderAdapter extends \Magento\Payment\Gateway\Data\Order\OrderAdapter
+class OrderAdapter extends \PayPal\Braintree\Gateway\Data\Order\OrderAdapter
 {
     /**
      * @var Order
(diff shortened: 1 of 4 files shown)


RokPopov:
Hi @mikimpe,

Thank you for your contribution. However, based on our experiences with this type of issue (Magento preferring Braintree module), we are changing where we are calling getQuoteId() from. This specific issue is solved in [this PR](https://github.com/Adyen/adyen-magento2/pull/2486), therefore I will be closing this PR.

Cheers,
Rok

PR #2523, opened 2024-02-26
[ECP-8908] Add storeId parameter while obtaining success page URL for headless integrations

raoulritter:
**Description**

Update `Return/Index` controller to include storeID for multistore sites in order to prevent issues on returnURL.

**Tested scenarios**

- Headless iDeal payments


Controller/Return/Index.php
```diff
@@ -92,6 +92,7 @@ public function execute(): void
     {
         // Receive all params as this could be a GET or POST request
         $redirectResponse = $this->getRequest()->getParams();
+        $storeId = $this->storeManager->getStore()->getId();
```
candemiralp, line 95: This will most probably return the default store id in that context. Using order object would be a safer object to get the storeId of the payment.

candemiralp:
@raoulritter Could you please add description to the PR?

Controller/Return/Index.php
raoulritter: I think it should be alright as we are also using it to get the [storeReturnURL](https://github.com/Adyen/adyen-magento2/blob/develop/Gateway/Request/ReturnUrlDataBuilder.php#L68) in the same manner.

khushboo-singhvi (approved):

candemiralp (dismissed):

candemiralp (approved):

khushboo-singhvi (approved):

PR #2528, opened 2024-02-28
[ECP-8895] Crosscheck storedPaymentMethods coming from paymentMethods response with vault DB table

RokPopov:
**Description**
The response from paymentMethods call contains all the recurring tokens that have been generated under that merchant account and might be coming from a different Magento storefront.

Idea of this ticket is filtering out the tokens in the paymentMethods response if that token doesn't exist in the vault_payment_token table. Admin panel configuration is also added, so this crosschecking can be disabled from the configuration if the merchant wants to use the tokens coming from a different store.

**Tested scenarios**
- the tokens that are not existing in vault_payment_token table are removed from the response of paymentMethods call

### DIFF
--- a/Helper/PaymentMethods.php
+++ b/Helper/PaymentMethods.php
@@ -34,7 +34,9 @@ use Magento\Sales\Model\Order;
 use Adyen\Payment\Helper\Data as AdyenDataHelper;
 use Magento\Store\Model\ScopeInterface;
 use Magento\Store\Model\Store;
-
+use Magento\Vault\Api\PaymentTokenRepositoryInterface;
+use Magento\Vault\Api\Data\PaymentTokenInterface;
+use Magento\Framework\Api\SearchCriteriaBuilder;
 class PaymentMethods extends AbstractHelper
 {
     const ADYEN_HPP = 'adyen_hpp';
@@ -71,6 +73,8 @@ class PaymentMethods extends AbstractHelper
     private Config $configHelper;
     private ManualCapture $manualCapture;
     private SerializerInterface $serializer;
+    private PaymentTokenRepositoryInterface $paymentTokenRepository;
+    private SearchCriteriaBuilder $searchCriteriaBuilder;
 
     public function __construct(
         Context $context,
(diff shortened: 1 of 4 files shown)


Helper/PaymentMethods.php
```diff
@@ -179,6 +187,29 @@ protected function fetchPaymentMethods(?string $country = null, ?string $shopper
...
+            $gatewayTokens = array_map(function ($paymentToken) {
+                return $paymentToken->getGatewayToken();
+            }, $paymentTokens);
+
+            if (isset($responseData['storedPaymentMethods'])) {
+                $storedPaymentMethods = $responseData['storedPaymentMethods'];
+                $responseData['storedPaymentMethods'] = array_filter($storedPaymentMethods, function ($method) use ($gatewayTokens) {
+                    return in_array($method['id'], $gatewayTokens);
+                });
+            }
```
candemiralp, line 210: This `if` condition is redundant. Availability of the variable `$responseData['storedPaymentMethods']` is already beindg checked on line 191.

Helper/PaymentMethods.php
```diff
@@ -179,6 +187,29 @@ protected function fetchPaymentMethods(?string $country = null, ?string $shopper
...
+        }
```
candemiralp, line 211: What would you think about moving this filtering algorithm to a private method to increase the maintainability and readability?

etc/config.xml
```diff
@@ -66,6 +66,7 @@
                 <title>Stored Cards (Adyen)</title>
                 <model>AdyenPaymentCardVaultFacade</model>
                 <require_cvc>1</require_cvc>
+                <allow_multistore_tokens>1</allow_multistore_tokens>
```
candemiralp, line 69: Since the configuration field was defined under `payment/adyen_abstract/allow_multistore_tokens` path, default value should be under `<adyen_abstract>` node of `config.xml`.

etc/adminhtml/system/adyen_tokenization.xml
```diff
@@ -28,5 +28,11 @@
             <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
             <config_path>payment/adyen_cc_vault/require_cvc</config_path>
         </field>
+        <field id="allow_mutistore_tokens" translate="label" type="select" sortOrder="30" showInDefault="1" showInWebsite="1"
+            showInStore="1">
+            <label>Allow multistore tokens</label>
+            <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+            <config_path>payment/adyen_abstract/allow_multistore_tokens</config_path>
+        </field>
```
candemiralp, line 36: Adding a tooltip might help us having a self-explanatory field.

candemiralp (dismissed):

peterojo (approved):

candemiralp (approved):

raoulritter (approved):

PR #2529, opened 2024-02-28
Fixed issue for order status pre-authorized state configuration

sarfarazbheda:
**Description**

As per the Webhook Helper[adyen/module-payment/Helper/Webhook.php] file, we can assign states if order is in pre-authorized condition.

image

But when we create new order status "adyen_authorized" [ref. from adyen-magento2/Model/Config/Source/PreAuthorized.php], and assign this status in Adyen configuration, it's not working.

image

When the order is Authorized, Even if we have assigned new status in configuration, it will show order in "Pending" status only. The reason is, below function only grab the first item from the assigned Statuses. [file ref. vendor/adyen/module-payment/Helper/Order.php]

image

So As solution, I switched the sequence in Webhook Helper[adyen/module-payment/Helper/Webhook.php] file to pick "adyen_authorized" first and then check for "NEW" status. See updated file below:

image

**Tested scenarios**

- Create new Order status from admin panel
- Assign order status for pre-authorized configuraiton
- Verify order status once Payment is authorized

Fixes  
There is no such issue reported.

### DIFF
--- a/Helper/Webhook.php
+++ b/Helper/Webhook.php
@@ -46,7 +46,7 @@ class Webhook
      * Indicative matrix for possible states to enter after given event
      */
     const STATE_TRANSITION_MATRIX = [
-        'payment_pre_authorized' => [Order::STATE_NEW, PreAuthorized::STATE_ADYEN_AUTHORIZED],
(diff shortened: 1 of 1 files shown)


sarfarazbheda:
Please approve workflows

sarfarazbheda:
@candemiralp  please check this PR

PR #2597, opened 2024-04-25
[ECP-9111] Fix multishipping failed if only one payment method is activated

RokPopov:
**Description**
The issue this PR is solving is that it is not possible to complete the payment in multishipping checkout if only one payment method is enabled in Adyen Customer Area. This is solved by including a nested div with a custom data attribute and an afterRender binding to trigger `selectPaymentMethod()`.

**Tested scenarios**
- card / alternative payment method multishipping payment when only one payment method is enabled in Adyen CA
- card / alternative payment method multishipping payment when multiple payment methods are enabled in Adyen CA


khushboo-singhvi (approved):

candemiralp (approved):

candemiralp:
Merging this PR with a failing pipeline since the issue is reproducible on `develop` branch. Issue will be solved separately.

PR #2611, opened 2024-05-13
[ECP-9092] - Remove redundant loadCheckoutComponent call on initialization to rely on observer updates

raoulritter:
**Description**

This commit removes the immediate invocation of `loadCheckoutComponent` during the module initialization.  The component will now rely solely on the observer's updates to trigger this function.

**Tested scenarios**

Tested the checkout process using both HPP and  CC PMs. 

Fixes

### DIFF
--- a/Test/Unit/Model/Config/Backend/DonationAmountsTest.php
+++ b/Test/Unit/Model/Config/Backend/DonationAmountsTest.php
@@ -52,7 +52,7 @@ class DonationAmountsTest extends \PHPUnit\Framework\TestCase
     public function testValidateDonationAmounts($donationAmounts, $expectedResult)
     {
         $result = $this->donationAmounts->validateDonationAmounts(explode(',', $donationAmounts));
-        $this->assertEquals($result, $expectedResult);
+        $this->assertEquals($expectedResult, $result);
     }
 
     public static function donationAmountsProvider()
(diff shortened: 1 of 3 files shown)


acampos1916 (approved):

candemiralp (approved):

khushboo-singhvi (approved):

LexKoomen:
Hee Adyen team, we just deployed our storefront with this release and found out that for a very few specific set of users the loader pops up for loading in the hpp-methods but that loader never disappears. The weird thing is that it is not related to accounts or browsers because logging in on their accounts couldnt reproduce the problem and opening an incognito also didn't solve the errors for the clients that have the problem.

By downgrading 1 bug version and reimplementing the hard loadComponent calls we fixed the issue.

Not sure yet if it is related to this specific version where the loadComponent calls are removed or that it is related to some other part of our codebase, but I thought it was good to know.

jansentjeu:
Hello @raoulritter,

We have the same issue as @LexKoomen. This occurs sporadically because it depends on how quickly the payment methods are retrieved. If they are already retrieved before the subscribe is initialized, loadCheckoutComponent()/loadAdyenPaymentMethods() is no longer called. This causes the loader to remain visible and the payment methods are not loaded.

This issue caused a customer of ours to miss a lot of orders. Please roll back these changes.

Issue #2607, opened 2024-05-07
[ECP-9196] Allow shopperReference to be linked to GuestCustomer


ryan-lambert-IEX:
**Is your feature request related to a problem? Please describe.**
We are trying to pass in a tokenized `storedPaymentMethodId` to a `guest-cart` checkout, and right now there doesn't appear to be a way to link the `shopperReference` with the guest user. In the current plugin, the `shopperReference` is either the customerID or a UUID, causing our `guest-cart` payment to fail, as there is no way to get or change the `uuid` value prior to submission.

```
// in Helper/Requests.php
    public function getShopperReference($customerId, $orderIncrementId): string
    {
        if ($customerId) {
            $shopperReference = $this->adyenHelper->padShopperReference($customerId);
        } else {
            $uuid = Uuid::generateV4();
            $guestCustomerId = $orderIncrementId . $uuid;
            $shopperReference = $guestCustomerId;
        }

        return $shopperReference;
    }
}
```
For added context, we have a headless implementation which tokenizes card information by making a 0 dollar charge (per [docs](https://docs.adyen.com/online-payments/tokenization/) ). Our implementation handles raw credit card information (we are PCI certified) and we require our AdobeCommerce partners to submit a SAQ-A-EP and provision us with our own API key as to limit the scope of our tokenization flow.

We are trying to pass our tokenized payment data (we cannot pass raw card data, as this would break the PCI chain, nor can we use the AdyenJS library in our environment) to AdobeCommerce for submission. We use guest-carts since our flow collects payment information outside the usual "primary" store workflow.

Using the guest-cart flow as described above works great for Braintree and Stripe, but this `shopperReference` uuid seems to break our intended flow in our Adyen implementation.

**Describe the solution you'd like**
I think any of these things could fix this problem:

1. Use the guest-cart uuid `cartId` for the `shopperReference` vs the current `orderId.uuid` you use now
2. Allow a `shopperReference` value to be passed into the `paymentMethod.additional_data` which overrides the current behavior
3. Allow us to retrieve the `uuid` value prior to submission; can be attached to the `guest-cart/${cartId}`

**Describe alternatives you've considered**
I understand why you use a uuid for guests, as the assumption probably is stored payment methods belong to "stored customers". 

An alternative would be for Adyen to provide a way to get a single-use-payment-token like Braintree and Stripe vs. a stored payment method (with a shopperReference), but that change is probably far too great to hope for :]

**Additional context**
[...]

PR #1901, opened 2023-01-05
[PW-7519] Allow headless merchants to specify recurring processing model in the request

candemiralp:
**Description**

`recurringProcessingModel` parameter add to the REST and GraphQL endpoints to overwrite the default values in the plugin configuration.

**Tested scenarios**

- Klarna (REST + GraphQL)
- Card payments (REST + GraphQL)

PR #2620, opened 2024-05-22
add optional quoteId to getShopperReference to allow tokenization for…

ryan-lambert-IEX:
**Description**
This PR addresses #2607 and adds a way to link guest-carts to a predictable but unique shopperReference (guest-cart prefix plus the cartId). If the `quoteId` is missing, it falls back to the old `uuid` behavior.

In the above Github issue, I outlined a few possible solutions to our issue. I believe this is the "least invasive" way to achieve our goals.

**Tested scenarios**
Extended existing unit testing strategy for `getShopperReference`

#2607

### DIFF
--- a/Gateway/Request/AdditionalDataLevel23DataBuilder.php
+++ b/Gateway/Request/AdditionalDataLevel23DataBuilder.php
@@ -72,7 +72,7 @@ class AdditionalDataLevel23DataBuilder implements BuilderInterface
 
             $prefix = 'enhancedSchemeData';
             $requestBody['additionalData'][$prefix . '.totalTaxAmount'] = $this->adyenHelper->formatAmount($order->getTaxAmount(), $currencyCode);
-            $requestBody['additionalData'][$prefix . '.customerReference'] = $this->adyenRequestHelper->getShopperReference($order->getCustomerId(), $order->getIncrementId());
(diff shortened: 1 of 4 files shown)


RokPopov:
Hi @ryan-lambert-IEX,

Thank you for your contribution, it is greatly appreciated. After reviewing your implementation internally, we have a few suggestions to align the solution with our overall strategy.

We believe it would be beneficial for merchants to have the flexibility to pass their own shopper reference in the request, overriding the default value provided by our plugin. This approach would empower merchants to manage shopper references as per their specific requirements, without the need for additional logic within the plugin. An example of the optional parameter implementation according to our strategy could be found in [this PR](https://github.com/Adyen/adyen-magento2/pull/1901).

Could you kindly adjust your PR to reflect this feedback? We are confident that these changes will enhance the functionality and usability of the plugin.

Thank you once again for your valuable contribution and collaboration.

Kind regards,
Rok

ryan-lambert-IEX:
@RokPopov I made the suggested changes and also tried to import the `getShopperReference` logic from `Requests` to Donations since it uses similar code.

I think this will achieve the basic aim of passing in a `shopperReference` when creating an order. It defaults to the old behavior.

RokPopov (approved):

PR #2648, opened 2024-05-31
[ECP-9249] Support new Ideal payment flow

candemiralp:
**Description**

Ideal has introduced a new payment flow with a full redirected shopper journey. This PR upgrades Adyen Web components version to v5.65.0 to support this new flow.

Issuer selection functionality has been removed with this update and moved to the Ideal hosted page.

Screenshot 2024-05-30 at 11 54 36

Screenshot 2024-05-30 at 11 54 46

**Tested scenarios**

- E2E test flow has been updated [here](https://github.com/Adyen/adyen-integration-tools-tests/pull/119).


khushboo-singhvi (approved):

raoulritter (approved):

aschrammel:
Hi @candemiralp 

your description says, that the issuer selection is removed with this update.
Unfortunately I'm running into the error, that the Adyen API returns a `Missing payment method details: issuer` error (caused by `\Adyen\Service::requestHttp` on line 90) as soon as I try to place the order. This - in the checkout - results in an `Error with payment method, please select a different payment method.` (catched in `\Adyen\Payment\Gateway\Validator\CheckoutResponseValidator::validate` on line 130).

We're on Magento 2.4.6-p6 and trying to upgrade the Adyen integration from 9.4.1 to 9.6.2 at the moment.
Could you please give any advice? Thanks a lot.

xxuanng:
Root cause for this: in TEST, if you already have iDEAL added as a payment method, you are probably using the acquirer account iDealSim. This is the simulator for the old flow and will break with this new feature. 

Solution: add iDEAL in Customer Area again and make sure the acquirer account added this time is AdyenIdeal. This is the acquirer account for the new flow. 

Note: in LIVE, the change to the new flow works ootb, no change in payment method setup in Customer Area required.

aschrammel:
Thanks a lot for the feedback - that was exactly the reason 😄

PR #2652, opened 2024-06-11
[ECP-9097] Remove usage of handleReject() for payment methods except for PayPal

SushmitaThakur:
**Description**
`handlereject()` method is not present for component in case of payment methods other than PayPal, this PR includes a base function `handleOnfailure()` for handling orderRedirect fail for in general payment methods and overrides this method for PayPal ensuring `handleReject()` is called only if the payment method is PayPal. 

**Tested scenarios**

1. With Ideal payment, `handleOnfailure()` should be utilized from base (ayden-pm-method)
2. With PayPal, `handleOnfailure()` should be utilized from PayPal method renderer


view/frontend/web/js/view/payment/method-renderer/adyen-paypal-method.js
```diff
@@ -35,6 +35,12 @@ define(
 
                 this.actionComponent = component.handleAction(action);
             },
+            handleOnFailure: function(response, component) {
+                const self = this;
+                self.isPlaceOrderAllowed(true);
```
candemiralp, line 40: Since execution happening in the same context, you don't need to define `self` and use it instead of `this`. Feel free to use `this` directly here.

view/frontend/web/js/view/payment/method-renderer/adyen-pm-method.js
```diff
@@ -320,6 +318,12 @@ define(
                 );
             },
 
+            handleOnFailure: function(response, component) {
+                const self = this;
+                self.isPlaceOrderAllowed(true);
```
candemiralp, line 323: Since execution happening in the same context, you don't need to define `self` and use it instead of `this`. Feel free to use `this` directly here.

candemiralp (approved):

khushboo-singhvi (approved):

PR #2674, opened 2024-07-08
[ECP-9194] Ensure Adyen payment methods load correctly

### LINKED ISSUE #2492: [ECP-9194][ECP-8983] Incompatibility with amasty/module-single-step-checkout
**Describe the bug**
After upgrading adyen/module-payment module from 8.17.0 to 9.0.7 we are having some issues with Amasty One Step Checkout.
If you go to the checkout page while logged in with an account that already has a default shipping address saved, initially, no Adyen payment methods will be visible. However, if you change the shipping address on the checkout page to another one already saved in the account or create a new one, Adyen payment methods will become visible at that point.
It seems that in the v9 Adyen payment methods initialization is bound to a function that is not called if the customer goes to the checkout with an address already selected.
In the 8.17.0 we don't have this issue.

**To Reproduce**
1. Install Amasty One Step Checkout **https://amasty.com/one-step-checkout-for-magento-2.html** and configure it
2. Create a customer account and set a default shipping address
3. Go to the checkout
4. Error: no Adyen payment methods visible

**Expected behavior**
You can see and use Adyen payment also without changing the shipping address on the checkout page

**Magento version**
2.4.5-p5

**Plugin version**
9.0.7

**Additional context**
If modifying https://github.com/Adyen/adyen-magento2/blob/develop/view/frontend/web/js/view/payment/adyen-methods.js initialize function from

```
            initialize: function () {
                this._super();

                // Virtual quote doesn't call payment-information or shipping-information endpoints.
                // payment-information endpoint should be called manually to fetch Adyen extension attributes.
                if (quote.isVirtual()) {
                    getPaymentInformation();
                }
            }
```
to

```
            initialize: function () {
                this._super();

                getPaymentInformation();
            }
```
the payment methods are visible and the payment process seems to work fine.

### LINKED ISSUE #2602: [ECP-9194] Adyen 9.2.0 conflicting with Avalara Avatax - Payment methods not loaded - adyen_payment_methods_response empty
**Describe the bug**
Adyen 9.2.0 is not working correctly with other extension (in this case Avalara Avatax).

After we have updated adyen/module-payment to version 9.2.0,
We have encountered a problem that the Adyen payment method is not showing up on Magento checkout.

I was checking what is the reason,
And the problem is that there is "conflict" within mixin in Avalara extension:
https://github.com/avadev/Avalara-AvaTax-for-Magento2/blob/master/view/frontend/web/js/model/shipping-save-processor/default.js

In this mixin there is asynchronous shipping-information request, and because of this response with adyen_payment_methods_response is received too late and adyenPaymentMethodsResponse is empty in:
html/vendor/adyen/module-payment/view/frontend/web/js/model/get-payment-information-mixin.js

image

**To Reproduce**
1. Install adyen/module-payment, version 9.2.0 & avalara/avatax-magento, version 2.5.0
2. Go to Checkout
3. Go to Payment Step

**Expected behavior**
We should have all payment method visible.

**Magento version**
2.4.5-p6

**Plugin version**
9.2.0

**PHP Version**
8.1

 - OS: MacOS 14.4.1
 - Browser Chrome 
 - Version 124.0.6367.92

**Additional context**
I'm aware that your extension is working perfectly find with clean Magento website and it might be the case that Avalara extension is "incorrectly" integrated that it's conflicting with yours.
However it would be great to know your opinion to define next steps.

RokPopov:
**Description**
This PR addresses an issue where Adyen payment methods were not visible at the Magento checkout when both the Adyen Payment plugin (version 9.2.0) and Avalara Avatax plugin (version 2.5.0) were installed. The problem was caused by a conflict within the mixins of these plugins, where Avalara's asynchronous shipping-information request delayed the response with adyen_payment_methods_response.

### Changes

- Refactored the initialization logic in the Adyen payment component to check the properties of the `adyen-payment-service.js` class instead of checking for a virtual quote
- Now calls `getPaymentInformation()` function if the `paymentMethods` or `connectedTerminals` observables are `null`

**Tested scenarios**
- proceeded through the checkout process with various combinations of products (virtual, physical)
- verified Adyen payment methods are correctly displayed in those scenarios

Fixes  #2602
Fixes  #2492


candemiralp (approved):

Djohn12:
Hi there team,

I believe there's a small error with the PR.

I have found that defining `Adyen_Payment/js/adyen-payment-service` as it is causes a 404 error and breaks all Adyen components init.

Right reference in the define array argument should be `Adyen_Payment/js/model/adyen-payment-service` in order to match the `vendor/adyen/module-payment/view/frontend/web/js/model/adyen-payment-service.js` file.

Hope this helps,
Best Regards

RokPopov:
Hi @Djohn12,

Thank you for catching and reporting this! I created another [PR](https://github.com/Adyen/adyen-magento2/pull/2691) to mitigate this so the release won't be broken.

Kind regards,
Rok

PR #2707, opened 2024-08-12
Prevent order cancellation on paymentsDetails Refused response from Klarna

### LINKED ISSUE #2702: [ECP-9444] Klarna payment gets cancelled on paymentDetails endpoint, but then authorised successfully via webhook
**Describe the bug**

We are facing an issue with Klarna through Adyen integration, where `paymentsDetails` response result code is refused, which makes Magento cancel the order, but then few minutes later Adyen sends webhook with Authorised event. However since order is already cancelled Magento cannot move it to processing state. This ends up with critical issue where funds are taken from customer's account but his/hers order is cancelled.

Example case:

image

**Aug 6, 2024, 21:14:46** Magento calls payments-details endpoint of Adyen integration. Adyen then forwards this request to Klarna. Klarna answers that payment is refused. 

image

So Adyen tells Magento to cancel the order here:

https://github.com/Adyen/adyen-magento2/blob/9.3.0/Helper/PaymentResponseHandler.php#L274

Magento cancels the order.

Aug 6, 2024, 21:14:53, 7 seconds later, Adyen receives Authorisation event from Klarna. Payment gets captured.

image

Aug 6, 2024, 21:16:17 Adyen sends webhook request to Magento to tell it that payment is successful and it should update order state to processing

Aug 6, 2024, 21:18:05 Magento cron runs and processes the webhook. At this point order is already cancelled, so Magento won't move it to processing state, because this kind of transition is not allowed.

We have opened a ticket via Adyen customer area too providing more data (case number **04124616**) but I'm also opening ticket here, because the answer of the support was:

`I checked in the logs for the shared payment and we get both of the two responses, however since the later one is the most reliable response, the first resultCode is overwritten and payment is authorised.
The second status is the one that is shared in the Authorisation Webhook as that is the most reliable source of payment status in this case. The best course of action for the future is to relay on the Authorisation Webhook rather than the details call for your order creation system as stated in our docs [here](https://docs.adyen.com/development-resources/webhooks/webhook-types/#event-codes).`

so I'm wondering why the module operates differently. 

I'd gladly help to resolve the issue but I'd need to first understand what is expected behaviour in such scenario. Unfortunately we don't know anything about the reasons of Klarna refusing the payment, thus I'm not able to replicate it on my own. 

I'm just guessing but should we not cancel the order at all in case Klarna sends refused resultCode, because the session is still open and it is possible that the same payment will end up successful? Or perhaps should Klarna payment orders automatically go to `payment_review` state instead of `pending_payment` here:

https://github.com/Adyen/adyen-magento2/blob/main/Observer/SetOrderStateAfterPaymentObserver.php#L119

and then `canCancel` method on order during `paymentsDetails` would return false and order would not get cancelled?

I know that there have been some changes in cancellation logic here:

https://github.com/Adyen/adyen-magento2/pull/2592/files

which were introduced in 9.5.1 (we are using 9.3.0) but my understanding is that it would not resolve our issue anyway as order in `pending_payment` would still return true on `canCancel()` call.

**To Reproduce**

Unfortunately at this point I don't have a clue how to replicate the issue with getting Payment refused response from Klarna with the same PSP reference and get it authorised 7 seconds later. We don't have enough data logged in Adyen dashboard / Magento logs to understand reason for Klarna refusal in the first place.

**Expected behavior**
Order is not cancelled on Klarna payment refusal if there is still a chance that the same payment session will get authorised later and processed successfully by webhook.

**Magento version**
2.4.6-p6

**Plugin version**
9.3.0

I don't think this is browser / device related issue as it seems to be an inconsistency between paymentDetails response and following webhook response.

I don't think this is browser / device related issue as it seems to be an inconsistency between paymentDetails response and following webhook response.

pknap:
**Description**

This PR relates to issue raised in https://github.com/Adyen/adyen-magento2/issues/2702

According to Adyen support there is some intermittent bug on Klarna side, where Klarna sends Refused result code on paymentsDetails request, but payer session is still active, and in some rare cases they are able to finish the payment successfully. This leads to a situation where Magento's order is cancelled by paymentsDetails handler and then when authorised webhooks comes in, cron process is not able to move the order back from cancelled to processing state. 

In the end customer's funds are captured but the order is not fulfilled which causes frustration.

This is my idea on how to fix it, I'm not sure if this is done 100% according to Adyen's idea on how this should be tackled, but please let me know if I can be of any help resolving this otherwise. 

**Tested scenarios**

Since this is a very intermittent issue and not replicable in a sandbox environment I was not able to replicate it for testing. 

Fixes #2702

### DIFF
--- a/Helper/PaymentResponseHandler.php
+++ b/Helper/PaymentResponseHandler.php
@@ -274,8 +274,14 @@ class PaymentResponseHandler
             case self::REFUSED:
             case self::CANCELLED:
                 // Cancel order in case result is refused
+                $result = false;
                 if (null !== $order) {
                     // Check if the current state allows for changing to new for cancellation
+                    if ($this->dataHelper->isPaymentMethodOfType($order->getPayment()->getMethod(), Data::KLARNA) && $paymentsDetailsResponse['resultCode'] === self::REFUSED) {
+                        // Klarna Refused payment response on paymentsDetails endpoint is not reliable. Wait for webhook notification in this case.
+                        $this->adyenLogger->addAdyenResult('Do nothing wait for the notification');
+                        break;
+                    }
                     if ($order->canCancel()) {
                         // Proceed to set cancellation action flag and cancel the order
                         $order->setActionFlag(\Magento\Sales\Model\Order::ACTION_FLAG_CANCEL, true);
@@ -284,7 +290,6 @@ class PaymentResponseHandler
                         $this->adyenLogger->addAdyenResult('The order cannot be cancelled');
                     }
                 }
-                $result = false;
                 break;
             default:
                 $this->adyenLogger->error(


candemiralp:
Hello @pknap,

Thank you for your contribution. I made a detailed explanation on the issue [here](https://github.com/Adyen/adyen-magento2/issues/2702#issuecomment-2328276648). Please see the comment.

Best Regards,
Can

candemiralp:
Hello @pknap,

Thank you for your contribution. I would like to inform you that this issue has been resolved on Adyen integration level. Please see my comment [here](https://github.com/Adyen/adyen-magento2/issues/2702#issuecomment-2376209741).

Since this PR is not relevant anymore, I am closing this one.

Best Regards,
Can

PR #2709, opened 2024-08-13
Fix order pending status if canceled

### LINKED ISSUE #2708: Order in pending status if already cancelled
**Describe the bug**

Bug is partially fixed in the following MR https://github.com/Adyen/adyen-magento2/pull/2592 but there is still a small problem.

Indeed, if the graphQl `adyenPaymentDetails` query is executed multiple times for, the order is passed to status `pending` instead of remaining `canceled`.

As the following screenshot shows, the order is in pending status after executing `adyenPaymentDetails` 2 times (or more)

Capture d’écran du 2024-08-13 12-26-25

Capture d’écran du 2024-08-13 14-18-52

**To Reproduce**

1. Try place order with adyen_twint payment
2. Cancel payment in Twint interface
3. Execute `adyenPaymentDetails` multiple times
4. The first time the order is cancelled with canceled status and after the order is in pending status ...

**Expected behavior**

If the order is already cancelled, the status must not change.

**Versions**

- Magento: ver. 2.4.5-p8
- Adyen module: 9.7.0

### LINKED ISSUE #2726: Race condition may cause the order status to be reverted to pending
**Description**
When a Customer is delayed between the PSP and the Magento return page the order status can be reversed from `processing`, or in fact any possible state (like `holded`) back to `pending`.

**Cause**
The Controller for the return page calls the PaymentResponseHandler, which in turn updates the order according to the response.

**Code**
1. https://github.com/Adyen/adyen-magento2/blob/365cd45433fd0922c7fbd985a8e732e94ee99376/Controller/Return/Index.php#L98
2. https://github.com/Adyen/adyen-magento2/blob/365cd45433fd0922c7fbd985a8e732e94ee99376/Controller/Return/Index.php#L164
3. https://github.com/Adyen/adyen-magento2/blob/365cd45433fd0922c7fbd985a8e732e94ee99376/Helper/PaymentResponseHandler.php#L210

**Solution**
Check the current order status before doing any changes. If the status isn't `pending_payment` leave the order as-is, since either a previous executing of the controller or the webhook processor has already updated the order.

**I also highly recommend to implement locking to prevent race conditions in case the return controller is executed at the same time as the notification processor (`Adyen\Payment\Helper\Webhook::processNotification(...)`). If no lock is applied (the current situation) those processes may collide. Please note you'll have to place the lock before loading the order, or you'll be using a cached and possible out-of-date model.**

_Note: Finished my previous LEGO set 😇_

dimitriBouteille:
**Description**

Update to check the order status before updating

**Tested scenarios**

- Refused orders.

Fixes #2708 
Fixes #2726


peterojo (dismissed):

Helper/PaymentResponseHandler.php
```diff
@@ -207,8 +207,10 @@ public function handlePaymentsDetailsResponse(
              * if no additional action is required according to /paymentsDetails response.
              * Otherwise keep the order state as pending_payment.
              */
-            $order = $this->orderHelper->setStatusOrderCreation($order);
-            $this->orderRepository->save($order);
+            if (!$order->isCanceled()) {
```
candemiralp, line 210: Hello @dimitriBouteille,

Thank you for your contribution. Obviously, this fix solves the issue for order cancelled cases. However, I would like to recommend a holistic approach as we also have some reports for other statuses rather than only `cancelled`.

So, instead of checking whether if the order is already cancelled in the inner if block, I would recommend removing the inner if block and adding a second condition to the if block on line 204 to check if the order state is `pending_payment` or not. Because, the logic in the outer if block should only run under the condition of order should be at `pending_payment` state AND order's `resultCode` should not be in `ACTION_REQUIRED_STATUSES`.

Best regards,
Can

candemiralp (approved):

peterojo (approved):

Helper/PaymentResponseHandler.php
candemiralp: Hello @dimitriBouteille,

Thank you for your contribution. That looks great and solves the issue indeed. After the reviews, this PR will be merged and released.

Best Regards,
Can

dimitriBouteille: Perfect :rocket: Is it possible to include it in the next release ([Release v9.7.3](https://github.com/Adyen/adyen-magento2/pull/2722)) ?

candemiralp: Hello @dimitriBouteille! 

Yes, we will include the PR in the coming release.

Best Regards,
Can

PR #2723, opened 2024-08-22
Remove unused var from adyen-pm-method.js

acampos1916:


SushmitaThakur (approved):

view/frontend/web/js/view/payment/method-renderer/adyen-pm-method.js
```diff
@@ -37,7 +37,6 @@ define(
         adyenCheckout
     ) {
         'use strict';
-        let popupModal;
```
candemiralp, line 40: This variable is used in the renderer to hold the name of the DOM element while creating a container for the action component.

candemiralp (changes requested):

PR #2724, opened 2024-08-26
[ECP-9408] Fix handling Giftcard payments responses

SushmitaThakur:
**Description**
This PR fixes partial payment with one or more giftcards + a final non-giftcard PM responses for the shopper to land on the correct page. 

1. This fix also ensures that `processGiftcards()` now includes all giftcard responses, rather than just the last one. Previously, `handleGiftcards()` returned only the final giftcard response (whether successful or failed) when handling multiple giftcards. It also includes a unit test for checking the response type for multiple giftcard scenario. 

2. The CheckoutResponse validator and handler classes are modified to handle an array of responses.

3. New Unit tests added for the following classes - 
  - CheckoutResponseValidator
  - CheckoutPaymentCommentHistoryHandler
  - CheckoutPaymentsDetailsHandler
  - CheckoutPaymentCommentHistoryHandler

** Manually Tested scenarios**
1. Payment with a single gift card with correct details - should succeed 
2. Payment with a single gift card with incorrect details - should fail 
3. Payment with two gift cards, one with correct details, one with incorrect details - should fail 
4. Payment with one gift card with correct details and credit card with correct details - should succeed 
5. Payment with one gift card with incorrect details and credit card with correct details - should fail 
6. Payment with two gift cards, one with correct details, one with incorrect details and one credit card with correct details - should fail


Gateway/Http/Client/TransactionPayment.php
```diff
@@ -204,6 +230,8 @@ private function handleGiftcardPayments(
             $response = $service->payments(new PaymentRequest($giftcardPaymentRequest));
```
candemiralp, line 230: Related to my first comment, what would you think about failing the order immediately here after getting an error or refusal here synchronously? IMO, it doesn't make sense to proceed to the next payments if the previous one already fails. Eventually, Adyen order will fail since one part has already failed.

peterojo (approved):

candemiralp (approved):

PR #2727, opened 2024-09-02
[ECP-9398] Solving the issue of inconsistency of Order state and status field

khushboo-singhvi:
Adding check in OFFER_CLOSE webhook process that if the order is already in canceled state, then do not update the state to new and proceed with cancel

**Description**

**Tested scenarios**

Fixes


candemiralp (approved):

SushmitaThakur (approved):

Helper/Webhook/OfferClosedWebhookHandler.php
```diff
@@ -105,7 +105,10 @@ public function handleWebhook(MagentoOrder $order, Notification $notification, s
         }
 
         // Move the order from PAYMENT_REVIEW to NEW, so that it can be cancelled
-        if (!$order->canCancel() && $this->configHelper->getNotificationsCanCancel($order->getStoreId())) {
+        if (!$order->isCanceled()
+            && !$order->canCancel()
```
candemiralp, line 109: Why we are checking the condition of `Order can not be cancelled` here?

candemiralp, line 109:
```suggestion
            && $order->canCancel()
```

candemiralp: This comment is not relevant anymore.

candemiralp (approved):

candemiralp (approved):

PR #2728, opened 2024-09-02
[ECP-9413] Implement CheckoutAnalytics API endpoints

candemiralp:
**Description**

As a part of the efforts related to the measuring the reliability of the plugin, we've decided to use `CheckoutAnalytics` API of Adyen.

This API has two endpoints to generate a checkout attempt and another one to send events (errors, info, logs) related to this attempt.

**Tested scenarios**

- Generating a checkout attempt and passing error messages related to that attempt
- Querying Log Search platform to see the error messages

### DIFF
--- a/Helper/CheckoutAnalytics.php
+++ b/Helper/CheckoutAnalytics.php
@@ -1,0 +1,297 @@
+<?php
+/**
+ *
+ * Adyen Payment module (https://www.adyen.com/)
+ *
+ * Copyright (c) 2024 Adyen N.V. (https://www.adyen.com/)
+ * See LICENSE.txt for license details.
+ *
+ * Author: Adyen <magento@adyen.com>
+ */
+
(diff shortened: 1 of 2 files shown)


Helper/CheckoutAnalytics.php
```diff
@@ -0,0 +1,297 @@
...
+
+    /**
+     * @param $response
+     * @return bool
+     * @throws InvalidArgumentException
+     */
+    private function validateInitiateCheckoutAttemptResponse($response): bool
+    {
+        if(!array_key_exists('checkoutAttemptId', $response)) {
+            throw new InvalidArgumentException(__('checkoutAttemptId is missing in the response!'));
```
peterojo, line 254: How about logging an error here and return `false`. So that this does not interrupt the flow for the shopper

khushboo-singhvi (approved):

PR #2771, opened 2024-10-15
[ECP-9431] Handling Refund Delay and Order Cancellation Issues when using Giftcard + redirected Payment method

khushboo-singhvi:
**Description**
When a customer places an order using a gift card in combination with any redirected payment method, an order is created which redeems the gift card balance, even if the full payment is not completed. As a result, the customer is unable to use the gift card again, despite the full order is not successfully placed.
This PR handles this situation wherein we cancel the already authorised partial payments(Giftcard Payments) if the payment with the redirect payment method fails.

**Tested scenarios**
Klarna, 3DS, Ideal Successful and failing situations tested.

Fixes


Helper/PaymentResponseHandler.php
```diff
@@ -279,6 +290,61 @@ public function handlePaymentsDetailsResponse(
...
+                            }
+                        } catch (\Exception $e) {
+                            // Log the error with relevant information for debugging
+                            $this->adyenLogger->error('Error canceling partial payments', [
+                                'exception' => $e->getMessage(),
+                                'giftcardData' => $giftcardData,
+                            ]);
+                        }
+                    }
+                }
```
candemiralp, line 347: What would you think about moving this logic to a separate private method? Currently, it decreases the readability and increases the overall complexity of the switch block.

Helper/PaymentResponseHandler.php
```diff
@@ -279,6 +290,61 @@ public function handlePaymentsDetailsResponse(
...
+                            $cancelRequest = [
+                                'order' => [
+                                    'pspReference' => $pspReference,
+                                    'orderData' => $orderData,
+                                ],
+                                'merchantAccount' => $merchantAccount,
+                            ];
+
+                            // Call the cancel service
+                            $cancelResponse = $service->cancelOrder(new CancelOrderRequest($cancelRequest));
```
candemiralp, line 328: Could you please log the request and response of `/orders/cancel` endpoint ([example](https://github.com/Adyen/adyen-magento2/blob/b5309b0c0ac0aff9755c82ee23a2e0e88dde5ce7/Gateway/Http/Client/TransactionPayment.php#L149))?

Helper/PaymentResponseHandler.php
```diff
@@ -279,6 +290,61 @@ public function handlePaymentsDetailsResponse(
...
+                            $response = $cancelResponse->toArray();
+
+                            if (is_null($response['resultCode'])) {
+                                // In case the result is unknown we log the request and don't update the history
+                                $this->adyenLogger->error(
+                                    "Unexpected result query parameter for cancel order request. Response: " . json_encode($response)
+                                );
+
+                                return false;
```
candemiralp, line 337: Returning `false` here will terminate the execution of the rest of the method where we cancel the Magento order. In addition, checking the validity of a single field of an API response is not scalable.

Alternatively, you can catch the exception thrown from the PHP API Library and log the error message only. For responseCode 200, there is only one available value in the [API Explorer](https://docs.adyen.com/api-explorer/Checkout/71/post/orders/cancel#responses-200) which is `resultCode: Received`. So, if there is no exception thrown from the library, we can assume everything is good.

.github/workflows/main.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php-version }}
-          tools: composer:v1
+          tools: composer:2
```
peterojo, line 27:
```suggestion
          tools: composer:v2
```

Test/Unit/Helper/PaymentResponseHandlerTest.php
```diff
@@ -399,6 +408,65 @@ public function testHandlePaymentsDetailsResponseCancelOrRefused($resultCode)
...
+                        "additionalData":{"paymentMethod":"svs","merchantReference":"123","acquirerCode":"Test"},
+                        "amount":{"currency":"EUR","value":5000},
+                        "merchantReference":"123",
+                        "order":{"amount":{"currency":"EUR","value":17800},"expiresAt":"2024-10-10T13:11:37Z",
+                        "orderData":"orderData....",
+                        "pspReference":"XYZ654",
+                        "reference":"123",
+                        "remainingAmount":{"currency":"EUR","value":12800}},
+                        "paymentMethod":{"brand":"svs","type":"giftcard"},
+                        "pspReference":"ABC123","resultCode":"Authorised"}'
```
peterojo, line 436:
```suggestion
                        "pspReference":"ABC123",
                        "resultCode":"Authorised"
                      }'
```

Helper/PaymentResponseHandler.php
```diff
@@ -341,4 +357,65 @@ private function isValidMerchantReference(array $paymentsDetailsResponse, OrderI
 
         return true;
     }
+
+    // Method to check for existing Gift Card payments
+    private function hasActiveGiftCardPayments($merchantReference, $order)
```
peterojo, line 362: This method not only checks if the payment "has active giftcard payments" but it also initializes the API client and attempts to cancel the order.
For proper separation of concerns, I would recommend implementing separate methods to check for used giftcards, and to make cancel calls (preferably using and reusable helper method used elsewhere)

peterojo (changes requested):
A few changes requested

candemiralp (dismissed):

Helper/PaymentResponseHandler.php
```diff
@@ -11,7 +11,10 @@
 
 namespace Adyen\Payment\Helper;
 
+use Adyen\Model\Checkout\CancelOrderRequest;
+use Adyen\Payment\Helper\Config as Config;
```
peterojo, line 15: import name is reduntant here
```suggestion
use Adyen\Payment\Helper\Config;
```

Test/Unit/Helper/PaymentResponseHandlerTest.php
```diff
@@ -29,6 +28,10 @@
 use Magento\Sales\Model\OrderRepository;
 use Magento\Sales\Model\Order\Status\HistoryFactory;
 use Adyen\Payment\Helper\StateData;
+use Adyen\Payment\Model\ResourceModel\PaymentResponse\Collection;
+use Adyen\Payment\Model\ResourceModel\PaymentResponse\CollectionFactory;
+use Adyen\Payment\Helper\Config as Config;
```
peterojo, line 33: Import alias is redundant
```suggestion
use Adyen\Payment\Helper\Config;
```

candemiralp (approved):

SushmitaThakur (approved):

peterojo (approved):

PR #2776, opened 2024-10-22
Fix Controller\Index::isResponseAlreadyProcessed() check with empty comment

### LINKED ISSUE #2775: [ECP-9534] Adyen\Payment\Controller\Return\Index::isResponseAlreadyProcessed() throw exception if comment is empty
In some case the order comment can be empty (null) so the function `str_contains` throw exception :(

Capture d’écran du 2024-10-22 10-34-44

> Deprecated Functionality: str_contains(): Passing null to parameter \#1 ($haystack) of type string is deprecated in /vendor/adyen/module-payment/Controller/Return/Index.php on line 213

dimitriBouteille:
**Description**

The PR fix the following error : 

> Deprecated Functionality: str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated in /Users/ansoumane/www/omega-corporate/vendor/adyen/module-payment/Controller/Return/Index.php on line 213

In some case, the order comment can be null : 

Capture d’écran du 2024-10-22 10-34-44

Fixes  #2775

### DIFF
--- a/Controller/Return/Index.php
+++ b/Controller/Return/Index.php
@@ -208,7 +208,7 @@ class Index extends Action
 
         foreach ($history as $status) {
             $comment = $status->getComment();
-            if (str_contains($comment, $pspReference) !== false) {
+            if (str_contains((string)$comment, $pspReference) !== false) {
                 return true;
             }
         }


candemiralp:
Hello @dimitriBouteille,

Thank you for creating this PR. However, we have a different proposal for the solution as `isResponseAlreadyProcessed()` method causes more issues. We will keep this PR here as a temporary solution but will introduce an abstract fix soon.

Best Ragards,
Can

candemiralp:
Hello @dimitriBouteille,

Thank you for your contrition. The fix to the given issue has been introduced in #2796 and will be release in the upcoming release this week. I am closing this issue as it's redundant after the introduced fix.

Best Regards,
Can

PR #2789, opened 2024-10-30
[ECP-9459]Solving the issue of Config "Refund strategy Giftcard" 

khushboo-singhvi:
**Description**
In a split payment, refund is not going to same payment method based on configuration setting "Refund strategy Giftcard" due inserting timestamp is same, it failed to follow the set strategy.
The solution in this PR aims at ORDER_CLOSED notification and getting the Order sort value and then depending on that we can process refunds based on strategy.

**Tested scenarios**

Processing Refunds for partial payments using refund strategy in place.

Fixes


Helper/Webhook/OrderClosedWebhookHandler.php
```diff
@@ -38,6 +39,11 @@ class OrderClosedWebhookHandler implements WebhookHandlerInterface
     /** @var OrderPaymentCollectionFactory */
     private $adyenOrderPaymentCollectionFactory;
 
+    /**
+     * @var SerializerInterface
+     */
+    private $serializer;
```
candemiralp, line 45: Could you please define the required serializer (Magento\Framework\Serialize\Serializer\Json) explicitly in `di.xml` for this class? Otherwise, it might use PHP serializer instead of JSON serializer. I assume JSON serializer is required here. [Example here](https://github.com/Adyen/adyen-magento2/blob/main/etc/di.xml#L1637)

Helper/Webhook/OrderClosedWebhookHandler.php
```diff
@@ -71,7 +79,50 @@ public function handleWebhook(
...
+            foreach ($additionalData as $key => $value) {
+                // Check if the key matches the pattern "order-X-pspReference"
+                if (preg_match('/^order-(\d+)-pspReference$/', $key, $matches)) {
+                    $orderIndex = (int)$matches[1];
+                    $pspReference = $value;
+                    $sortValue = $orderIndex;
+
+                    // Retrieve adyen_order_payment for this pspReference
+                    $adyenOrderPayment = $this->adyenOrderPaymentCollectionFactory->create()
+                        ->addFieldToFilter('pspreference', $pspReference)
```
candemiralp, line 97: As `pspreference` field is being used as a filter, shall we add an index for this field to the DB definition?

khushboo-singhvi: Ass discussed, we already have an UNIQUE index defined for pspreference and that would work.

candemiralp (approved):

peterojo (approved):

PR #2799, opened 2024-11-13
Handle errors when using the placeOrder GraphQL mutation

### LINKED ISSUE #2622: [ECP-9229] Undefined array key "order" in Plugin/GraphQlPlaceOrderAddCartId.php on line 63
**Describe the bug**

Some of our clients (very small proportion) cannot place order (Internal Server Error)

We were not able to reproduce, so we don't know how. But we found this error in logs :  

```
`main.ERROR: Warning: Undefined array key "order" in /var/www/html/vendor/adyen/module-payment/Plugin/GraphQlPlaceOrderAddCartId.php on line 63`
```

And indeed, that plugin does not take into consideration that original method can return an array without order : 
```
        if ($this->errors) {
            return [
                'errors' =>
                    $this->errors
            ];
        }
        return [
            'order' => [
                'order_number' => $order->getIncrementId(),
                // @deprecated The order_id field is deprecated, use order_number instead
                'order_id' => $order->getIncrementId(),
            ],
            'orderV2' => $this->orderFormatter->format($order),
            'errors' => []
        ];
```

**Magento version**
2.4.7

**Plugin version**
9.4.1

pmzandbergen:
**Description**
When an error occurs, the [Magento GraphQL PlaceOrder resolver](https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/QuoteGraphQl/Model/Resolver/PlaceOrder.php#L106
) does not return an order. Previously, the plugin did not consider that the original method could return an array without an order, which caused issues when using the `placeOrder` GraphQL mutation. This PR fixes this issue. 

**Tested scenarios**
1. Create a cart that an order cannot be created from (missing shipping info, no stock, etc.).
2. Place the order. 

The order is returned correctly, otherwise a PHP error would be thrown due to the undefined array key. 

Fixes #2622


pmzandbergen:
Fixes https://github.com/Adyen/adyen-magento2/issues/2622

candemiralp (approved):

candemiralp:
Hello @pmzandbergen,

Thank you for your contribution. We will merge and release this PR after completing the reviews.

Best Regards,
Can

pmzandbergen:
@candemiralp could you merge this PR soon? We would like to clean-up unnecessary patches before our updating the Adyen Payment module on our projects.

Thx, Pieter

PR #2800, opened 2024-11-13
[ECP-9450] Implement configurable 3DS2 authentication flow

candemiralp:
**Description**

This pull request introduces a new field to configure native and redirect methods for 3D Secure 2 card payment authentications. 

This has become a necessity after Adobe has introduces strict CSP ruling on payment pages to comply with PCI DSS 4.0 requirements. This requirement enforces our merchants to add all of the 3D Secure 2 domain names to their CSP allowlist. Therefore, this PR enables our merchants to fall back [3D Secure 2 Redirect](https://docs.adyen.com/online-payments/3d-secure/redirect-3ds2/) authentication if they are not able to add domains to their allow lists.

Screenshot 2024-11-13 at 15 27 18

**Tested scenarios**

- Card payments with 3D Secure 2 with native and redirect flows configured from the plugin configuration.


Gateway/Request/CheckoutDataBuilder.php
```diff
@@ -209,7 +210,8 @@ public function build(array $buildSubject): array
             unset($requestBody['installments']);
         }
 
-        $requestBody['additionalData']['allow3DS2'] = true;
+        $requestBody['additionalData']['allow3DS2'] =
+            $this->configHelper->getThreeDSMode($storeId) === ThreeDSMode::THREEDS_MODE_NATIVE;
```
peterojo, line 214: Won't this will be false if the new config value is not yet set?
I think we should keep the default behaviour until the merchant sets the config to redirect

candemiralp: The default value will always be `native` as we have set it in `config.xml`.

SushmitaThakur (dismissed):

SushmitaThakur (approved):

PR #2802, opened 2024-11-14
[ECP-9450-v8] Implement configurable 3DS2 authentication flow

candemiralp:
**Description**

This pull request introduces a new field to configure native and redirect methods for 3D Secure 2 card payment authentications. 

This has become a necessity after Adobe has introduces strict CSP ruling on payment pages to comply with PCI DSS 4.0 requirements. This requirement enforces our merchants to add all of the 3D Secure 2 domain names to their CSP allowlist. Therefore, this PR enables our merchants to fall back [3D Secure 2 Redirect](https://docs.adyen.com/online-payments/3d-secure/redirect-3ds2/) authentication if they are not able to add domains to their allow lists.

Screenshot 2024-11-14 at 13 35 14

**Tested scenarios**

- Card payments with 3D Secure 2 with native and redirect flows configured from the plugin configuration.

### DIFF
--- a/Gateway/Request/CheckoutDataBuilder.php
+++ b/Gateway/Request/CheckoutDataBuilder.php
@@ -16,6 +16,7 @@ use Adyen\Payment\Helper\Config;
 use Adyen\Payment\Helper\Data;
 use Adyen\Payment\Helper\StateData;
 use Adyen\Payment\Helper\OpenInvoice;
+use Adyen\Payment\Model\Config\Source\ThreeDSFlow;
 use Adyen\Payment\Model\Ui\AdyenBoletoConfigProvider;
 use Adyen\Payment\Model\Ui\AdyenPayByLinkConfigProvider;
 use Adyen\Payment\Observer\AdyenCcDataAssignObserver;
@@ -216,7 +217,8 @@ class CheckoutDataBuilder implements BuilderInterface
             unset($requestBody['installments']);
         }
 
-        $requestBody['additionalData']['allow3DS2'] = true;
+        $requestBody['additionalData']['allow3DS2'] =
+            $this->configHelper->getThreeDSFlow($storeId) === ThreeDSFlow::THREEDS_NATIVE;
 
         if (isset($requestBodyPaymentMethod)) {
             $requestBody['paymentMethod'] = $requestBodyPaymentMethod;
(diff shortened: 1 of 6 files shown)


peterojo (approved):

Gateway/Request/CheckoutDataBuilder.php
```diff
@@ -216,7 +217,8 @@ public function build(array $buildSubject): array
             unset($requestBody['installments']);
         }
 
-        $requestBody['additionalData']['allow3DS2'] = true;
+        $requestBody['additionalData']['allow3DS2'] =
```
acampos1916, line 220: I see that the field in deprecated in the API, is this a concern? https://docs.adyen.com/api-explorer/Checkout/71/post/payments#request-additionalData-AdditionalData3DSecure-allow3DS2

etc/adminhtml/system/adyen_cc.xml
```diff
@@ -47,8 +47,13 @@
             <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
             <config_path>payment/adyen_cc/enable_click_to_pay</config_path>
         </field>
+        <field id="threeds_flow" translate="label" type="select" sortOrder="70" showInDefault="1" showInWebsite="1">
```
acampos1916, line 50: Can we add a tooltip to explain the two different behaviors?

Gateway/Request/CheckoutDataBuilder.php
candemiralp: Hey @acampos1916, thank you for the review.

This is not a concern, yet. Though, we have a separate ticket to handle it. The new field is currently wired to the legacy one in Checkout API backend. So, we are still safe.

etc/config.xml
```diff
@@ -54,6 +54,7 @@
                 <can_cancel>1</can_cancel>
                 <can_authorize_vault>1</can_authorize_vault>
                 <can_capture_vault>1</can_capture_vault>
+                <threeds_flow>native</threeds_flow>
```
acampos1916, line 57: I see we're setting as default. Do we need a migration script to update old configs? That might be too much of a breaking change though

candemiralp: As this field has been introduced with this PR, there is no need to update/migrate the config table. The default value will be obtained from this file in any case.

acampos1916: Ah you're right, this is a new field

acampos1916 (approved):

PR #2810, opened 2024-11-21
[ECP-9476]Updating Paybylink description for Magento Orders

khushboo-singhvi:
**Description**

This PR solves an issue where the description in the pay by link created from Magento admin order is diplaying {} instead of the usual value that used to have `Order XXXXXX00 from Main Website Store` on the payment screen for the customer facing.

**Tested scenarios**

Fixes

### DIFF
--- a/Gateway/Request/DescriptionDataBuilder.php
+++ b/Gateway/Request/DescriptionDataBuilder.php
@@ -29,13 +29,7 @@ class DescriptionDataBuilder implements BuilderInterface
         $payment = $paymentDataObject->getPayment();
         /** @var Order $order */
         $order = $payment->getOrder();
-
-        $request['body']['description'] = __(
-            'Order %1 from %2',
-            $order->getIncrementId(),
-            $order->getStore()->getGroup()->getName()
-        );
-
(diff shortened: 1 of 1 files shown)


candemiralp (dismissed):

Gateway/Request/DescriptionDataBuilder.php
```diff
@@ -29,13 +29,7 @@ public function build(array $buildSubject)
...
         /** @var Order $order */
         $order = $payment->getOrder();
-
-        $request['body']['description'] = __(
-            'Order %1 from %2',
-            $order->getIncrementId(),
-            $order->getStore()->getGroup()->getName()
-        );
-
+        $request['body']['description'] = 'Order ' . $order->getIncrementId() . ' from ' . $order->getStore()->getGroup()->getName();
```
candemiralp, line 32: We will lose the localisation capabilities after removing the phrase and merchants will not be able to add translation. Do we know why this is an issue now especially after migrating to model based library?

candemiralp (approved):

acampos1916 (approved):

PR #2819, opened 2024-12-03
[ECP-9482] Refactor Header Data Builder For Transaction Clients

SushmitaThakur:
**Description**
This PR -
- Refactors `HeaderDatabuilder` to use a `HeaderDataBuiderInterface` that declares all the header constants. 
- Updates `TransactionClients` (capture/cancel/refund etc.) accordingly
- Updates the path to header builder in `di.xml` and add the builder for capture, cancel, refund and paybylink
- Updates relevant observers to use correct `frontendType` constant
- Synchronises the `frontendType` value to `default` in JS renderers
- Updates relevant unit tests
- Adds unit tests for TransactionPaymentLinks class

**Tested Scenarios**
- TransactionPayment
- TransactionCancel
- TransactionCapture
- PaymentDetails Call
- TransactionPaymentLinks


Gateway/Request/Header/HeaderDataBuilderInterface.php
```diff
@@ -0,0 +1,16 @@
...
+{
+    const EXTERNAL_PLATFORM_NAME = 'external-platform-name';
+    const EXTERNAL_PLATFORM_VERSION = 'external-platform-version';
+    const EXTERNAL_PLATFORM_EDITION = 'external-platform-edition';
+    const EXTERNAL_PLATFORM_FRONTEND_TYPE = 'external-platform-frontendtype';
+    const MERCHANT_APPLICATION_NAME = 'merchant-application-name';
+    const MERCHANT_APPLICATION_VERSION = 'merchant-application-version';
+
+    const ADDITIONAL_DATA_FRONTEND_TYPE_KEY = 'frontendType';
+    const FRONTEND_TYPE_HEADLESS_VALUE = 'headless';
```
candemiralp, line 15: This value has been set but never been used. How are we planning to use this constant?

Helper/Data.php
```diff
@@ -1180,9 +1180,9 @@ public function buildRequestHeaders($payment = null)
             'merchant-application-version' => $this->getModuleVersion()
```
candemiralp, line 1180: Could you please use the constants?

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -294,7 +294,7 @@ define(
                         //This is required by magento to store the token
                         'is_active_payment_token_enabler' : this.storeCc,
                         'number_of_installments': this.installment(),
-                        'frontendType': 'default'
+                        'frontendType': 'luma'
```
acampos1916, line 297: Does it have to be 'luma'? Luma is only a default theme for magento2, just like 'blank', so in most cases it won't be strictly 'luma' being used

Gateway/Request/Header/HeaderDataBuilderInterface.php
SushmitaThakur: Yes, it was supposed to be used but never reached that point. Removed this for now.

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
SushmitaThakur: No, it doesn't but we wanted to synchronize frontendType throughout the project since we have both `luma` and `default` at different places, and we want to differentiate from other frontendTypes, 'hyva' for example. I decided on `luma` since it looked more informative but I have no preference, I can set them to `default` if that is more appropriate. Should I?

acampos1916: Then I think default would be better indeed! If it doesn't cause issues of course

candemiralp (approved):

acampos1916 (approved):

khushboo-singhvi (approved):

PR #2822, opened 2024-12-09
[ECP-9509] Fix the refund response handler to handle exceptions thrown by Checkout API 

khushboo-singhvi:
**Description**
We added the `error` key in the catch block of  the `TransactionRefund` class so that the plugin can handle exceptions thrown by Checkout API while refunding a payment.


Gateway/Http/Client/TransactionRefund.php
```diff
@@ -88,6 +88,8 @@ public function placeRequest(TransferInterface $transferObject): array
                 $this->adyenHelper->logResponse($responseData);
             } catch (AdyenException $e) {
                 $this->adyenHelper->logAdyenException($e);
+                $responseData['error'] = $e->getMessage();
+                $responseData['errorCode'] = $e->getAdyenErrorCode();
```
SushmitaThakur, line 92: There is `errorCode` present in the responseData, in case of failure ([here](https://docs.adyen.com/api-explorer/Checkout/71/post/payments/(paymentPspReference)/refunds#responses-400)), should we be overwriting that?

khushboo-singhvi: Hey Sushmita, You are correct, but there we only handle common http status. To prevent redirecting to success page in any other scenarios, we need to handle the exception with correct parameters.

candemiralp (approved):

acampos1916 (approved):

PR #2841, opened 2024-12-31
Migrating `allow3DS2` to `nativeThreeDS` on `/payments` request

khushboo-singhvi:
**Description**
As Checkout API has deprecated usage of [allow3DS2](https://docs.adyen.com/api-explorer/Checkout/71/post/payments#request-additionalData-listOfValues-allow3DS2) under additionalData on V69 and recommends using [nativeThreeDS](https://docs.adyen.com/api-explorer/Checkout/71/post/payments#request-authenticationData-threeDSRequestData-nativeThreeDS) flag instead.
This PR introduces the `nativeThreeDS` flag in the Payments request.

**Tested scenarios**

Fixes

### DIFF
--- a/Model/Config/Source/ThreeDSFlow.php
+++ b/Model/Config/Source/ThreeDSFlow.php
@@ -15,8 +15,8 @@ use Magento\Framework\Data\OptionSourceInterface;
 
 class ThreeDSFlow implements OptionSourceInterface
 {
-    const THREEDS_NATIVE = 'native';
-    const THREEDS_REDIRECT = 'redirect';
+    const THREEDS_NATIVE = 'preferred';
+    const THREEDS_REDIRECT = 'disabled';
 
     /**
      * @return array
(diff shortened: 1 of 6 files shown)


Model/Config/Source/ThreeDSFlow.php
```diff
@@ -15,8 +15,8 @@
 
 class ThreeDSFlow implements OptionSourceInterface
 {
-    const THREEDS_NATIVE = 'native';
-    const THREEDS_REDIRECT = 'redirect';
+    const THREEDS_NATIVE = 'preferred';
+    const THREEDS_REDIRECT = 'disabled';
```
acampos1916, line 19: Would it be a good idea to also change the const names to reflect the value better? Just to minimize confusion

khushboo-singhvi: I did gave it a thought, but the Possible values:
preferred: Use native 3D Secure authentication when available.
disabled: Only use the redirect 3D Secure authentication flow.

candemiralp: Hey @khushboo-singhvi, these are the constants of a public class. Modifying the values might introduce a breaking change. Besides that, the values are used in the configuration table. The changes in the values require data patch for `core_config_data`.

candemiralp (approved):

SushmitaThakur (approved):

PR #2847, opened 2025-01-09
[ECP-9480] Setting cc_type value for Carte Bancaire

khushboo-singhvi:
**Description**
When using a Cartes Bancaire card to complete the payments, the `cc_type` in the `sales_order_payment` table is null. This is the case for a normal non co branded CB card and with co-branded when selecting CB. Other card payments add `cc_type` correctly.
This PR aims to add the correct `cc_type` to the database table if it is null or not set already.

**Tested scenarios**
- Tested the correct value for non co branded CB card and with co-branded when selecting CB, and other redirect payment methods.

Fixes

### DIFF
--- a/Helper/PaymentResponseHandler.php
+++ b/Helper/PaymentResponseHandler.php
@@ -221,9 +221,8 @@ class PaymentResponseHandler
         $ccType = $payment->getAdditionalInformation('cc_type');
 
         if (!empty($paymentsDetailsResponse['additionalData']['paymentMethod']) &&
-            !is_null($ccType) &&
-            $isWalletPaymentMethod &&
-            $isCardPaymentMethod
+            is_null($ccType) &&
+            ($isWalletPaymentMethod || $isCardPaymentMethod)
         ) {
             $ccType = $paymentsDetailsResponse['additionalData']['paymentMethod'];
             $payment->setAdditionalInformation('cc_type', $ccType);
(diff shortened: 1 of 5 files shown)


Helper/PaymentResponseHandler.php
```diff
@@ -211,6 +211,14 @@ public function handlePaymentsDetailsResponse(
             $payment->setAdditionalInformation('donationToken', $paymentsDetailsResponse['donationToken']);
         }
 
+        $ccType = $payment->getAdditionalInformation('cc_type');
+
+        if (!empty($paymentsDetailsResponse['additionalData']['paymentMethod']) && $ccType == null) {
+            $ccType = $paymentsDetailsResponse['additionalData']['paymentMethod'];
+            $payment->setAdditionalInformation('cc_type', $ccType);
```
candemiralp, line 218: - This field needs to be populated only for card payments. How would you think about adding a condition for `adyen_cc` and `adyen_cc_vault` payment methods? Similarly, what is your opinion about wallet payment methods?

- Shall we implement this to `/payments` response handling?

khushboo-singhvi: Hey, Wallet Payments were already handled with details call, but adding a more specific check for card and wallet. Also, as per discussion to be on safe side, that cc_type is not skipped, adding it on `/payments` response too

candemiralp (approved):

SushmitaThakur (approved):

PR #2856, opened 2025-01-20
[ECP-9078] Implement is_open_invoice payment method configuration field

candemiralp:
**Description**

The plugin uses a hardcoded list of tx_variants to identify open invoice payment methods. However, this approach is not scalable and not adhering the usage of separate payment methods. This PR implements a new payment method configuration field to determine whether if the payment method is open invoice or not. 

This implementation now checks the configuration property `is_open_invoice` in `config.xml` instead of the hardcoded values.

**Tested scenarios**

- Klarna payment and modification tests
- Regular card payments


Gateway/Request/RefundDataBuilder.php
```diff
@@ -82,10 +89,10 @@ public function build(array $buildSubject): array
         // check if it contains a partial payment
         $orderPaymentCollection = $this->orderPaymentCollectionFactory
             ->create()
-            ->addFieldToFilter('payment_id', $payment->getId());
+            ->addFieldToFilter(OrderPaymentInterface::PAYMENT_ID, $payment->getId());
 
         // partial refund if multiple payments check refund strategy
-        if ($orderPaymentCollection->getSize() > self::REFUND_STRATEGY_ASCENDING_ORDER) {
+        if ($orderPaymentCollection->getSize() > 1) {
```
acampos1916, line 95: Why is this changing?

acampos1916 (changes requested):

Gateway/Request/RefundDataBuilder.php
candemiralp: This condition checks the size of the collection and it is not relevant to the constant. It's literally an integer comparison. `REFUND_STRATEGY_ASCENDING_ORDER` is used for the values of the dropdown refund strategy configuration field (meaning refund from last payment first).

.github/Makefile
```diff
@@ -15,7 +15,7 @@ magento:
 install:
 	composer config --json repositories.local '{"type": "path", "url": "/data/extensions/workdir", "options": { "symlink": false } }'
 	composer require "adyen/module-payment:*"
-	vendor/bin/phpcs --standard=Magento2 --extensions=php,phtml --error-severity=10 --ignore-annotations -n -p vendor/adyen/module-payment
+	vendor/bin/phpcs --standard=Magento2 --extensions=php,phtml --error-severity=10 -n -p vendor/adyen/module-payment
```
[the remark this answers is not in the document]
candemiralp: `--ignore-annotations` parameter makes PHPCS to ignore any inline ignore statements like `// phpcs:ignore`.

As the usage of `ArrayObject` class is restricted, I needed to ignore this condition in the test file.  `ArrayObject` is required to build the correct mock object for the collection. Otherwise `getIterator()` will not return the expected value. 

Besides that, the given `phpcs` execution command was not compatible with the one in our main workflow [here](https://github.com/Adyen/adyen-magento2/blob/a54c952fa26580733394f4f13f14087d2aa72283/.github/workflows/main.yml#L40). This change unblocks the mocking in the test and make the parameters consistent throughout the workflows.

acampos1916 (approved):

khushboo-singhvi (approved):

PR #2859, opened 2025-01-23
[ECP-9582] Fix broken Magento compatibility by removing union return type

### LINKED ISSUE #2842: DownloadApplePayDomainAssociationFile::execute return union type causes DIC compile error with Magento 2.4.4
**Describe the bug**
The return type for the new class DownloadApplePayDomainAssociationFile execute method has a union type `ResultInterface|ResponseInterface`. It seems thatMagento 2.4.4 framework does not properly support union types, which causes an error during our DIC compilation process.

**To Reproduce**
1. install adyen 9.12.2 via composer in a magento 2.4.4-p11
2. run DIC compile magento command `php bin/magento setup:di:compile`
3. During 4th step of the DIC compilation (Interceptors generation) an exception is thrown `call to undefined method ReflectionUnionType::getName() error`

**Expected behavior**
DIC compile should run and finalise without errors.

**Magento version**
5.4.4-p11

**Plugin version**
9.12.2

**Screenshots**
-

- Not a client side issue -

- Not a client side issue -

**Additional context**
We could argue that Magento 2.4.4 should be supporting union types, however... we could also argue that it is advisable not to use union types if they are rarely used elsewhere in the code base (and they have potential to cause trouble due to framework limitations).
See:
[Add PHP 8 Union Types Support for function Return Type](https://github.com/magento/magento2/issues/34264)
See also a quality patch that seems to fix a related issue for installation steps of Magento 2.4.4 (it does not seem to fix DIC compilation according to my quick tests):
[ACSD-59280: ReflectionUnionType::getName() error in 2.4.4-pX installations](https://experienceleague.adobe.com/en/docs/commerce-operations/tools/quality-patches-tool/patches-available-in-qpt/v1-1-50/acsd-59280-fix-for-reflection-union-type-error)

candemiralp:
**Description**

After introducing a union return type to `DownloadApplePayDomainAssociationFile` class, Magento 2.4.4 and 2.4.5 compatibility of the plugin was broken. Even though union return type is supported on PHP 8, it is not supported by that Magento versions. This PR removes the union return type from the aforementioned class.

**Tested scenarios**

- Downloading Apple Pay domain association file

Fixes #2842

### DIFF
--- a/Controller/Adminhtml/Configuration/DownloadApplePayDomainAssociationFile.php
+++ b/Controller/Adminhtml/Configuration/DownloadApplePayDomainAssociationFile.php
@@ -13,9 +13,7 @@ namespace Adyen\Payment\Controller\Adminhtml\Configuration;
 
 use Adyen\Payment\Logger\AdyenLogger;
 use Exception;
-use Magento\Framework\App\ResponseInterface;
 use Magento\Framework\Controller\ResultInterface;
-use Magento\Framework\Exception\FileSystemException;
 use Magento\Framework\Filesystem\DirectoryList;
 use Magento\Backend\App\Action\Context;
 use Magento\Framework\Controller\ResultFactory;
@@ -30,7 +28,7 @@ class DownloadApplePayDomainAssociationFile extends Action
     const WELL_KNOWN_PATH = '.well-known';
 
     public function __construct(
-        private readonly Context $context,
+        Context $context,
         private readonly DirectoryList $directoryList,
         private readonly File $fileIo,
         private readonly AdyenLogger $adyenLogger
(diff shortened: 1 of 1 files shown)


khushboo-singhvi (approved):

aschrammel:
@candemiralp Just FYI: We faced similar problems in the past on the union return type in [\Adyen\Payment\Model\OrderRepository::getOrderByQuoteId`.

aschrammel:
Hi @candemiralp

As said we received that error in the past on `bin/magento setup:die:compile`.
Can't say for sure how it is now, as we patched it on our end, removing the union type.

Not sure on which versions we've been, when it happened. Can give it a try, when I'm back if it still happens.

candemiralp:
Hello @aschrammel,

Thank you for the explanation. I've tried compiling the generated files on Magento 2.4.4 and PHP 8.2 without being able to reproduce the issue. Please let us know if the issue is persistent on your end.

Best Regards,
Can

aschrammel:
Good morning @candemiralp,

Sorry for the late reply. Seems to be fine on our end as well now.

Best regards
Andreas

PR #2868, opened 2025-01-28
Issue/2867 use async pos payments

### LINKED ISSUE #2867: Slow POS payments cause proxy timeout
### Already working on a fix, please await the PR

**Describe the bug**
When initiating a POS payment using one of the following endpoints:
- `/V1/adyen/orders/guest-carts/:cartId/pos-payment`
- `/V1/adyen/orders/carts/mine/pos-payment`

The browser might receive a timeout if the POS payment takes a bit of time and the webserver (or proxy) is configured with a reasonable (shorter) timeout period. Cloudflare for example has a timeout of 100 seconds, a value which only can modified when using one of the Enterprise plans. This could be fixed using some kind of polling mechanism. Actually the frontend code already seems to include logic for this purpose: https://github.com/Adyen/adyen-magento2/blob/cacef78df4de40d7970c4291cd438769840d0567/view/frontend/web/js/view/payment/method-renderer/adyen-pos-cloud-method.js#L116

When the browser encounters a (proxy) timeout the response will generally not include "In Progress" and this solution is also not working out-of-the-box for PWA's. 

**To Reproduce**
1. Ensure the webserver or proxy is configured with a timeout of, for example, 60 seconds
2. Start a POS / terminal payment
3. Wait for a minute before finishing the payment

**Expected behavior**
The payment should be polled as long as the POS payment status isn't final.

**Magento version**
2.4.7-p3

**Plugin version**
Tested on 9.6.0, the logic is the same at the current version (9.14.0).

**Screenshots**
Image

 - OS: MacOS
 - Browser: Firefox
 - Version: 134.0.2

N/A

**Additional context**
N/A

pmzandbergen:
**Description**
Using an asynchronous call to initiate POS payment request, followed by synchronous status calls. Polls every ~2 seconds for the status until it's final.

**Tested scenarios**
1. Place order
2. Start POS payment
  a. Cancel payment, check result (OK)
  b. Approve payment, check result (OK)

I've created functional tests for our PWA platform. Didn't test the Magento frontend. Since it's already checking for the ["In Progress" string](https://github.com/Adyen/adyen-magento2/blob/cacef78df4de40d7970c4291cd438769840d0567/view/frontend/web/js/view/payment/method-renderer/adyen-pos-cloud-method.js#L116) if the POS payment request fails, it should be working. Please add the necessary unit / functional tests if desired.

We could exchange the status requests for a request with a more reasonable timeout (e.g. 30 seconds) which gives a response as soon as the notification has been received by the webhook, or when it hits the timeout (in which case the client should retry). This would eliminate the 2 seconds delay and provide a quicker response. Since this requires more changes, I have left it at that for now.

_Note: If I would have implemented the Adyen (POS) API directly in a custom module things would probably have been organized a bit different in order to create a more robust solution ([see comments in the created issue](https://github.com/Adyen/adyen-magento2/issues/2867#issuecomment-2615389332))._

Fixes  #2867


pmzandbergen:
Replaced by: https://github.com/Adyen/adyen-magento2/pull/2869 (based on `main` branch)

PR #2909, opened 2025-03-10
[ECP-9602] Adding Uplift API keys

khushboo-singhvi:
**Description**

Add ShopperConversionID STRING (format similar to sessionID). Including these fields will enable Uplift features for our merchants.


Helper/Requests.php
```diff
@@ -301,20 +301,23 @@ public function buildAddressData($billingAddress, $shippingAddress, $storeId, $r
...
      * @param $currencyCode
      * @param $reference
+     * @param $shopperConversionId
+     * @param array $request
      * @return array
      */
-    public function buildPaymentData($amount, $currencyCode, $reference, array $request = [])
+    public function buildPaymentData($amount, $currencyCode, $reference, $shopperConversionId,array $request = []): array
     {
+        var_dump($amount);
```
candemiralp: Could you please remove this debug statement?

Helper/Requests.php
```diff
@@ -301,20 +301,23 @@ public function buildAddressData($billingAddress, $shippingAddress, $storeId, $r
...
-     * @param array $request
      * @param $amount
      * @param $currencyCode
      * @param $reference
+     * @param $shopperConversionId
+     * @param array $request
      * @return array
      */
-    public function buildPaymentData($amount, $currencyCode, $reference, array $request = [])
+    public function buildPaymentData($amount, $currencyCode, $reference, $shopperConversionId,array $request = []): array
```
candemiralp: Changing the signature of a public method is a breaking change. Could you please pass the `$shopperConversionId` as the latest optional argument? 

Asides this matter, have you considered generating a new data builder only for shopper conversion ID? So that, you don't need to modify these mature classes.

Helper/PaymentMethods.php
```diff
@@ -474,6 +479,18 @@ protected function getPaymentMethodsRequest(
                 $this->adyenHelper->padShopperReference($this->getCurrentShopperReference());
         }
 
+        $quote = $this->checkoutSession->getQuote();
```
candemiralp: There is already a `$quote` argument in the method signature. Can we leverage that instead of injecting the `CheckoutSession`?

Helper/GenerateShopperConversionId.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Adyen\Payment\Helper;
+
+use Magento\Framework\App\Helper\AbstractHelper;
+use Magento\Framework\App\Helper\Context;
+use Adyen\Payment\Helper\Util\Uuid;
+use Magento\Checkout\Model\Session as CheckoutSession;
+
+class GenerateShopperConversionId extends AbstractHelper
```
candemiralp: For a better naming convention, could you please use a noun instead of a verb in the class name like `ShopperConversionIdGenerator`? Or even `ShopperConversionId` as we can add more methods to this scope eventually.

Helper/GenerateShopperConversionId.php
```diff
@@ -0,0 +1,52 @@
...
+        parent::__construct($context);
+        $this->checkoutSession = $checkoutSession;
+    }
+
+    /**
+     * Get or generate a ShopperConversionID
+     *
+     * @return string
+     */
+    public function getShopperConversionId(): string
```
candemiralp: We always have a valid Quote object in the places where we need to leverage this method. IMO, injecting `CheckoutSession` is redundant. We can pass `$quote` as an argument to this method to simplify the class.

Helper/GenerateShopperConversionId.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        $shopperConversionId = Uuid::generateV4();
+
+        $quote = $this->checkoutSession->getQuote();
+        $payment = $quote->getPayment();
+
+        // Store shopperConversionId in additional information
+        $payment->setAdditionalInformation(self::SHOPPER_CONVERSION_ID, json_encode($shopperConversionId));
```
candemiralp: `generateV4()` method already returns a string. Is there a valid reason to convert `$shopperConversionId` value into a JSON?

Helper/GenerateShopperConversionId.php
```diff
@@ -0,0 +1,52 @@
...
+    public function __construct(
+        Context $context,
+        CheckoutSession $checkoutSession,
+    ) {
+        parent::__construct($context);
+        $this->checkoutSession = $checkoutSession;
+    }
+
+    /**
+     * Get or generate a ShopperConversionID
```
candemiralp: The comment is not accurate. This method always generates a new shopper conversion ID.

Helper/GenerateShopperConversionId.php
```diff
@@ -0,0 +1,52 @@
...
+
+        $quote = $this->checkoutSession->getQuote();
+        $payment = $quote->getPayment();
+
+        // Store shopperConversionId in additional information
+        $payment->setAdditionalInformation(self::SHOPPER_CONVERSION_ID, json_encode($shopperConversionId));
+
+        // Save the quote to persist additional_information
+        $quote->setPayment($payment);
+        $quote->save();
```
candemiralp: Could you please leverage `CartRepositoryInterface` to persist the quote object into the DB? Using the object itself to persist the data has been deprecated and will throw issues on PhpStan.

Gateway/Request/PaymentDataBuilder.php
```diff
@@ -55,20 +64,23 @@ public function build(array $buildSubject): array
...
+        $quote = $this->checkoutSession->getQuote();
         $amountCurrency = $this->chargedCurrency->getOrderAmountCurrency($fullOrder);
         $currencyCode = $amountCurrency->getCurrencyCode();
         $amount = $amountCurrency->getAmount();
-        $reference = $order->getOrderIncrementId();
+        $reference = $fullOrder->getIncrementId();
+
+        $shopperConversionIdData = $quote->getPayment()->getAdditionalInformation('shopper_conversion_id');
+
+        $shopperConversionId = !empty($shopperConversionIdData) ? (string) json_decode($shopperConversionIdData, true) : '';
```
candemiralp: Referring to my previous comment on the helper class; this block can be simplified if encoding the shopper conversion ID as JSON is redundant on the generator.

Helper/PaymentMethods.php
```diff
@@ -474,6 +479,18 @@ protected function getPaymentMethodsRequest(
...
+
+        $shopperConversionIdData = $quote->getPayment()->getAdditionalInformation('shopper_conversion_id');
+
+        $shopperConversionId = !empty($shopperConversionIdData) ? (string) json_decode($shopperConversionIdData, true) : '';
+
+        if(!empty($shopperConversionId)) {
+            $shopperConversionId = $this->generateShopperConversionId->getShopperConversionId();
+            $paymentMethodRequest["shopperConversionId"] = $shopperConversionId;
+        }
```
candemiralp: This block might produce incorrect `shopperConversionId` in the request. Please see the following.

1. There won't be a `shopperConversionId` in the request if the additional information does not have it. However, `getShopperConversionId()` has never been called if there is no additional information. So that, this if block will be skipped forever.
2. Generating a new shopper conversion ID for each `/paymentMethods` call might be a good idea without checking it's existence since the end goal is tracking the correlation between `/paymentMethods` and `/payments` calls.

Apart from that, referring to my previous comments about JSON encoding and decoding to simplify the implementation.

Gateway/Request/PaymentDataBuilder.php
candemiralp, line 1: Have you considered generating a new data builder only for shopper conversion ID? So that, you don't need to modify these mature classes.

Helper/Requests.php
```diff
@@ -301,20 +301,23 @@ public function buildAddressData($billingAddress, $shippingAddress, $storeId, $r
...
     {
+        var_dump($amount);
         $request['amount'] = [
             'currency' => $currencyCode,
             'value' => $this->adyenHelper->formatAmount($amount, $currencyCode)
         ];
 
         $request["reference"] = $reference;
+        $request["shopperConversionId"] = $shopperConversionId;
```
candemiralp: If you decide to continue using this method instead of a separate data builder, this line should be covered with an if block checking the existence of `$shopperConversionId` as it will be an optional argument and could be `null` if it is not set explicitly.

Helper/GenerateShopperConversionId.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * Get or generate a ShopperConversionID
+     *
+     * @return string
+     */
+    public function getShopperConversionId(): string
+    {
+        $shopperConversionId = Uuid::generateV4();
```
candemiralp: `generateV4` method can throw `RuntimeException`. However, this exception is not handled here.

This will result in failing the whole `/paymentMethods` request. IMO, this needs to be handled silently without failing the API call and causing negative impact on the downstream operations like `/payments` API call. What would you think about logging an error message and returning `null` if an exception is thrown here?

candemiralp:
@khushboo-singhvi could you please enable `main` workflow temporarily on this feature branch?

Helper/PaymentMethods.php
```diff
@@ -102,6 +103,8 @@ class PaymentMethods extends AbstractHelper
      * @param PaymentTokenRepositoryInterface $paymentTokenRepository
      * @param SearchCriteriaBuilder $searchCriteriaBuilder
      * @param Locale $localeHelper
+     * @param ShopperConversionId $generateShopperConversionId
```
candemiralp, line 106:
```suggestion
     * @param ShopperConversionId $shopperConversionIdHelper
```

Helper/PaymentMethods.php
```diff
@@ -474,6 +481,12 @@ protected function getPaymentMethodsRequest(
                 $this->adyenHelper->padShopperReference($this->getCurrentShopperReference());
         }
 
+        $shopperConversionId = $this->generateShopperConversionId->getShopperConversionId($quote);
```
candemiralp, line 484:
```suggestion
        $shopperConversionId = $this->shopperConversionIdHelper->getShopperConversionId($quote);
```

candemiralp (approved):

Issue #2858, opened 2025-01-23
Provide adminhtml configuration option for Payment Methods


DanieliMi:
**Is your feature request related to a problem? Please describe.**
Currently there is no easy way to configure the payment method titles for non developers. They are all configured in `config.xml` but no adminhtml configuration is provided via `system.xml`.

**Describe the solution you'd like**
It would be nice to have the configuration available in the adminhtml so the customers can change method titles themselves without a developer having to set the config via CLI.

**Describe alternatives you've considered**
An alternative would be to not use the Magento configuration at all but rather use the title provided by the adyen backend response.

**Additional context**
Might be realted to #2854.

aschrammel:
#2854 is pretty sure related this one.

For Credit Card it'll just work fine, but for all other payment methods configuration won't have any effect unless configured in `Default` scope. That's what is outlined in the linked issue 😄

candemiralp:
Hello @DanieliMi,

Thank you for raising your concern and creating a issue for this. As @aschrammel mentioned, we already have an internal investigation ticket to tackle this issue. Therefore, I'm adding this as the sub-task of #2854 as it's a duplicate.

Best Regards,
Can

candemiralp:
Hello all,

Adding sub-task wasn't a good idea as closing the main issue might not auto complete the sub-tasks.

I am closing this duplicate issue and please follow #2854 as we already have an internal ticket assigned to that issue.

Best Regards,
Can

DanieliMi:
Hi @candemiralp, I am not 100% certain it is a duplicate. #2854 aims to provide store scoped translations, while this issue aims to provide configurable options within the adminhtml. If adding adminhtml configuration (via system.xml) as well as store scoped translations is within the scope of the the other ticket the close is perfectly fine. But as I said options within the adminhtml would be a nice feature.

candemiralp:
Hi @DanieliMi,

Thank you for your timely response. I partially agree with you. We've also just now realised #2854 is complaining about only the configuration paths are not used while determining the title in different scopes. You can see my comment there.

As we support ~100 payment methods, it's not feasible to add a title input field for each one in the admin area. We are leaning towards using Adyen's translations for titles provided by Checkout API in `/paymentMethods` API response. This response contains the titles translated based on the `shopperLocale` field provided in the API request.

However, this will be a breaking change so we've decided to implement it as a part of V10 of the plugin. ETA of V10 is around the end of Q1 2025.

[...]

PR #2923, opened 2025-03-25
[ECP-9564]Use payment method titles from /paymentMethods response for translations

khushboo-singhvi:
**Description**

Previously, payment method titles were hardcoded in config.xml, which prevented proper localization of the payment method names in the checkout, even when shopperLocale was passed to the `/paymentMethods` API call.

To support localized payment method names, the frontend should now use the name field provided in the /paymentMethods response. These titles are dynamically localized based on the shopperLocale passed in the API call.
If the shopperLocale is set to "fi_FI", the names returned will be localized accordingly (if supported by Adyen).
**Tested scenarios**

Fixes  
#2858


view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -438,6 +438,28 @@ define(
...
+                    const schemeMethod = paymentMethodsObservable().paymentMethodsResponse.paymentMethods.find(function (pm) {
+                        return pm.type === 'scheme';
+                    });
+
+                    if (schemeMethod && schemeMethod.name) {
+                        return schemeMethod.name;
+                    }
+                }
+
+                return this.getCode(); // fallback if not found
```
candemiralp, line 460: Can we fall back to parent's value for fallback instead of payment method code?

view/frontend/web/js/view/payment/method-renderer/adyen-pm-method.js
```diff
@@ -81,6 +81,29 @@ define(
...
+                    const matchedMethod = paymentMethodsObservable().paymentMethodsResponse.paymentMethods.find(function (pm) {
+                        return pm.type === methodCode;
+                    });
+
+                    if (matchedMethod && matchedMethod.name) {
+                        return matchedMethod.name;
+                    }
+                }
+
+                return this.getCode();
```
candemiralp, line 104: Can we fall back to parent's value for fallback instead of payment method code?

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -438,6 +438,28 @@ define(
             getCode: function() {
                 return window.checkoutConfig.payment.adyenCc.methodCode;
             },
+
+            getTitle: function () {
+                const paymentMethodsObservable = adyenPaymentService.getPaymentMethods();
```
candemiralp, line 443: This implementation might be prune to a race condition if the payment methods response arrives later than page load. What would you think about creating a subscriber to the value of the payment methods observable and also updating the template file to use a new observable for the title instead of `getTitle()` function?

khushboo-singhvi: For this particular case, this might not be the case, as we already have a check on the phtml files where we depend on the PaymentMethodsResponse and then comes in this case. So, I think we are safe for this one. As discussed, We don't need any more changes here.

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
khushboo-singhvi: Yes, for sure

candemiralp (dismissed):

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -438,6 +438,28 @@ define(
...
+
+                if (
+                    typeof paymentMethodsObservable === 'function' &&
+                    paymentMethodsObservable() &&
+                    paymentMethodsObservable().paymentMethodsResponse &&
+                    Array.isArray(paymentMethodsObservable().paymentMethodsResponse.paymentMethods)
+                ) {
```
amihajlovski, line 450: I think there can be a better approach here in order to get the data from that function cause this way you are running it three times in order to get the paymentMethods

view/frontend/web/js/view/payment/method-renderer/adyen-pm-method.js
```diff
@@ -81,6 +81,29 @@ define(
...
+                const methodCode = this.getTxVariant();
+
+                if (
+                    paymentMethodsObservable &&
+                    typeof paymentMethodsObservable === 'function' &&
+                    paymentMethodsObservable() &&
+                    paymentMethodsObservable().paymentMethodsResponse &&
+                    Array.isArray(paymentMethodsObservable().paymentMethodsResponse.paymentMethods)
+                ) {
+                    const matchedMethod = paymentMethodsObservable().paymentMethodsResponse.paymentMethods.find(function (pm) {
```
amihajlovski, line 95: also here :)

candemiralp (dismissed):

view/frontend/web/js/view/payment/method-renderer/adyen-cc-method.js
```diff
@@ -438,6 +438,23 @@ define(
             getCode: function() {
                 return window.checkoutConfig.payment.adyenCc.methodCode;
             },
+
+            getTitle: function () {
+                const paymentMethodsObservable = adyenPaymentService.getPaymentMethods();
+                const methods = paymentMethodsObservable?.()?.paymentMethodsResponse?.paymentMethods;
+
+                if (Array.isArray(methods)) {
+                    const schemeMethod = paymentMethodsObservable().paymentMethodsResponse.paymentMethods.find(function (pm) {
```
amihajlovski, line 447: you can replace the right hand operation with methods.find(...)

PR #2950, opened 2025-04-29
[ECP-9387] Implement Adyen Giving Campaign manager V2

khushboo-singhvi:
**Description**

Giving launched a campaign manager as a lightweight way to merchants to add new campaigns. With this PR, introducing the implementation to new Campaign Manager.
Now Donations are done through /campaignManager, which means that no extra configuration is required in plugin configuration to use Adyen Giving. Merchant just needs to enable this feature.
**Tested scenarios**
Happy and Unhappy flow to donations

Fixes


Helper/DonationsHelper.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * @throws NoSuchEntityException
+     * @throws LocalizedException
+     */
+    public function fetchDonationCampaigns(array $payloadData, int $storeId): array
+    {
+        $request = new DonationCampaignsRequest($payloadData);
+
+        try {
```
zenit2001, line 42: Is the locale passed here already? 
If not, can we pass it, since it will return the config with translations if it is configured in campaign manager.

Helper/Requests.php
```diff
@@ -411,8 +411,8 @@ public function buildDonationData($buildSubject, $storeId): array
                 'type' => $paymentMethodCode
             ],
             'donationToken' => $buildSubject['donationToken'],
+            'donationCampaignId' => $buildSubject['id'],
             'donationOriginalPspReference' => $buildSubject['donationOriginalPspReference'],
-            'donationAccount' => $this->adyenConfig->getCharityMerchantAccount($storeId),
             'returnUrl' => $buildSubject['returnUrl'],
             'merchantAccount' => $this->adyenHelper->getAdyenMerchantAccount('adyen_giving', $storeId),
             'shopperInteraction' => self::SHOPPER_INTERACTION_CONTAUTH
```
zenit2001, line 418: Shopper interaction can be removed since it is not required anymore with the new implementation.

view/frontend/templates/checkout/success.phtml
```diff
@@ -45,71 +45,95 @@
...
+                        console.error('Failed to initialize Adyen Giving component:', error);
+                    }
+                }
+
+                function handleOnDonate(state, component) {
+                    if (state.isValid) {
+                        const payload = state.data;
+                        payload.returnUrl = window.location.href;
+                        payload.id = component.props.reference;
+                        debugger;
```
bponomarenko, line 119: I guess those `debugger` statements should be removed before the merge.

view/frontend/templates/checkout/success.phtml
```diff
@@ -45,71 +45,95 @@
...
-            environment: '<?= $block->escapeJs($block->getEnvironment()); ?>',
-            clientKey: '<?= $block->escapeJs($block->getClientKey()); ?>'
+                if (!merchantAccount || !currency) {
+                    console.warn('Missing merchant account or currency for donation component.');
+                    return;
+                }
+
+                try {
+                    const payload = {
+                        merchantAccount: merchantAccount,
```
bponomarenko, line 75: Is this code would be evaluated in the browser or in the backend? `/donationCampaigns` call should be done on the BE and never from within a browser.
Also `/donationCampaigns` call should only be done if the previous transaction response includes `donationToken` property.

Helper/DonationsHelper.php
khushboo-singhvi: Hey @zenit2001 - The Locale is passed already in the payload in [success.phtml](https://github.com/Adyen/adyen-magento2/blob/5a5ce729ab8f51e5be8e737470c88b848b9661b3/view/frontend/templates/checkout/success.phtml#L74)
Thanks for raising the concerns.

Helper/Requests.php
khushboo-singhvi: Hey @zenit2001, I can see the reference in the API Explorer [here](https://docs.adyen.com/api-explorer/Checkout/71/post/donations#request-shopperInteraction)
Do you have any link saying that this is obsolete for CM V2, that I might be missing?

zenit2001: Hi @khushboo-singhvi , with the previous implementation of giving the shopperInteraction was a required field otherwise the donation would fail. With the new implementation using campaign manager, we are not required to send it anymore. [Docs](https://docs.adyen.com/online-payments/donations/web-component/#make-a-donation)

Model/Api/AdyenDonationCampaigns.php
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * {@inheritdoc}
+     */
+
+    public function getCampaigns(int $orderId, string $payload): string
+    {
+        $order = $this->orderRepository->get($orderId);
+
+        if (!$order->getEntityId()) {
+            throw new LocalizedException(__('Order not found.'));
```
candemiralp, line 38: `LocalizedException` shouldn't give details about error details. This can be misused by the malicious actors. The a generic exception message can be used (ideally identical to the message we use in DonationHelper class for LocalizedException - ie: `Donation failed!`). Besides that, it might also throw EntityNotFound (or similar) exception if the user is not authorized to fetch the given order.

Model/Api/AdyenDonations.php
```diff
@@ -86,18 +86,6 @@ public function makeDonation(string $payload, OrderInterface $order): void
...
 
-        $donationAmounts = explode(',', $this->config->getAdyenGivingDonationAmounts($order->getStoreId()));
-        $formatter = $this->dataHelper;
-        $donationAmountsMinorUnits = array_map(
-            function ($amount) use ($formatter, $currencyCode) {
-                return $formatter->formatAmount($amount, $currencyCode);
-            },
-            $donationAmounts
-        );
-        if (!in_array($payload['amount']['value'], $donationAmountsMinorUnits)) {
```
candemiralp, line 97: Does Adyen block invalid donation amounts based on the campaign values?

This if-block had been introduced to mitigate a security vulnerability in the past. If Adyen accepts any amount besides the defined values in the campaign, we need to validate the values before sending the donation request.

etc/webapi.xml
```diff
@@ -124,6 +124,20 @@
...
     </route>
 
+    <route url="/V1/adyen/orders/carts/mine/donation-campaigns" method="POST">
+        <service class="Adyen\Payment\Api\AdyenDonationCampaignsInterface" method="getCampaigns"/>
+        <resources>
+            <resource ref="self"/>
+        </resources>
+    </route>
+
+    <route url="/V1/adyen/orders/guest-carts/donation-campaigns" method="POST">
```
candemiralp, line 134: Why don't we follow the Magento best practice of adding the `cartId` to the guest-user related endpoint URLs?

view/frontend/templates/checkout/success.phtml
```diff
@@ -45,71 +45,92 @@
...
-    (async function () { // RequireJS does not support async callback
-        function handleOnDonate(state, component) {
-            if (state.isValid) {
-                let payload = state.data;
-                payload.returnUrl = window.location.href;
-
-                const orderId = <?= $block->escapeJs($block->getOrder()->getId()) ?>;
+    ?>
+    <div id="donation-container"></div>
+    <script>
```
candemiralp, line 55: Could you please use `secureRenderer()` for rendering scripts? Otherwise, we might face with nonce related CSP issues.

Model/Api/AdyenDonations.php
khushboo-singhvi: Hey @candemiralp, Yes this is now being handled by Adyen. It blocks invalid transactions

view/frontend/templates/checkout/success.phtml
```diff
@@ -45,71 +45,95 @@
...
+                async function mountAdyenGivingComponent(campaign) {
+                    try {
+                        const checkoutComponent = await window.AdyenWeb.AdyenCheckout({
+                            locale: locale,
+                            countryCode: '<?= $block->escapeJs($block->getOrder()->getBillingAddress()->getCountryId()); ?>',
+                            environment: environment,
+                            clientKey: clientKey
+                        });
+
+                        const donationConfig = {
```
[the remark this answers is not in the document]
bponomarenko: @khushboo-singhvi Is there a reason for not supporting Roundup in the current plugin implementation?

candemiralp (dismissed):

view/frontend/templates/checkout/success.phtml
```diff
@@ -12,104 +12,151 @@
...
+                const locale = '$locale';
+                const environment = '$environment';
+                const clientKey = '$clientKey';
+                const orderId = '$orderId';
+                const isLoggedIn = Boolean($isLoggedIn);
+                const maskedQuoteId = '$maskedQuoteId';
+
+                try {
+                    const payload = {
+                        currency: currency,
```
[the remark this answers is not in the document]
khushboo-singhvi: I also do the double verification check, before sending the request. [Ref](https://github.com/Adyen/adyen-magento2/pull/2950/files#diff-867d636bed21cf41124ab4b1530484220124007e7d61a77f13fe6ad4198ddac7)

Helper/DonationsHelper.php
```diff
@@ -0,0 +1,82 @@
...
+        $this->adyenHelper = $adyenHelper;
+        $this->configHelper = $configHelper;
+        $this->adyenLogger = $adyenLogger;
+    }
+
+    /**
+     * @throws NoSuchEntityException
+     * @throws LocalizedException
+     */
+    public function fetchDonationCampaigns(array $payloadData, int $storeId): array
```
bponomarenko, line 42: Btw, this endpoint can be simplified to always return 1 or 0 campaign details instead of an array. 
`fetchDonationCampaign(): {}`

view/frontend/templates/checkout/success.phtml
```diff
@@ -12,104 +12,151 @@
...
-            showCancelButton: true,
-            onDonate: handleOnDonate,
-            onCancel: handleOnCancel
-        };
-        var checkoutComponent = await AdyenCheckout({
-            locale: '<?= $block->escapeJs($block->getLocale()); ?>',
-            environment: '<?= $block->escapeJs($block->getEnvironment()); ?>',
-            clientKey: '<?= $block->escapeJs($block->getClientKey()); ?>'
+
+<?php if ($block->showAdyenGiving()): ?>
```
bponomarenko, line 61: Do we also has to check for the donation token here?

view/frontend/templates/checkout/success.phtml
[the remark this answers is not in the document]
bponomarenko: It is helpful to have a verification check, indeed. But my question is more about **is this parameter (and verification for it) is necessary in the first place**? Can we remove `currency` from the FE payload altogether and instead always populate it on the BE from the cart/order?

zenit2001 (approved):

PR #2993, opened 2025-06-18
[ECP-9735] Update Filter check for case sensitive check while filtering Payment Methods

khushboo-singhvi:
**Description**
The current payment method filter mechanism performs a strict case-sensitive string comparison, which causes the payment method to be incorrectly filtered out from listing on the checkout. To address this, the filtering logic is updated to handle case-insensitive comparisons in this PR.

**Tested scenarios**

Fixes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,8 +20,7 @@
     "magento/module-graph-ql": ">=100.4.4",
     "magento/module-instant-purchase": ">=100.4.3",
     "magento/module-checkout-agreements": ">=100.4.3",
-    "ext-json": "*",
-    "magento/module-csp": "*"
+    "ext-json": "*"
   },
   "require-dev": {
     "phpunit/phpunit": "~9.6.1",


Helper/PaymentMethodsFilter.php
```diff
@@ -84,8 +84,15 @@ private function filterPaymentMethods(array $magentoPaymentMethods, array $adyen
                 );
             }
 
-            if ($txVariant &&
-                !in_array($txVariant, array_column($adyenPaymentMethods, 'type'), true)) {
+            if ($txVariant && is_string($txVariant) &&
```
candemiralp, line 87: `$txVariant` can only be `string` as `paymentMethodTypeReplace()` method has a return type `string`.

Do we really need to check explicitly here whether if it's a string or not using `is_string()`?

candemiralp (approved):

PR #3000, opened 2025-06-24
Reset quote reserved order ID with `checkout_submit_before` event

### LINKED ISSUE #2999: [Bug]: Payment canceled with multiple payment attempts
### Description

Payment canceled with multiple payment attempts

### Steps to reproduce

1. Make a first payment attempt that is refused
2. Then make a second payment attempt which is accepted

### Actual behavior

When a customer makes a first payment attempt that fails, then makes a second attempt that works, the order is sometimes canceled because the webhook of the first attempt is executed.

Here is an example of the bug :  

- at 7:07 - `X5BDWNSMTT5FBHG6` : the customer makes a first payment which is refused because fraud
- at 7:07 : Adyen send webhook AUTHORISATION for PSP `X5BDWNSMTT5FBHG6` with `success=false` in body
- at 7:10 - `G632MTL5G45BS6H3` : the client makes a new payment again which is accepted
- at 7:10 : Adyen send webhook AUTHORISATION for PSP `G632MTL5G45BS6H3`  with `success=true` in body
- at 7:12 : The notification for PSP `G632MTL5G45BS6H3` is executed
- at 7:19 : The notification for PSP `X5BDWNSMTT5FBHG6`is executed
- at 7:19 : Order is canceled with `/cancel` request to

Image

Image

Image

### Expected behavior

In this case, the order should not have been canceled since the second payment is valid.

When canceling the order, I think it is necessary to check that the PSP associated with the last payment is identical to the PSP used in the webhook.

### Code snippet or screenshots (if applicable)

_No response_

### Adyen Magento Plugin version

9.18.1

### Magento version

2.4.7-p6

### Operating System

None

### Browser (if applicable)

_No response_

### Additional context and logs

The reference `ORD-US-2025051739` is used for both attempts since the first attempt does create the order. More details here: https://github.com/Adyen/adyen-magento2/issues/2917#issuecomment-2757000213

dimitriBouteille:
**Description**

The objective of this PR is to improve the reset of the `reserved_order_id` by using the event `checkout_submit_before` which is called in graphQl and REST.

The plugins implemented in PR https://github.com/Adyen/adyen-magento2/pull/830 do not work in graphQl.

In graphQl context, the resolver `placeOrder` call `\Magento\Quote\Api\Data\PaymentInterface\CartManagementInterface::placeOrder()` and in REST context this class is also called.

`CartManagementInterface::placeOrder()` function trigger `checkout_submit_before`.

---

Fixes #2999

### DIFF
--- a/Observer/ResetQuoteReservedOrderId.php
+++ b/Observer/ResetQuoteReservedOrderId.php
@@ -53,7 +53,7 @@ class ResetQuoteReservedOrderId implements ObserverInterface
             }
         } catch (\Exception $e) {
             $this->adyenLogger->error("Failed to reset reservedOrderId for guest shopper" . $e->getMessage(), [
-                'quote_id'
+                'quote_id' => $quote->getId(),
             ]);
         }
     }


candemiralp:
Hello @dimitriBouteille,

Thank you for your contribution. However, the above mentioned plugins to reset incrementId have successfully called on our environments. Therefore, this PR might be redundant.

I left a comment [here](https://github.com/Adyen/adyen-magento2/issues/2999#issuecomment-3135647311). Let's continue discussing on the other thread.

Best Regards,
Can

PR #3005, opened 2025-06-30
[ECP-9706] Accept and save token lifecycle webhooks

khushboo-singhvi:
**Description**

Our plugin exposes `/adyen/webhook` endpoint to accept webhooks. Currently, this endpoint can only accept standard webhook payload. With this PR we are refactoring the authentication and object building functionalities to accept webhooks with different payloads (only token lifecycle webhooks at the moment but can be extended later).


Api/Webhook/WebhookAcceptorInterface.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Adyen\Payment\Api\Webhook;
+
+use Adyen\Payment\Model\Notification;
+
+interface WebhookAcceptorInterface
+{
+    public const TYPE_STANDARD = 'standard';
+    public const TYPE_TOKEN = 'token';
```
candemiralp: The requirement of updating the interface when a new acceptor type is introduced doesn't sound right (Dependency Inversion Principle). I believe these constants should be defined in their own classes.

Controller/Webhook/Index.php
```diff
@@ -11,352 +11,135 @@
...
-use Adyen\Payment\Helper\IpAddress;
-use Adyen\Payment\Helper\RateLimiter;
 use Adyen\Payment\Logger\AdyenLogger;
-use Adyen\Payment\Model\Notification;
-use Adyen\Payment\Model\NotificationFactory;
-use Adyen\Webhook\Exception\HMACKeyValidationException;
-use Adyen\Webhook\Exception\InvalidDataException;
-use Adyen\Webhook\Receiver\HmacSignature;
+use Adyen\Payment\Model\Webhook\WebhookAcceptorFactory;
+use Adyen\Payment\API\Webhook\WebhookAcceptorInterface;
```
candemiralp:
```suggestion
use Adyen\Payment\Api\Webhook\WebhookAcceptorInterface;
```

Controller/Webhook/Index.php
```diff
@@ -11,352 +11,135 @@
...
             }
 
-
             $this->getResponse()
                 ->clearHeader('Content-Type')
                 ->setHeader('Content-Type', 'text/html')
                 ->setBody($acceptedMessage);
             return;
+
         } catch (Exception $e) {
```
candemiralp: Not part of this task but, it caught my attention.

It looks like importing `Symfony\Component\Config\Definition\Exception\Exception` is not required and it creates an unnecessary external dependency. I believe it should be high level PHP `\Exception` only.

Controller/Webhook/Index.php
```diff
@@ -11,352 +11,135 @@
...
 use Magento\Framework\App\Request\Http;
 use Magento\Framework\Exception\LocalizedException;
-use Magento\Framework\Serialize\SerializerInterface;
 use Symfony\Component\Config\Definition\Exception\Exception;
-use Magento\Framework\HTTP\PhpEnvironment\RemoteAddress;
 
-/**
- * Class Json extends Action
- */
 class Index extends Action
```
candemiralp: The `Action` class has been deprecated. It can be refactored on another task. Let's catch it up later separately.

Controller/Webhook/Index.php
```diff
@@ -11,352 +11,135 @@
...
-            // Process each notification item
             $acceptedMessage = '';
-            foreach ($notificationItems['notificationItems'] as $notificationItem) {
-                $status = $this->processNotification(
-                    $notificationItem['NotificationRequestItem'],
-                    $notificationMode
-                );
 
-                if ($status !== true) {
+            foreach ($notificationItems['notificationItems'] as $notificationItemWrapper) {
```
candemiralp: Token lifecycle webhooks do not have `notificationItems` array field. This block might fail for non-standard webhook.

Controller/Webhook/Index.php
```diff
@@ -11,352 +11,135 @@
...
-            // from the Adyen Customer Area
-            $cronCheckTest = $notificationItems['notificationItems'][0]['NotificationRequestItem']['pspReference'];
-            if ($this->notificationReceiver->isTestNotification($cronCheckTest)) {
+            // Optional: check for unprocessed notifications for test mode
+            $cronCheckTest = $notificationItems['notificationItems'][0]['NotificationRequestItem']['pspReference']
+                ?? $notificationItems['notificationItems'][0]['tokenId']
+                ?? null;
+
+            if ($cronCheckTest && $this->notificationReceiver->isTestNotification($cronCheckTest)) {
                 $unprocessedNotifications = $this->adyenHelper->getUnprocessedNotifications();
```
candemiralp: The property `adyenHelper` is missing on the class.

Controller/Webhook/Index.php
```diff
@@ -11,352 +11,135 @@
...
-        $hasHmacCheck = $this->configHelper->getNotificationsHmacKey() &&
-            $this->hmacSignature->isHmacSupportedEventCode($response);
-        if ($hasHmacCheck && !$this->notificationReceiver->validateHmac(
-            $response,
-            $this->configHelper->getNotificationsHmacKey()
-        )) {
-            $this->adyenLogger->addAdyenNotification(
-                'HMAC key validation failed ' . json_encode($response)
-            );
+        if (!isset($payload['live'])) {
```
candemiralp, line 117: This block will always return `false` for non-standard webhooks as they do not have `live` field. So, it ends up with returning `401` HTTP response.

Helper/Webhook.php
```diff
@@ -467,4 +472,27 @@ private function addNotificationErrorComment(Order $order, string $errorMessage)
...
+        if (!$this->ipAddressHelper->isIpAddressValid($ip)) {
+            $this->logger->addAdyenNotification("Invalid IP for $context", $payload);
+            return false;
+        }
+        return true;
+    }
+
+    public function isMerchantAccountValid(string $incoming, array $payload, string $context = 'webhook'): bool
+    {
+        $expected = $this->configHelper->getMerchantAccount();
```
candemiralp, line 488: There is a known issue for MOTO payments. Is this compatible with ECP-8974?

Model/Webhook/StandardWebhookAcceptor.php
```diff
@@ -0,0 +1,103 @@
...
+class StandardWebhookAcceptor implements WebhookAcceptorInterface
+{
+    public function __construct(
+        private readonly Config      $configHelper,
+        private readonly NotificationFactory  $notificationFactory,
+        private readonly NotificationReceiver $notificationReceiver,
+        private readonly HmacSignature        $hmacSignature,
+        private readonly SerializerInterface  $serializer,
+        private readonly AdyenLogger          $adyenLogger,
+        private readonly Webhook              $webhookHelper,
```
candemiralp:
```suggestion
        private readonly Webhook              $webhookHelper
```

Model/Webhook/WebhookAcceptorFactory.php
```diff
@@ -0,0 +1,22 @@
...
+        private readonly StandardWebhookAcceptor $standardWebhookAcceptor,
+        private readonly TokenWebhookAcceptor $tokenWebhookAcceptor
+    ) {}
+
+    public function getAcceptor(string $type): WebhookAcceptorInterface
+    {
+        return match ($type) {
+            WebhookAcceptorInterface::TYPE_STANDARD => $this->standardWebhookAcceptor,
+            WebhookAcceptorInterface::TYPE_TOKEN    => $this->tokenWebhookAcceptor,
+            default                                 => throw new \InvalidArgumentException("Unsupported webhook type [$type]"),
```
candemiralp:
```suggestion
            default                                 => throw new \InvalidArgumentException("Unsupported webhook type [$type]")
```

candemiralp (approved):

zenit2001 (approved):

PR #3010, opened 2025-07-02
fix: fix code generation error of the OrderRepository class

aadmathijssen:
In developer mode, code generation of the `OrderRepository` class failed because of the union return type of the `getOrderByQuoteId` method. This PR resolves this by changing the resturn type from `OrderInterface|false` to `?OrderInterface`. Although this changes the error value from `false` to `null`, this does not affect usages of this method in this module.

Fixes  #3006.

Issue #3006, opened 2025-06-30
[Bug]: OrderRepository::getOrderByQuoteId return union type causes code generation error


aadmathijssen:
### Description

On-the-fly code generation of the `\Adyen\Payment\Model\Sales\OrderRepository` class fails in developer mode because the `getOrderByQuoteId` method has a union return type `false|OrderInterface`. It seems that the code generation cannot handle union return types.

### Steps to reproduce

1. Install plugin version v9.19.0 via Composer in a Magento 2.4.5-p13.
1. Make sure you are in developer mode
2. Remove the `generated/code/Adyen/Payment/Model/Sales/OrderRepository/Interceptor.php` file if it exists.
3. Run the Magento application such that instantiation of the `\Adyen\Payment\Model\Sales\OrderRepository` class is triggered using the object manager. Although this can be done by reaching the order success page after placing an order using an Adyen payment method, the easiest way to reproduce this issue is by running the following from the command-line:

```bash
php -r 'require "app/bootstrap.php"; \Magento\Framework\App\Bootstrap::create(BP, [])->getObjectManager()->get(\Adyen\Payment\Model\Sales\OrderRepository::class);'
```

### Actual behavior

The command exists with exit code 255 and prints the following to stderr:

```
PHP Fatal error:  Uncaught Error: Call to undefined method ReflectionUnionType::getName() in /app/vendor/magento/framework/Interception/Code/Generator/Interceptor.php:216
Stack trace:
#0 /app/vendor/magento/framework/Interception/Code/Generator/Interceptor.php(105): Magento\Framework\Interception\Code\Generator\Interceptor->getReturnTypeValue(Object(ReflectionMethod))
#1 /app/vendor/magento/framework/Interception/Code/Generator/Interceptor.php(76): Magento\Framework\Interception\Code\Generator\Interceptor->_getMethodInfo(Object(ReflectionMethod))
#2 /app/vendor/magento/framework/Code/Generator/EntityAbstract.php(228): Magento\Framework\Interception\Code\Generator\Interceptor->_getClassMethods()
#3 /app/vendor/magento/framework/Interception/Code/Generator/Interceptor.php(177): Magento\Framework\Code\Generator\EntityAbstract->_generateCode()
#4 /app/vendor/magento/framework/Code/Generator/EntityAbstract.php(108): Magento\Framework\Interception\Code\Generator\Interceptor->_generateCode()
[...]

PR #3011, opened 2025-07-02
[ECP-9753] Fix compiling issues related to mix return types

khushboo-singhvi:
**Description**

Even though PHP 8 supports multiple method return types, Magento 2 does not support it and throws exception while compiling the generated files on 2.4.4 and 2.4.5.

### DIFF
--- a/Model/Api/GuestAdyenPosCloud.php
+++ b/Model/Api/GuestAdyenPosCloud.php
@@ -11,13 +11,13 @@
 
 namespace Adyen\Payment\Model\Api;
 
+use Adyen\AdyenException;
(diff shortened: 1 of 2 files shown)


amihajlovski (approved):

aadmathijssen:
Hi, just before this PR was I created, I already created PR #3010 for issue #3006. This also resolves the code generation issue, but instead of removing the return type, the return type is changed from `OrderInterface|false` to `?OrderInterface`. Although I prefer keeping the return type, I understand this might be considered a breaking change, so removing the return type altogether might be the best solution. So I'm ok with closing my PR in favor of this one, which also fixes a TypeError in the `payByCart` method.

zenit2001 (approved):

Issue #3074, opened 2025-09-08
[ECP-9804] [Bug]: V10.2.0 Webhook validation fails when using different merchant accounts per website scope


f123248:
### Description

Our Magento site is configured with different merchant accounts for different website scopes

### Steps to reproduce

1. Install Adyen module v10.2.0 and Magento 2.4.8-p2
2. Configure different merchant accounts under website scope in Stores > Configuration > Payment Methods > Adyen
3. Place an order on a non-default website using its specific merchant account
4. Trigger a webhook event from Adyen
5. Observe webhook validation fails

### Actual behavior

1. Webhook requests fail validation if the merchant account is not the default one.
2. All websites except the default fail webhook processing.

### Expected behavior

1. The validation should check against all configured merchant accounts (across website scopes).
2. Implementation should collect all scope values into an array and confirm that the incoming webhook merchant account exists in that array.

### Code snippet or screenshots (if applicable)

Helper/Webhook.php
```php
    public function isMerchantAccountValid(string $incoming, array $payload, string $context = 'webhook'): bool
    {
        $expected = $this->configHelper->getMerchantAccount();

        if ($expected === null) {
            $expected = $this->configHelper->getMotoMerchantAccounts();
        }

        $isValid = is_array($expected)
            ? in_array($incoming, $expected, true)
            : $incoming === $expected;

        if (!$isValid) {
            $expectedDisplay = is_array($expected) ? implode(', ', $expected) : (string)$expected;
            $this->logger->addAdyenNotification(
                "Merchant account mismatch for $context. Expected: $expectedDisplay, Received: $incoming",
                $payload
            );
            return false;
        }

        return true;
    }
```

### Adyen Magento Plugin version

10.2.0

### Magento version

2.4.8-p2

### Operating System

Linux

### Additional context and logs

Logs:
```
[2025-09-08T08:08:22.947940+00:00] Adyen Logger.INFO: Merchant account mismatch for webhook. Expected: xxx, Received: xxyz
```

<img width="994" height="392" alt="Image" src="https://github.com/user-attachments/assets/41d6d589-2c08-4058-af00-bf1f3c476ccb" />

khushboo-singhvi:
Hey @f123248,

Thanks for reporting this! We’ve created an internal ticket to track it. We’ll pick this up soon and aim to include the fix in the next upcoming release. We’ll update this issue as soon as it’s merged and released.

Regards!
Khushboo

PR #3080, opened 2025-09-15
[ECP-9804] Webhook validation fails when using different merchant accounts per website scope

khushboo-singhvi:
**Description**
If the Magento is configured with the different stores, it always picks up the default merchant account and not the one configured on the Store.

This results in failing of the Webhook processing validation on all the stores except for the default one.
This PR aims to pass the correct store Id to fetch the relevant merchant account.

**Tested scenarios**

Fixes  
#3074


Helper/Webhook.php
```diff
@@ -492,10 +495,25 @@ public function isIpValid(array $payload, string $context = 'webhook'): bool
 
     public function isMerchantAccountValid(string $incoming, array $payload, string $context = 'webhook'): bool
     {
-        $expected = $this->configHelper->getMerchantAccount();
+        $originalReference = $payload['pspReference'] ?? $payload['data']['storedPaymentMethodId'] ?? null;
```
candemiralp: `$payload['data']['storedPaymentMethodId']` does not provide the payment pspReference but the recurring detail reference. Therefore, it can't be used to obtain the payment object.

Helper/Webhook.php
```diff
@@ -490,12 +490,14 @@ public function isIpValid(array $payload, string $context = 'webhook'): bool
         return true;
     }
 
-    public function isMerchantAccountValid(string $incoming, array $payload, string $context = 'webhook'): bool
+    public function isMerchantAccountValid(string $incoming, array $payload, $order, string $context = 'webhook'): bool
```
candemiralp: Updating the method signature introduces a breaking change. Considering we only need the `storeId`, what would you think about passing an optional nullable `storeId` as the last argument of this method?

Model/Webhook/TokenWebhookAcceptor.php
```diff
@@ -83,20 +104,27 @@ private function validate(array $payload): void
             }
         }
 
-        $isLiveMode = $payload['environment'] === 'live' ? 'true' : 'false';
-
-        if (!$this->notificationReceiver->validateNotificationMode($isLiveMode, $this->configHelper->isDemoMode())) {
+        if (!$this->notificationReceiver->validateNotificationMode($isLive, $this->configHelper->isDemoMode())) {
```
candemiralp: It looks like `storeId` is also required for `isDemoMode()` method of Config class.

Model/Webhook/TokenWebhookAcceptor.php
```diff
@@ -60,12 +60,33 @@ public function __construct(
...
+     */
     public function getNotifications(array $payload): array
     {
-        $this->validate($payload);
+        $order = null;
+
+        if (!isset($payload['eventId'])) {
+            $this->adyenLogger->addAdyenNotification('Missing required field [eventId] in token webhook', $payload);
+            throw new InvalidDataException();
+        }
```
candemiralp: This block seems like redundant. The same validation runs in the `validate()` method afterwards. If we move the next block (in which the order object is obtained) to `validate()` function after validating required fields, we can get rid of this redundancy.

We still need to pass the order object to multiple places. For that purpose, a private nullable `OrderInterface` _property_ (`null` by-default) can be defined on the `TokenWebhookAcceptor` class and this property could be set with the same logic in the `validate()` method. So that, you don't need to pass the order object to `toNotification()` method in this case and you can directly access order property of the class.

Model/Webhook/StandardWebhookAcceptor.php
```diff
@@ -74,15 +94,15 @@ public function getNotifications(array $payload): array
      * @throws HMACKeyValidationException
      * @throws AuthenticationException
      */
-    private function validate(array $item, string $isLiveMode): void
+    private function validate(array $item, string $isLiveMode, $order): void
```
candemiralp: If the previous discussion makes sense, only the 'storeId` could be passed here.

Model/Webhook/TokenWebhookAcceptor.php
```diff
@@ -83,20 +104,27 @@ private function validate(array $payload): void
...
-                "Merchant account mismatch while handling the webhook!",
-                $payload
-            );
-            throw new InvalidDataException();
+
+        // TODO: Skip merchantAccount validation for `recurring.token.disabled` token webhook notifications.
+        // Reason: The token-lifecycle `recurring.token.disabled` payload does not include the original payment PSP reference
+        // or any order/payment association. Without that, we cannot reliably resolve the store scope to validate the
+        // incoming merchantAccount against the store configuration. Therefore, we temporarily skip merchantAccount validation
+        // for this specific event type.
```
candemiralp, line 123: This detailed explanation makes a lot of sense, thank you!

Model/Webhook/TokenWebhookAcceptor.php
```diff
@@ -83,20 +96,42 @@ private function validate(array $payload): void
             }
         }
 
-        $isLiveMode = $payload['environment'] === 'live' ? 'true' : 'false';
+        // Resolve order & store and keep them as class state
+        $this->order = null;
```
candemiralp: This has already been defined above. The assignment is unnecessary.

Model/Webhook/TokenWebhookAcceptor.php
```diff
@@ -60,11 +64,20 @@ public function __construct(
...
+     * @throws InvalidDataException
+     * @throws AuthenticationException
+     */
     public function getNotifications(array $payload): array
     {
-        $this->validate($payload);
+        if (!isset($payload['eventId'])) {
+            $this->adyenLogger->addAdyenNotification('Missing required field [eventId] in token webhook', $payload);
+            throw new InvalidDataException();
+        }
```
candemiralp: Do we still need this block?

Model/Webhook/TokenWebhookAcceptor.php
```diff
@@ -60,11 +64,20 @@ public function __construct(
...
 
         $isLive = $payload['environment'] === 'live' ? 'true' : 'false';
+        $this->validate($payload, $isLive);
```
candemiralp, line 80: I see the point of parametrising `$isLive` and passing it to the downstream for convenience. However, the validation of `environment` field has been sacrificed here.

PR #3081, opened 2025-09-16
[ECP-8987] Implement EXPIRE webhook handler

### LINKED ISSUE #2871: [ECP-8987]Create a handler for EXPIRE webhook
please see:
Expired payments #2489

Could you consider this?

When Adyen sent the webhook to magneto, that the payment is expired that we could change the order status?

See the webhook in magento that we already receive, but we cannot change the status of the order, ex from approved to payment expired:
Approved (Payment Authorized) Customer Not Notified
The order failed to update: Unsupported webhook notification: EXPIRE

candemiralp:
**Description**

This PR implements a handler to process EXPIRE webhook. Also, a new configuration field has been introduced to enable this feature. Note that, this handler is disabled by-default to prevent unexpected order cancellations. 

The feature can be enabled by setting `Testing and performance / Ignore expire webhook` field to `No`.

This handler cancels the order if the whole authorisation expires and the order should not have any shipment. Otherwise, partial captures and shipments create an undesired complexity on the business logic and this complexity can be handled in many different ways by our merchants.

**Tested scenarios**

- Expiry on an order with shipment
- Expiry of a partially captured authorisation
- Expiry of an uncaptured authorisation.

Fixes #2871

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -127,7 +127,7 @@ class Data extends AbstractHelper
     {
         return [
             CaptureMode::CAPTURE_MODE_AUTO => CaptureMode::CAPTURE_MODE_AUTO_LABEL,
-            CaptureMode::CAPTURE_MODE_MANUAL => CaptureMode::CAPTURE_MODE_MANUAL,
+            CaptureMode::CAPTURE_MODE_MANUAL => CaptureMode::CAPTURE_MODE_MANUAL_LABEL,
             CaptureMode::CAPTURE_MODE_ONSHIPMENT => CaptureMode::CAPTURE_MODE_ONSHIPMENT_LABEL
         ];
     }


zenit2001 (approved):

dronerdk:
Hello

You shold make this more dynamic. Why not make the user choose what order status the order should have when the webhook with expire is sent?

Then we can make a new pay by link to the customer, without cancelling the order.

candemiralp:
Hello @dronerdk,

Thanks for the follow-up. This flow is already an edge case and has not been requested by many merchants. Instead of building more moving blocks for this feature, we wanted to keep it simple and handle at least order cancellation flow by default. Also, this implementation allows merchants to customize the flow according to their business flows.

Therefore, I would encourage you to write a plugin for `ExpireWebhookHandler` to implement your own business logic.

Best Regards,
Can

dronerdk:
Thank. No worries. We all ready made a module that handles the webhook and send a email to the customer with a payment link with the expired due a mount and handles the payment of the link and changes the order status

candemiralp:
Sounds good, thank you for your understanding @dronerdk!

Best Regards,
Can

PR #3139, opened 2025-11-14
[ECP-9670] Handling Intermediate Result Codes

khushboo-singhvi:
**Description**
Magento currently handles '[Intermediate result codes](https://docs.adyen.com/online-payments/build-your-integration/payment-result-codes/#intermediate)' as Authorised Outcomes. Shopper experience and merchant operations are impacted in case the payment was not authorised or error occurred. Now, the shopper lands on a Success page when Received and Pending are returned.

All payment methods that send out an intermediate result code are affected. 
This causes shoppers to land on a success page instead of an intermediate page where they won't be under the impression that they have successfully completed the order.

This PR aims to solve the user experience with the clear message of the Order status

### DIFF
--- a/view/frontend/templates/checkout/success.phtml
+++ b/view/frontend/templates/checkout/success.phtml
@@ -32,7 +32,7 @@ if ($block->isIntermediateResult()):
     </div>
 
     <?php
-    $scriptTitle = <<<JS
+    $scriptPendingMessage = <<<JS
     require(['jquery'], function($) {
     try {
         $('.adyen-pending-message').show();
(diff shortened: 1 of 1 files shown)


Block/Checkout/Success.php
```diff
@@ -187,4 +189,53 @@ public function getIsCustomerLoggedIn(): bool
...
+     *
+     * This message will be displayed when the payment is still being processed.
+     * The message will contain the order increment id if available.
+     *
+     * @return string
+     * @throws LocalizedException
+     */
+    public function getPendingMessage(): string
+    {
+        return (string)__(
```
candemiralp: String casting here might create some issues in the downstream as it won't be a Phrase instance anymore. Therefore, the translations might be broken. Is there a strong reason to cast the type here?

candemiralp (approved):

shubhamk67 (approved):

PR #3154, opened 2025-11-25
[ECP-9864] ApplePay browser compatibility fix for Android non safari browser

khushboo-singhvi:
**Description**
This PR hides the Apple Pay payment method on unsupported devices via checking the device type via UserAgent

**Tested scenarios**

Fixes

### DIFF
--- a/view/frontend/web/js/view/payment/method-renderer/adyen-applepay-method.js
+++ b/view/frontend/web/js/view/payment/method-renderer/adyen-applepay-method.js
@@ -40,6 +40,11 @@ define(
                 }
 
                 return applePayConfiguration;
+            },
+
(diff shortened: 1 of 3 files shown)


Funfelek:
earlier the code was:
`checkBrowserCompatibility: function () {
                // Disables Apple Pay for non-Safari browsers
                return /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
            }`
            
now it is like that:
`checkBrowserCompatibility: function () {
                // Disables Apple Pay for non-Safari browsers
                return /^((?!android).)*safari/i.test(navigator.userAgent);
            }`            
            
so i can see apple pay button on Chrome browsers, i think its not correct

PR #3161, opened 2025-12-05
[ECP-9860] Fix Adyen payment methods returned when plugin is disabled for specific store

shubhamk67:
**Description**
This PR adds store-level config check to prevent Adyen payment methods from being returned when plugin is disabled for a specific store in multistore setup.


Helper/MagentoPaymentDetails.php
```diff
@@ -46,7 +52,6 @@ public function addAdyenExtensionAttributes(PaymentDetailsInterface $result, int
 
         $extensionAttributes->setAdyenPaymentMethodsResponse($adyenPaymentMethodsResponse);
 
-        $storeId = $quote->getStoreId();
         $isAdyenPosCloudEnabled = $this->configHelper->getAdyenPosCloudConfigData('active', $storeId, true);
```
khushboo-singhvi: It seems like we are checking POS Cloud activation twice — Line 42 and line 55. Since both checks rely on the same store ID and configuration, can we simplify this by reusing the result.

candemiralp (approved):

PR #3206, opened 2026-01-02
Reliability tracking framework

candemiralp:
Reliability Tracking Framework is a built-in feature of the plugin where the SLOs of the critical payment flows can be measured by calculating the ratio of the number of successful relevant events. As a part of this initiative, the following actions are being tracked if the feature is enabled.
1. Obtaining Adyen payment methods
2. Fetching redeemed gift cards
3. Place order action with an Adyen payment method
4. Submitting payment's details to Adyen
5. Donations and campaign management

Note that, this feature is disabled by default and can be enabled by setting `Collect and submit application reliability data` field to `Yes` under `Testing and performance` section of the configuration block. 

This framework collects the events for the above mentioned actions and submits to Adyen every minute. The framework is especially useful for Adyen's operational team's to proactively monitor the health of merchants' instances and for the development team to measure the quality of the releases.

**There is no payment data or PII collected for the desired purpose.**
The plugin only sends the state of the tracked event name together with the timestamp.

### DIFF
--- a/Gateway/Request/RefundDataBuilder.php
+++ b/Gateway/Request/RefundDataBuilder.php
@@ -171,20 +171,23 @@ class RefundDataBuilder implements BuilderInterface
                 ]
             ];
 
-            if ($method === PaymentMethods::ADYEN_PAYPAL) {
+            $requiresCapturePspreference = $this->paymentMethodsHelper->getRefundRequiresCapturePspreference(
+                $paymentMethodInstance
+            );
+
+            /*
+             * Following section adds `capturePspReference` to the refund request if the payment method requires it.
+             * This is used to link the refund to the capture on Adyen on multiple partial capture cases. Note that,
+             * AdyenInvoice only exists if the capture mode is manual. Hence, checking the existence of AdyenInvoice
+             * is sufficient to determine the capture mode.
+             */
+            if ($requiresCapturePspreference) {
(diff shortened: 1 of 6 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -3,7 +3,7 @@ run-name: E2E test workflow with test suite branch ${{inputs.testBranch || 'deve
 
 on:
   pull_request:
-    branches: [main]
+    branches: [main, feature/reliability]
```
candemiralp:
```suggestion
    branches: [main, feature/reliability]
```
```suggestion
    branches: [main]
```

.github/workflows/main.yml
```diff
@@ -3,9 +3,9 @@ run-name: Main
 
 on:
   pull_request:
-    branches: [main]
+    branches: [main, feature/reliability]
```
candemiralp:
```suggestion
    branches: [main]
```

.github/workflows/main.yml
```diff
@@ -5,7 +5,7 @@ on:
   pull_request:
     branches: [main]
   push:
-    branches: [main]
+    branches: [main, feature/reliability]
```
candemiralp:
```suggestion
    branches: [main]
```

PR #3285, opened 2026-03-26
[ECP-9693] Add payment method title override configuration

### LINKED ISSUE #2858: Provide adminhtml configuration option for Payment Methods

shubhamk67:
**Description**
This PR adds a dynamic configuration table under Accepting Payments > Online Checkout that allows merchants to override payment method titles returned by the /paymentMethods API response.

Fixes https://github.com/Adyen/adyen-magento2/issues/2858

### DIFF
--- a/Block/Adminhtml/System/Config/Field/PaymentMethodTitles.php
+++ b/Block/Adminhtml/System/Config/Field/PaymentMethodTitles.php
@@ -1,0 +1,81 @@
+<?php
+/**
+ *
+ * Adyen Payment module (https://www.adyen.com/)
+ *
+ * Copyright (c) 2026 Adyen N.V. (https://www.adyen.com/)
+ * See LICENSE.txt for license details.
+ *
+ * Author: Adyen <magento@adyen.com>
+ */
+
+namespace Adyen\Payment\Block\Adminhtml\System\Config\Field;
+
+use Magento\Config\Block\System\Config\Form\Field\FieldArray\AbstractFieldArray;
+use Magento\Framework\DataObject;
+use Magento\Framework\Exception\LocalizedException;
+
+class PaymentMethodTitles extends AbstractFieldArray
+{
+    private ?PaymentMethodType $_paymentMethodTypeRenderer = null;
+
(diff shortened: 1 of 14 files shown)


shanikantsingh (dismissed):

Block/Adminhtml/System/Config/Field/PaymentMethodTitles.php
```diff
@@ -0,0 +1,81 @@
...
+
+namespace Adyen\Payment\Block\Adminhtml\System\Config\Field;
+
+use Magento\Config\Block\System\Config\Form\Field\FieldArray\AbstractFieldArray;
+use Magento\Framework\DataObject;
+use Magento\Framework\Exception\LocalizedException;
+
+class PaymentMethodTitles extends AbstractFieldArray
+{
+    private ?PaymentMethodType $_paymentMethodTypeRenderer = null;
```
candemiralp: Could you please change the property name to conform with Sonar rules?

Model/Config/Backend/PaymentMethodTitles.php
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * Serialize the dynamic table rows before saving to the database.
+     * Duplicate payment method types are deduplicated (last row wins).
+     *
+     * @return $this
+     */
+    public function beforeSave(): static
+    {
+        $value = $this->getValue();
```
candemiralp, line 61: What would you think about adding a validator to check the uniqueness of the values? So that, each payment method can only have one override per store.

candemiralp (approved):

shanikantsingh (approved):

